PR #326, opened 2014-02-07
Feature/logger

ZzubbuzZ:
logger module with dynamic allocation in AHB0 of the log buffer.


wolfmanjm:
has conflicts and cannot be merged. I also think this is not really needed.

PR #340, opened 2014-03-24
Laser coolant flow sensor support

jasonharper:
Retrying, without extraneous commits.


wolfmanjm:
Edge has changed too much to be able to merge this.Also interrupt pins have been added to Pins.h
Reopen when it can be merged

PR #440, opened 2014-06-22
Feature/set output pin

Lodran:
A simple method for configuring default pin values for pins that need to be set at startup, but don't need to be modified thereafter.

Example configuration:

set_output_pin.alpha_ms1_pin           10.29!
set_output_pin.alpha_ms2_pin           10.30
set_output_pin.alpha_ms3_pin           10.31!

Would set the stepper motor mode pins for the alpha axis to 0b101 (32 micro-steps).


stv0g:
Is there a common coding style for indentions? Also the

```
#ifndef __configuration_h__
#define __configuration_h__
```

do not look very consistent.

wolfmanjm:
not sure this is particularly useful in its current state. Will reopen if needed or requested.

PR #558, opened 2014-12-17
Added 25point grid level strategy.

qharley:
Required for Morgan SCARA systems, and large Cartesian machines.

Will correct for minor mechanical tolerance problems


qharley:
Grid code will now read the file, and allocate memory for the size of the grid on the fly.  You can also specify the grid size using a gcode command

Usage
    -----
    G32 probes the probe points and defines the bed ZGrid, this will remain in effect until reset or M370
    G31 reports the status

    M370 clears the ZGrid and the bed leveling is disabled until G32 is run again
    M370 X7 Y9 allocates a new grid size of 7x9 and clears as above

    M371 moves the head to the next calibration position without saving for manual calibration
    M372 move the head to the next calibration position after saving the current probe point to memory - manual calibration

    M373 completes calibration and enables the Z compensation grid

    M500 saves the probe points and the probe offsets
    M503 displays the current settings

qharley:
Outstanding: A way to save and load grid different grid file, with a name specified.

src/modules/tools/zprobe/ZGridStrategy.cpp
```diff
@@ -0,0 +1,597 @@
...
+float ZGridStrategy::getZOffset(float X, float Y)
+{
+    int xIndex2, yIndex2;
+
+    float xdiff = X / this->bed_div_x;
+    float ydiff = Y / this->bed_div_y;
+
+    float dCartX1, dCartX2;
+
+    int xIndex = (int) xdiff;                  // Get the current sector (X)
```
wolfmanjm: Shouldn't this be rounded up to the next whole number? or is it deliberately truncated to the lowest integer?

qharley: Deliberately truncated down.  You want the gridvalue just below and above the value you have to calculate.

wolfmanjm: Ok then usually we would use floorf(xdiff) so that is then obvious what the intention is. No need to change it though.

qharley:
wolfmanjm: Have you started working on this, or could I add some more improvements to this code in the mean time?

wolfmanjm:
go ahead let me know when you are done :)

qharley:
Just adding in normalization to prevent large jumps

qharley:
After some regression, I have everything working again...

The normalization will add the offset to the M206 Z homing offset, removing the need for large corrections.

Ready to go wolfmanjm...

wolfmanjm:
Is that M206 optional? that is not something I want done unless I ask it to do it. setting Z is a user preference thing.

qharley:
Not optional at the moment, but trivial since the effect remains the same.  It makes the correction smaller by adding or subtracting from M206

qharley:
What I mean, is that M206 will be effected even if it is not changed

qharley:
I took sleep out... or did I?
Yes - did not work at all!

qharley:
Whoo Hoo!

PR #564, opened 2014-12-19
Add option to oversample thermistor ADC values.

PetteriAimonen:
This improves the stability of PID control by reducing the effect of quantization
on the operation of the D term.

Also switched adc_value_to_temperature() to use float math instead of unnecessarily
precise double calculations.

http://koti.kapsi.fi/jpa/stuff/other/therm_test/report.html (current edge / new default is first image, pull request + config is last image)
This feature is off by default, set temperature_control.hotend.oversample_freq to e.g. 1000 Hz to enable it.
By oversampling the ADC value, we get more accurate and less quantized temperature readings.
This in turn makes the PID D-term behave better and improves temperature stability.
The effect is largest at high temperatures (>200 C), where the steps are larger with 100K NTC.

### DIFF
--- a/src/modules/tools/temperaturecontrol/Thermistor.cpp
+++ b/src/modules/tools/temperaturecontrol/Thermistor.cpp
@@ -14,6 +14,7 @@
 #include "ConfigValue.h"
 #include "libs/Median.h"
 #include "Thermistor.h"
+#include "SlowTicker.h"
 
 // a const list of predefined thermistors
 #include "predefined_thermistors.h"
@@ -31,6 +32,7 @@
 #define r1_checksum                        CHECKSUM("r1")
 #define r2_checksum                        CHECKSUM("r2")
 #define thermistor_pin_checksum            CHECKSUM("thermistor_pin")
+#define oversample_freq_checksum           CHECKSUM("oversample_freq")
 
 Thermistor::Thermistor()
 {
@@ -69,6 +71,7 @@ void Thermistor::UpdateConfig(uint16_t module_checksum, uint16_t name_checksum)
     this->beta = THEKERNEL->config->value(module_checksum, name_checksum, beta_checksum)->by_default(this->beta)->as_number(); // Thermistor beta rating. See http://reprap.org/bin/view/Main/MeasuringThermistorBeta
     this->r1 = THEKERNEL->config->value(module_checksum, name_checksum, r1_checksum  )->by_default(this->r1  )->as_number();
     this->r2 = THEKERNEL->config->value(module_checksum, name_checksum, r2_checksum  )->by_default(this->r2  )->as_number();
+    this->oversample_freq = THEKERNEL->config->value(module_checksum, name_checksum, oversample_freq_checksum)->by_default(0)->as_int();
 
     // Thermistor math
     j = (1.0 / beta);
@@ -77,6 +80,13 @@ void Thermistor::UpdateConfig(uint16_t module_checksum, uint16_t name_checksum)
     // Thermistor pin for ADC readings
     this->thermistor_pin.from_string(THEKERNEL->config->value(module_checksum, name_checksum, thermistor_pin_checksum )->required()->as_string());
     THEKERNEL->adc->enable_pin(&thermistor_pin);
+    
+    if (this->oversample_freq > 0)
+    {
+        this->oversample_count = 1;
+        this->oversample_sum = 0;
+        THEKERNEL->slow_ticker->attach(this->oversample_freq, this, &Thermistor::oversample_tick);
+    }
 }
 
 float Thermistor::get_temperature()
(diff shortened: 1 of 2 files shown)


PetteriAimonen:
Does that close mean, "No, we don't want oversampling no matter how it is implemented" or "The current pull request is not ok, but some other way would be"? I think my data proves that atleast in some cases, oversampling is very beneficial for PID stability.

Note also that the proposed patch only add 8 bytes per thermistor and no other overhead when disabled. On typical configurations, that is only 16 bytes across the whole system, so I can't really understand why this cannot be included as an option.

minad:
@PetteriAimonen maybe @wolfmanjm considered the pull request just too old. I also observed oscillation behaviour when using PID. Will this improve when you use oversampling?

I have not yet tested your pull request. I suggest that you rebase and reopen it and I will also run a test later :)

PetteriAimonen:
@minad It helped for me. IIRC wolfmanjm said it didn't help for him, but he didn't really have any oscillation in the first place either. I haven't seen any thermistor changes in smoothieware since, I think this can still be cleanly cherry-picked. I will rebase if I receive confirmation that it would get merged, but so far it seems they don't want it.

You can try the version from my branch, it works well for me - having info from other users could give more reason to merge this. There is also a python script in my results page which can be used to take graphs of the temperature.

minad:
@PetteriAimonen Ok, thx for the response. I will check later if your patch works for me. However I wonder if it not too much to have oversampling/mean + median filtering? Wouldn't it be better to just have something like the mean value over a shifting window, maybe also with omission of spikes as you implemented it?

At least currently I am observing oscillations and I would like to get rid of those. Maybe it has something to do with the filtering and that median filtering is just not good enough?

PetteriAimonen:
@minad The filtering is not overkill. Apparently the LPC1769 ADC is very sensitive to the noise from the stepper drivers. I couldn't find any reason for this in the PCB layout, but there are a few posts on google that seem to indicate that the LPC1769 ADC is just not very good.

For me, the oscillations were caused because the D term wasn't working effectively because the temperature value didn't have enough accuracy. Oversampling helps with that.

minad:
Ok, my point is just that with your patch we have to filtering mechanisms lying over each other: Mean + Median. I wonder if that is necessary and if we could just replace everything with one mean window filter. What I am saying somehow: Is not oversampling enough then? Do we still need the median filter?

PetteriAimonen:
@minad I know what you mean :) But no, it isn't enough, I foolishly tried it at first and everything broke as soon as I started the steppers (which lead to issue 584..). The median filter is very much necessary; the spike filter in the oversampling gives a bit more protection from noise in case someone has a noisier board than mine.

minad:
Ok, maybe that is because the two filters work at different frequencies? The oversampling filters high frequent noise and the median lower frequent noise. However I still don't completely get it...

Right now I am just speculating. I have to try it with different filtering to see what happens.

PetteriAimonen:
@minad Looks like I don't have any raw data captures anymore. I did however find this data of plain average filter before & after the steppers are turned on:
https://github.com/PetteriAimonen/Smoothieware/commit/329be601c91057dbc4926d03b2cc99c534a342c0

 The noise in ADC readings takes the form of positive spikes. Often the value of the spike is 4090-4095, but it can be anywhere between the real value and 4095. In the graph the real temp (which starts rising) is the upper part. The "middle" line is when there is one spike inside average buffer and lower values when multiple spikes.

My rationale for the stackup as it is now:
1) The median filter is very good at removing the spikes.
2) The average filter is good at improving resolution.
but:
3) If you put median before average, you remove all the tiny noise that the average filter uses for resolution.
4) If you put average before median, just a single spike inside average buffer will cause the average to be wrong => there is a larger number of noisy readings => median doesn't work that well.

So my solution was to filter out most of the noise before averaging with the spike filter, and then let the median take out the rest. Seems to work well in practice, and it is pretty lightweight also.

minad:
@PetteriAimonen Cherrypicking/Rebasing doesn't work since @wolfmanjm added steinhard hart. Could you please do that?

PetteriAimonen:
@minad Nah, I don't like trying to push untested patches and I can't bother testing it atop newest edge when there is very little chance of it getting merged anyway (note the absolutely no communication from wolfmanjm). I'll just run my own branch till eternity probably.

minad:
It will be merged if it makes sense. However nobody can test it if it is not cherry pickable. I am running edge and rely on in its new features/fixes.

Why is everybody so complicated around here :-/

Am 27. Februar 2015 06:55:58 MEZ, schrieb Petteri Aimonen <notifications@github.com>:

PetteriAimonen:
@minad Basically there was enough time for merging this back in 2014, so I know that this patch in its current form is not acceptable (why it is not, I do not understand). I am still willing to work on an acceptable patch, but currently I have no information on what I need to do to get it accepted.

Getting patches through to Smoothieware seems to be very hard, atleast for me. I have complained about this in the past and it has made things even harder. So that is why I'm complicated :)

minad:
Well the point is: I am only a contributor to this project. I got some my pull requests merged (admittedly after a longer discussion, see #592, #597). I cannot guarantee that it will get merged. It might be difficult to work with @wolfmanjm since he is strongly opinated and sees himself in the position of a "gatekeeper" of the project. However I think he is open to reasoning if there are strong enough facts supporting a patch. But these facts cannot be produced when nobody can test the patch. It is not getting better if anyone behaves in a complicated way. Also there are others who are open to reasoning. I talked to @arthurwolf and he is also very reasonable and tries to settle any disputes. In general I think anyone involved in this project should calm down a bit. I have observed better atmosphere in other communities. I hope that this will improve at some point...

Now to the technical point: I don't know if your patch works for me. But as shown above I observe these oscillations in temperature which seem quite high to my taste. Maybe your patch fixes it, maybe not. But I am not going to fix your pull request to cleanly merge, since:

1. I can write my own patch, something really overengineered with different pluggable filters just for me to experiment with it. It would be fun. Maybe.
2. I want to motivate people to contribute, since this is the only way things improve - if multiple people work together. Otherwise we could do everything ourselves and waste time forever.
3. I don't want to waste my time since you know your code better and already did some experimentation. But as of now I already lost more time, I think...

minad:
Btw. I think the current median filter is not good enough, just judging from the steps in the plots. So there is already enough reason imho for a different filter. But the stability issue would be much more important for me.

PetteriAimonen:
@minad Hmm yeah, maybe extra test reports will put more pressure on this. I'll rebase my patch next week when I have next chance to test it out, or if you want I'll rebase it without testing a bit sooner.

I'm sorry if I seem a bit rude at times. It just feels that trying to contribute to Smoothieware is just waste of time - if things stay this way it is doomed to fail and would be better for me to switch to MachineKit now than later. But haven't got the money to buy new HW, so I'd like to get it working atleast somehow. (Currently the lost steps issues make my printer unusable at reasonable speeds, wolfmanjm has said that they are probably pretty difficult to fix but I'll try anyway.. just finished up ITM tracing support in sigrok so that I can debug it.)

Steps in output are not due to median filter, but due to insufficient ADC resolution without oversampling (0.2 C resolution at around 200 C temp). The most optimal filter for this would probably be trimmed mean, but I don't think there is a way to implement it within the cpu & ram limits of smoothie.

PetteriAimonen:
(Definitely wrong place for this rant, but the patch acceptance policy of wolfmanjm is just totally random. Sometimes if you fight him long enough, he'll just accept a patch without any comment or explanation why it is accepted now. So he ends up accepting a lot of bad code, and rejecting a lot of good code. Or maybe I'm just an idiot who cannot code, like he seems to think.)

minad:
If you rebase sooner, I will test it :)

Actually I am not the right person to be addressed concerning these problems. Please talk with @arthurwolf in private on IRC. You know I had or have similar problems. There are more people having problems and at some point some have to rethink their attitute or @arthurwolf has to mediate in such cases.

But in general I agree with the assessment, that a more open code review and a more transparent merging policy would help the code quality and the project.

PetteriAimonen:
@minad Rebased: https://github.com/PetteriAimonen/Smoothieware/tree/thermistor_oversample_2

I suggest this config:
temperature_control.hotend.windup           true 
temperature_control.hotend.oversample_freq  1000

minad:
@PetteriAimonen I tested it and it seems to work very well.

Only windup activated -> Oscillations:
temp-windup

With windup + oversample (two independent runs, autotuning redone for each run):
temp-windup-oversample
temp-windup-oversampled

In both cases much higher stability. I would suggest that you reopen this rebased pull request and  ask @wolfmanjm for reconsideration.

PR #580, opened 2015-01-19
add missing include

sorki:
resolves:
modules/communication/GcodeDispatch.h:29:5: error: 'FILE' does not name a type


wolfmanjm:
this compiles fine without that, what are you doing different that requires this?

sorki:
Not sure. Wasn't able to build current edge with make clean all.

Tried looking at what has changed in that file and putting back this line resolved it:
-#include "libs/StreamOutput.h"

wolfmanjm:
this is the wrong place to put it. the incude needs to go in appendfilestream, and it looks like there is a typo in that using "" instead of <>. I'll fix it upstream.

PR #585, opened 2015-01-23
Prevent dangerous extrusions: Prevent extruding when the hotend is not hot enough 

irrenhaus:
This PR adds the functionality to prevent the extruder motor from turning when the hotend is too cold to melt the plastic on 3D printers.
This should prevent any damage to the machine caused by the extruder pushing against a too cold hotend.

Note: This is my first code written for SmoothieWare, so be gentle ;) This also is part of my "The Smoothie Contest" entry ("Safety features").


wolfmanjm:
This is a good first try, thank you.

However it is the wrong way to achieve this goal.

The temp checks should be in temperature control and only tells the extruder to stop if an alarm situation occurs. Maybe issue an ON_HALT instead. (that is debatable).

The advantage is that it is more efficient as temperature control already knows what the temperature is, and already knows what the associated extruder is, and the extruder can easily be disabled. This is efficient in processing time and memory usage.

arthurwolf:
Hi there !

So, I missed something when I reviewed this first via email, and I'm really sorry about that.
There is actually a much more efficient way to implement this, and I missed this completely.

It shouldn't be -that- huge of a change, but it requires you write it the other way around.

So, you made Extruder aware of it's associated TemperatureControl, but the thing is, TemperatureControl modules are already aware of their associated Extruder module !

So what this should do, is actually work in TemperatureControl, and if a TemperatureControl module is bellow it's min temperature, the TemperatureControl disables the associated Extruder module ( by setting a "disabled" flag in Extruder ).

Tell me if that makes sense, and very very sorry for not telling you this when I first reviewed your code.

Also, not very important, but you shouldn't change the alignment of the values in the .h like you did as it makes it harder to track changes :)

I hope it's not too much work to re-code this the other way around, tell me if you need any help with it.

Cheers !

irrenhaus:
Ah, I see, thanks to both of you :)

In fact I did first have a look if I could see anything like this in the TemperatureControl code but it seems I have overlooked it.

No worries, this is the reason I started with the smallest feature of my to-do list ;).

I'll close this, throw it back into the bucket, shake it a few times and we'll see what then comes out of it!

arthurwolf:
Oh by the way, there is also a G-code to disable this feature temporarily : http://reprap.org/wiki/G-code#M302:_Allow_cold_extrudes

irrenhaus:
Good to know, I'll implement that one, too!

irrenhaus:
@arthurwolf @wolfmanjm 
I found some time to have another look at my cold extrusion prevention code.

I'm currently looking into what exactly the correct way to disable an extruder would be.
From what I'm seeing my take on this would be to add an disable_extruder_checksum. If this checksum is present in a public data request as the second element (the same way the extruder checks for save_state and restore_state checksum) the disable() method on the extruder should be called.

Since the ToolManager instance only gets created when there is more than one extruder this seems to be the correct place to do this.

The TemperatureControl then (on under-temperature) would use PublicData::set_value to issue an disable_extruder public data request using the own name as the third element (which also is the extruder identifier). The extruder whose name matches the given name then disables itself. Would this be the correct way to go forward? Just want to make sure :).

PS: For long questions like this, would you prefer me asking this in IRC or here on Github?

Issue #584, opened 2015-01-23
Stricter limits for TemperatureControl MINTEMP protection


PetteriAimonen:
Sometimes thermistors break, which could cause the NTC to read very low temperature and cause TemperatureControl to overpower the heater. Currently Smoothie detects this condition only when ADC value after median filtering is 0 or 4095.

However, there are a few cases when this protection is not enough:
1) Thermistors do not always break completely, which could cause very wrong temperatures but not fully saturated.
2) Noise can cause the ADC value to be not fully saturated after median filtering. This happens especially when using oversampling.
3) When messing up with PID values (yeah, my own fault..) I have had it overshoot to 400 C. Some protection against user errors would be a nice side-effect.

I would suggest to either:
1) Have hardcoded limits of e.g. -20 C to +350 C.
2) Have configurable min & max limits. This would require 8 bytes storage per TemperatureControl channel.
3) Have just configurable min limit. This would require 4 bytes per TemperatureControl channel.

I would also suggest to do a full ON_HALT / M999 stop (like limit switches) instead of just turning of the heater. This avoids trying to push filament to a malfunctioning extruder.

Would a patch implementing one of these options be considered for merge? Which one would be preferred?

qharley:
I have a min limit that I can pull request...

wolfmanjm:
Most of this has been done

PR #586, opened 2015-01-29
Thermistor safety (minimum temperatures)

qharley:
Added a way to specify a minimum temperature to protect against damaged thermistors. If not set, the machine will act as before.

### DIFF
--- a/ConfigSamples/Smoothieboard/config
+++ b/ConfigSamples/Smoothieboard/config
@@ -130,6 +130,8 @@ temperature_control.hotend.thermistor        EPCOS100K        # see http://smoot
 temperature_control.hotend.set_m_code        104              #
 temperature_control.hotend.set_and_wait_m_code 109            #
 temperature_control.hotend.designator        T                #
+#temperature_control.hotend.max_temp         250              # Set maximum temperature
+#temperature_control.hotend.min_temp         0                # Set minimum temperature - Will prevent heating below if set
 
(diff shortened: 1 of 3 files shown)


qharley:
Problems here?

wolfmanjm:
Yea it is not necessary we already check for very low values and handle them here...

https://github.com/RepRapMorgan/Smoothieware/blob/3a014cbbf048d483241d1a7c98f1fc8602e4166c/src/modules/tools/temperaturecontrol/Thermistor.cpp#L95

PetteriAimonen:
This is also discussed in this issue report: https://github.com/Smoothieware/Smoothieware/issues/584

The current check has a few problems:
1) If the thermistor is not 100% broken or if there is a noise on the line, the limit may not activate.
2) As soon as the limit clears, it will start heating again.
3) It does not protect against bad PID config or bad G code.

Myself I run this patch that fully halts on crazy temperatures:
https://github.com/PetteriAimonen/Smoothieware/commit/329be601c91057dbc4926d03b2cc99c534a342c0

IMO it is better to be safe, when there is no real need to handle such excessive temperatures.

qharley:
I have a unique problem that on certain hardware an open circuit thermistor reads -6 volts.
This is a fatal problem, since smoothieware will heat the hot end indefinitely... until a piece melts off of course.

Adding this will protect systems on those smoothie clones, only if configured of course.

Issue #592, opened 2015-02-05
Plan9 network filesystem support


minad:
This is the simplest network filesystem protocol I know of. There are a lot of [implementations](http://9p.cat-v.org/implementations) and even a [very small one for arduinos](https://github.com/echoline/NinePea) which could be easily adapted and C++ified. 9p is supported by the [linux kernel]( https://www.kernel.org/doc/Documentation/filesystems/9p.txt) with the tcp transport and thus would make it very easy to browse around on the smoothie compared to this crude sftp solution that we are having right now. I would love to see this and I also offer to implement that if you are interested.

arthurwolf:
Hey.

That sounds like a very interresting feature.
I'll let others jump in if it's not the case, but I think if you want to implement this it would get merged.
Being able to access the Smoothie file system over the network does sound pretty cool :)

Cheers !

minad:
Great!  I also dropped you an email concerning that. Since I am very enthusiastic with my 3d printer right now I might find time later today to hack on it. Expect something soon :grinning:

mhackney:
Yes! And count me in if you need help/testers. I’m on a mac and struggle with USB. I am now (as of last night) actually printing (!) and connecting via eth to upload images/configs and gcode using the python script. It works but a little inelegant. I was thinking this morning, “a simple get/put over tcp would be a nice entry for the smoothie contest. You beat me to it. I’ve implemented network transfer protocols.

regards,
Michael

arthurwolf:
I can't find much info about plan9. Do you have some info/tutorial about how to use it on a Linux system ? It sounds a tad obscure, compared to things like samba and nfs.

minad:
I added a link in the above https://www.kernel.org/doc/Documentation/filesystems/9p.txt

Plan9 is obscure but in fact well supported. For example KVM uses it to expose the host filesystem to the VMs. It uses the virtio transport for that. We will use the tcp transport.

minad:
It is also very simple I have to add, so you can easily write a client in python or whatever. So the sitatution is at least as good as with this sftp upload thing that we have right now.

minad:
@mhackney What about 9p support on macosx?

mhackney:
See http://9p.cat-v.org/implementations <http://9p.cat-v.org/implementations>
 and http://9p.cat-v.org/applications <http://9p.cat-v.org/applications> lists an OSX 9p application (programmers editor. There should be enough there to figure it out.

[...]

PR #597, opened 2015-02-10
Plan9 filesystem support for Smoothieboard!

minad:
See #592 for the prior discussion.

smoothie

## Usage on Linux

1. Add the option `network.plan9.enable true` to the config.
2. Reset the smoothie
3. Mount with `mount -t 9p <smoothie-ip> /mnt/smoothie` and access `/mnt/smoothie` like a normal filesystem. Have fun!1!!

The filesystem will be synchronous which means everything will be written immediately to the SD and there is no possibility for incomplete writes etc. If your `9p.ko` Linux kernel module is compiled with debugging support you can also mount with the debug option, i.e. `mount -t 9p -o debug=0x200 <smoothie-ip> /mnt/smoothie`. The messages will appear in dmesg.

## Summary of the changes

* The `Network` class was extended with calls to the plan9 handler
* The plan9 module was added which handles the 9p messages and relies on `opendir/fopen/fwrite` etc to access the filesystem.

## Known issues

1. Currently the files are opened/closed for every single packet since there is reportedly some sd card issue. This behaviour can be changed later. However like this I haven't got any issues except that maybe performance could be better. It worked well for me to edit the config files, copy over firmware.bin etc.
2. mv in the same directory is broken because the stdio rename call is broken for subdirectories. This can be reproduced via telnet mv. The problem can be easily fixed probably. I just wanted to keep the changes restricted to the Network/plan9 module for now.

### DIFF
--- a/src/libs/Network/uip/Network.cpp
+++ b/src/libs/Network/uip/Network.cpp
@@ -23,7 +23,7 @@
 #include "webserver.h"
 #include "dhcpc.h"
 #include "sftpd.h"
-
+#include "plan9.h"
 
 #include <mri.h>
 
@@ -32,6 +32,7 @@
 #define network_enable_checksum CHECKSUM("enable")
 #define network_webserver_checksum CHECKSUM("webserver")
 #define network_telnet_checksum CHECKSUM("telnet")
+#define network_plan9_checksum CHECKSUM("plan9")
 #define network_mac_override_checksum CHECKSUM("mac_override")
 #define network_ip_address_checksum CHECKSUM("ip_address")
 #define network_hostname_checksum CHECKSUM("hostname")
@@ -43,7 +44,7 @@ extern "C" void uip_log(char *m)
     printf("uIP log message: %s\n", m);
 }
 
-static bool webserver_enabled, telnet_enabled, use_dhcp;
+static bool webserver_enabled, telnet_enabled, plan9_enabled, use_dhcp;
 static Network *theNetwork;
 static Sftpd *sftpd;
 static CommandQueue *command_q= CommandQueue::getInstance();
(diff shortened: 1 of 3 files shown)


wolfmanjm:
Nice work, good job!

The only issue I have is that the uid/gid returned is 4294967294, it would be nice if this could be set in config as an option so I don't have to access it as root.

Thanks

minad:
Thx for merging!

I have to think about this uid problem. Thanks for pointing out. I will see what I can do about that. I will do some follow-up pull requests fixing that and the other issues.

arthurwolf:
Hey !

Thanks for this :)

Would you mind adding information about it here : http://smoothieware.org/network ?
Mention it in the "servers" section, and maybe do a section at the bottom specially for plan9 with the config option and the linux mount commands.
If you don't have time to do this, tell me and I'll do it ( and ask you to check if it's correct ).

Cheers.

minad:
@arthurwolf I will add that to the wiki :) However I want to stabilize it a bit first. Do you get my emails btw or not?

arthurwolf:
Yep I do, I just have a lot of backlog with the contest.

On Tue, Feb 10, 2015 at 3:08 PM, Daniel Mendler <notifications@github.com>
wrote:

> @arthurwolf <https://github.com/arthurwolf> I will add that to the wiki
> :) However I want to stabilize it a bit first. Do you get my emails btw or
> not?
>
> —
> <https://github.com/Smoothieware/Smoothieware/pull/597#issuecomment-73704266>
> .
>



-- 
Courage et bonne humeur.

minad:
@wolfmanjm Concerning the uid/gid issue - you can use the mount option dfltuid for that. I will document in the wiki.

PR #599, opened 2015-02-10
Rename issues

minad:
Hi Jim,

I fixed the rename issue in plan9/simpleshell. Rename works now in telnet and also in plan9 as expected.

Daniel


minad:
Great thx! You are quick :+1:

PR #602, opened 2015-02-11
stylistic change: replace unnecessary version class by constants

minad:


wolfmanjm:
I tried to recall why I used the class, and I think it is because I wanted to not use the data directly so I could add calculated values if necessary. This really does not save any ram as they are both constant.
This is a typical OO vs non-OO argument with no real benefit one over the other. I prefer to hide the data the way it is currently done. So on further consideration I think we will leave it the way it is. But thank you for your contribution.

wolfmanjm:
FWIW I could make the methods static so avoiding the need to instantiate the class.

PR #607, opened 2015-02-12
Fixed Makefile to use the correct standard libraries.

stv0g:
This PR may need a detailed explanation:
According to my understanding the "_s" postfix is outdated
and not used anymore.
These small versions of the newlib standard libraries have
been renamed. Nowadays the "_nano" prefix should be used (since GCC 4.8).
I assume, that "_s" stands for "small".

But instead of adding a prefix we should use the spec file.

I checked this patch against the latest Debian Jessie arm-none-eabi (GCC
4.8) toolchain as well as against the toolchain installed by
"linux_install".

### DIFF
--- a/build/common.mk
+++ b/build/common.mk
@@ -142,7 +142,7 @@ endif
 
 # Libraries to be linked into final binary
 MBED_LIBS = $(MBED_DIR)/$(DEVICE)/GCC_ARM/libmbed.a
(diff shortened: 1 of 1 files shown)


stv0g:
Did this PR breaks your buildsystem?

wolfmanjm:
probably Not I'll check it...

wolfmanjm:
for some reason this does break my build, however I have hand picked the changes and that seems to work.
Thanks and sorry for the delay.

PR #608, opened 2015-02-12
Renamed makefile -> Makefile because thats the recommondation

stv0g:
The GNU guys recommend to name the Makefile with an uppercase first
letter: http://www.gnu.org/software/make/manual/make.html#Makefile-Names

Personally, I also prefer it this way because its more consistend with
the included Rakefile.


wolfmanjm:
personal preferences are not a good reason to change what is there. We use stuff taken direct from gcc4mbed that is supported by Adam Green so we use whatever he decides to use.

stv0g:
Okay, I did'nt knew about this project before. Then I'll try to merge it upstream.

PR #611, opened 2015-02-13
simpleshell - add set command

minad:
For consistency I added a "set temp" command for consistency with "get temp". Maybe we also add more set subcommands in the future.


wolfmanjm:
Please note the comment...  "used to test out the get public data events" This was never meant to be a real command for everyday use. there is a perfectly goof Mxxx command to use instead. I actually need to remove these commands I added to test stuff.

PR #618, opened 2015-02-18
Removed file read during handle_config to prevent hang at boot.

qharley:
Related to consecutive fopen on SD.

### DIFF
--- a/src/modules/tools/zprobe/ZGridStrategy.cpp
+++ b/src/modules/tools/zprobe/ZGridStrategy.cpp
@@ -247,17 +247,28 @@ bool ZGridStrategy::handleGcode(Gcode *gcode)
 
             // M374: Save grid
             case 374:{
-                if(this->saveGrid(args)) {
+               char gridname[5];
+
+               if(gcode->has_letter('S'))  // Custom grid number
+                    sprintf(gridname, "S%3.0f", gcode->get_value('S'));
+
+                if(this->saveGrid(gridname)) {
                     gcode->stream->printf("Grid saved\n");
                 }
             }
             return true;
 
-            case 375: // Load grid values
-                if(this->loadGrid(args)){
+            case 375:{ // Load grid values
+                char gridname[5];
+
+                if(gcode->has_letter('S'))  // Custom grid number
+                    sprintf(gridname, "S%3.0f", gcode->get_value('S'));
+
+                if(this->loadGrid(gridname)){
                     this->setAdjustFunction(true); // Enable leveling code
                     gcode->stream->printf("Grid loaded\n");
                 }
+            }
             return true;
 
 /*          case 376: { // Check grid value calculations: For debug only.
--- a/src/modules/tools/zprobe/ZGridStrategy.cpp
+++ b/src/modules/tools/zprobe/ZGridStrategy.cpp
@@ -132,8 +132,11 @@ bool ZGridStrategy::handleConfig()
         this->pData[i] = 0.0F;        // Clear the grid
     }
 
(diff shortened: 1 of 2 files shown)


qharley:
One of my machines would sporadically hang on boot - found that consecutive fopen needs a delay to prevent problems.

wolfmanjm:
ping me on irc dev channel we can discuss how to do this. I have an idea how to do it.

qharley:
You can now save a grid with:
M374 MYCUSTOMGRID
and it will save it in "/sd/Zgrid.MYCUSTOMGRID"

qharley:
I will test it on a machine later today

src/modules/tools/zprobe/ZGridStrategy.cpp
```diff
@@ -251,20 +247,30 @@ bool ZGridStrategy::handleGcode(Gcode *gcode)
 
             // M374: Save grid
             case 374:{
-                if(this->saveGrid()) {
+               char gridname[5];
+
+               if(gcode->has_letter('S'))  // Custom grid number
+                    sprintf(gridname, "S%3.0f", gcode->get_value('S'));
```
wolfmanjm: This is too dangerous you will get stack corruption and crash if someone enter M374 S123456
You must use snprintf(gridname, 4, ...)

src/modules/tools/zprobe/ZGridStrategy.cpp
```diff
@@ -251,20 +247,30 @@ bool ZGridStrategy::handleGcode(Gcode *gcode)
...
             }
             return true;
 
-            case 375: // Load grid values
-                if(this->loadGrid()){
+            case 375:{ // Load grid values
+                char gridname[5];
+
+                if(gcode->has_letter('S'))  // Custom grid number
+                    sprintf(gridname, "S%3.0f", gcode->get_value('S'));
```
wolfmanjm: ditto here use snprintf()

src/modules/tools/zprobe/ZGridStrategy.cpp
```diff
@@ -288,6 +294,8 @@ bool ZGridStrategy::handleGcode(Gcode *gcode)
             return true;
 
             case 500: // M500 saves probe_offsets config override file
+                gcode->stream->printf("M375  ; Load default grid\n");
```
wolfmanjm: only if the config option save grid is set to true. as not everyone will want the grid loaded on startup.

src/modules/tools/zprobe/ZGridStrategy.cpp
```diff
@@ -327,24 +332,19 @@ bool ZGridStrategy::saveGrid() //std::string *args)
 
 }
 
-bool ZGridStrategy::loadGrid()
+bool ZGridStrategy::loadGrid(std::string args)
```
wolfmanjm: inefficient.. should be const std::string& args, or better still const char* args)

qharley: It may be inefficient, but the pointers are messing with my head today.

At least this works. ;-)

minad: @wolfmanjm I was already wondering about references in the current codebase which is not consistent in using those where it makes sense.

wolfmanjm:
FYI a bug is here https://github.com/RepRapMorgan/Smoothieware/blob/zgrid/src/modules/tools/zprobe/ZGridStrategy.cpp#L144

The pData is never initialized to nullptr so can crash randomly here. someone already ran into this bug.

PR #620, opened 2015-02-18
Implemented Servo controlled probe functionality at the ZProbe module.

barsamgr:
The following configuration parameters where added:

zprobe.servo_pin 1.24                       # Servo PWM pin. Xmin is a good connection as the pin there is PWM
                                                           # and there is also a 5V supply for the servo
                                                           # If Xmax limit is not used, you can switch it with X
min zprobe.servo_percent 85            # Percentage of movement for deployment. Default 100
zprobe.servo_disable true                 # Disable servo after a deployment or retract.
                                                          # Use if there is a lot of jittering. Default false

Also two M codes where added (same as Marlin M codes)
M401 For servo deploy
M402 For servo rectract

I did this for my own purpose, but I know that there are a lot of people looking for this functionality. I though it's good to share.

### DIFF
--- a/src/modules/tools/zprobe/ZProbe.cpp
+++ b/src/modules/tools/zprobe/ZProbe.cpp
@@ -26,6 +26,7 @@
 #include "PublicData.h"
 #include "LevelingStrategy.h"
 #include "StepTicker.h"
+#include "PwmOut.h" // mbed.h lib
 
 // strategies we know about
 #include "DeltaCalibrationStrategy.h"
@@ -39,6 +40,9 @@
 #define fast_feedrate_checksum   CHECKSUM("fast_feedrate")
 #define probe_height_checksum    CHECKSUM("probe_height")
 #define gamma_max_checksum       CHECKSUM("gamma_max")
+#define servo_pin_checksum       CHECKSUM("servo_pin")
+#define servo_percent_checksum   CHECKSUM("servo_percent")
+#define servo_disable_checksum   CHECKSUM("servo_disable")
 
 // from endstop section
 #define delta_homing_checksum    CHECKSUM("delta_homing")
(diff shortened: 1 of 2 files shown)


barsamgr:
For some reason the Comment is a little bit messed up.
The config parameters are:
zprobe.servo_pin 1.24                 # Servo PWM pin. Xmin is a good connection as the pin there is PWM
                                                    # and there is also a 5V supply for the servo
                                                    # If Xmax limit is not used, you can switch it with X
min zprobe.servo_percent 85     # Percentage of movement for deployment. Default 100
zprobe.servo_disable true          # Disable servo after a deployment or retract.
                                                   # Use if there is a lot of jittering. Default false

barsamgr:
I suppose there is no way to align the comments....
sorry for the mess, I am kind of new to GitHub managing.

src/modules/tools/zprobe/ZProbe.cpp
```diff
@@ -390,3 +411,29 @@ float ZProbe::zsteps_to_mm(float steps)
...
+
+void ZProbe::deploy_servo()
+{
+	if(this->servo_pin == NULL) return;
+
+	this->servo_pin->period_ms(20);
+	this->servo_pin->pulsewidth_us(1000 + this->servo_percent * 10);
+
+	if(this->servo_disable) {				//Disable servo to eliminate jitter
+		wait_ms(250);
```
wolfmanjm: You cannot put delays like this in the code. it needs to loop on an on_idle call, you can look at other modules for examples

src/modules/tools/zprobe/ZProbe.h
```diff
@@ -10,9 +10,14 @@
 
 #include "Module.h"
 #include "Pin.h"
+#include "wait_api.h"
```
wolfmanjm: this would go in the cpp file not the header. but as delays are not allowed it should be removed.

barsamgr:
I have done the changes, but because I deleted the original fork, I will close this request and open a new one.

PR #621, opened 2015-02-18
M32 fix

### LINKED ISSUE #619: Wrong progress reports
The number of bytes in total is wrong. Maybe some fs problem.

~~~
0 % complete, elapsed time: 1687 s, est time: 1245307818 s
SD printing byte 5652/3735928559
~~~

Also this elapsed time report is wrong. I just let it run for a few seconds.
This reports somehow the full 4GB?

Someone else seeing this?

minad:
Fixes #619

### DIFF
--- a/src/modules/utils/player/Player.cpp
+++ b/src/modules/utils/player/Player.cpp
@@ -218,6 +218,8 @@ void Player::on_gcode_received(void *argument)
                 }
             }
 
+            this->played_cnt = 0;
+            this->elapsed_secs = 0;
         } else if (gcode->m == 600) { // suspend print, Not entirely Marlin compliant
             this->suspend_command("", &(StreamOutput::NullStream));
         }
--- a/src/modules/utils/player/Player.cpp
+++ b/src/modules/utils/player/Player.cpp
@@ -207,6 +207,15 @@ void Player::on_gcode_received(void *argument)
                 gcode->stream->printf("file.open failed: %s\r\n", this->filename.c_str());
             } else {
                 this->playing_file = true;
+
+                // get size of file
+                int result = fseek(this->current_file_handler, 0, SEEK_END);
+                if (0 != result) {
+                        file_size = 0;
+                } else {
+                        file_size = ftell(this->current_file_handler);
+                        fseek(this->current_file_handler, 0, SEEK_SET);
+                }
             }
 
         } else if (gcode->m == 600) { // suspend print, Not entirely Marlin compliant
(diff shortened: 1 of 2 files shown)


wolfmanjm:
Did you consider that using seek to get file size is no longer required? how about changing it to get the file size from the opendir instead?

minad:
No, I just wanted this to be fixed. I think this class should be cleaned up so or so. But I am not doing that right now.

minad:
I also don't think that it would make it better here. The issue here is just code repetition.

minad:
@wolfmanjm Ping. What about this issue? Since you touched over 82% of the lines of the file Player.cpp, I think you are the right contact person in this case :)

src/modules/utils/player/Player.cpp
```diff
@@ -135,7 +135,7 @@ void Player::on_gcode_received(void *argument)
                 // get size of file
                 int result = fseek(this->current_file_handler, 0, SEEK_END);
                 if (0 != result) {
-                    this->file_size = 0;
+                    this->file_size = -1;
```
wolfmanjm: reasons for changing this to -1?

wolfmanjm:
this should be a simple fix, not sure why you also changed the initial file size to -1 it was totally unnecessary IMO.

minad:
Yes there is a reason, just read the commit messages. I want to avoid divisons by zero.

wolfmanjm:
as far as I can see divisions by zero cannot occur as they are tested for, so i still so no reason to set it to -1

minad:
You are right. My commit messages somehow states implicitly that I did not check everywhere ;) However ftell also returns negative values in case of an error. So it is only consistent to set it to -1 in case of an fseek error. This was the reason why I changed it.

If you don't like this particular commit, then cherry-pick please.

minad:
@wolfmanjm I removed the -1 commit for you

minad:
but maybe we should still change to long instead of unsigned long to check for ftell errors?

src/modules/utils/player/Player.h
```diff
@@ -50,8 +50,8 @@ class Player : public Module {
         StreamOutput* suspend_stream;
 
         FILE* current_file_handler;
-        unsigned long file_size, played_cnt;
-        unsigned long elapsed_secs;
+        long file_size, played_cnt;
+        long elapsed_secs;
```
wolfmanjm: why is elapsed seconds long now? i mace them unsigned for a reason. i understand why file size is long bit i disagree with that too as the error should be detected at source.

minad:
now?

PR #623, opened 2015-02-21
Implemented Servo controlled probe functionality at the ZProbe module

barsamgr:
The following configuration parameters where added:

zprobe.servo_pin 1.24               # Servo PWM pin. Xmin is a good connection as the pin there is PWM
                                                  # and there is also a 5V supply for the servo
                                                  # If Xmax limit is not used, you can switch it with X
min zprobe.servo_percent 85   # Percentage of movement for deployment. Default 100
zprobe.servo_disable true        # Disable servo after a deployment or retract.
                                                 # Use if there is a lot of jittering. Default false

Also two M codes where added (same as Marlin M codes)
M401 For servo deploy
M402 For servo rectract

I did this for my own purpose, but I know that there are a lot of people looking for this functionality. I though it's good to share.

### DIFF
--- a/src/modules/tools/zprobe/ZProbe.h
+++ b/src/modules/tools/zprobe/ZProbe.h
@@ -8,11 +8,9 @@
 #ifndef ZPROBE_H_
 #define ZPROBE_H_
 
+#include <vector>
 #include "Module.h"
 #include "Pin.h"
-#include "wait_api.h"
-
-#include <vector>
 
 namespace mbed {
     class PwmOut;
--- a/src/modules/tools/zprobe/ZProbe.h
+++ b/src/modules/tools/zprobe/ZProbe.h
@@ -10,8 +10,13 @@
 
 #include "Module.h"
 #include "Pin.h"
+#include "wait_api.h"
 
 #include <vector>
+
+namespace mbed {
+    class PwmOut;
+}
 
 // defined here as they are used in multiple files
 #define zprobe_checksum            CHECKSUM("zprobe")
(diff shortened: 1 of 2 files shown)


wolfmanjm:
This should work well. However I'm thinking that the servo stuff should actually go into Switch as it is generally useful for more than just zprobe deploy/retract.

I need to think how best that would work. - Thanks

REulberg:
I'm getting ready to test this, worth noting that servo pinout goes -,+,S where the end stop pins are +,-,S
So one must remember to switch - and + on the servo plug to match the endstop pins

wolfmanjm:
endstop pins are probably a bad use for this as they have external pullup resistors and caps on them.

REulberg:
where on the board would be a good place to pull 5v@ around .4amps?

wolfmanjm:
nowhere on the board... you need to supply the 5v from an external power source.

barsamgr:
It is working fine for me using the X Min pin. I couldn't find any other 5V place and I have check through the schematic. Also there is no other free pwm pin, than X min, when you use all 5 drivers and all mosfets.

REulberg:
okay-- conceivably, if one is using the voltage regulator for logic power, you could solder a wire off of the +5
Any possible harm from doing so?

REulberg:
okay.. the m401 and M402 commands are doing nothing. as is the g32 command. I think I've done something wrong.

barsamgr:
If G32 is doing nothing, meaning not even moving Z axis, most probably you have not activated Zprobe module in the config file. Does G30 reacting?

REulberg:
it's amazing what a corrupt config will do -- I followed the warning about notepad++, but windows notepad apparently can mess things up too.

barsamgr:
:) tell me about it... I lost 2 days and couple years of my life trying to figure out what I was doing wrong when I got the board. Apparently Text Wrangler (OS X) is not config friendly. I ended up using notepad from my virtual machine windows. So far I don't have any problems.

src/modules/tools/zprobe/ZProbe.h
```diff
@@ -39,21 +42,25 @@ class ZProbe: public Module
...
     void accelerate(int c);
+    mbed::PwmOut *servo_pin;    // PWM output to regulate the laser power
 
     volatile float current_feedrate;
     float slow_feedrate;
     float fast_feedrate;
     float probe_height;
     float max_z;
+    int servo_percent;
+    bool servo_disable;
```
wolfmanjm: move into the bool struct with :1 saves memory

wolfmanjm:
This has been implemented in pull request #675

PR #624, opened 2015-02-23
gcode dispatch: send ok for M28/M29 upload gcodes

minad:
Add missing ok


wolfmanjm:
did you test this with pronterface and octoprint? It seems to have been working fine without this.

minad:
Works fine in pronterface with this patch. I didn't test octoprint.

minad:
thx :)

PR #625, opened 2015-02-23
telnet: show first prompt only when prompt is enabled

minad:
I tested it.

### DIFF
--- a/src/libs/Network/uip/telnetd/shell.cpp
+++ b/src/libs/Network/uip/telnetd/shell.cpp
@@ -224,7 +224,8 @@ int Shell::command_result(const char *str, void *p)
 /*---------------------------------------------------------------------------*/
 void Shell::start()
 {
-    telnet->output("Smoothie command shell\r\n> ");
+    telnet->output("Smoothie command shell\r\n");
+    telnet->output_prompt(SHELL_PROMPT);
 }
 
 int Shell::queue_size()


wolfmanjm:
sorry this breaks some of my scripts that wait for the initial > to start sending.
Again there is a reason I do things, removing them or changing them for no particular reason will break things you may not be aware of.

minad:
well then fix your scripts :)

What is then the prompt removal function for anyway if you don't implement it completely?

Can we not just agree that things should be implemented in the "right" way not some hacky way which stays compatible with your scripts? This is a very egoistic argument of yours. And if you don't care that people contribute then just say so.

My opinion is that things should not break, for no one using some open source tools. If you just want to stay compatible with your own hacky scripts then this is just not an argument. If you show me your script, I fix it for you.

Any maybe it is really the case that this cannot be fixed but then your argument is still invalid as long as you haven't shown me why.

wolfmanjm:
I told you why. My scripts look for > to know when the connection is complete.

minad:
Well then fix your script. There is no necessity to check for >. You can also just wait if the connection is established. It should also work. And if this is not the case and you really have to rely on the prompt why are you using the prompt flag after all in your script? Just keep prompts enabled and your scripts will work. Alternatively your script could also scan for "Smoothie...". Your argument is not a good one, sorry.

I just don't see any point in having this 0x55 prompt flag implemented when it is not implemented completely. Then you could also remove this feature alltogether.

arthurwolf:
@minad I'm pretty sure wolfmanjm has a better reason for this than just "i don't want to change two lines in my script", but he does a LOT of work in Smoothie, so he doesn't always have time to fully explain everything.
Also please don't call people "egoistic" in smoothie github issues, that's not cool, let's keep things purely technical.

minad:
Ok, sorry. Technically speaking, I didn't call him egoistic, but his argument. I don't think anyone contributing that much to open source is egoistic. But now to the problem: How do you want people to contribute here? It seems @wolfmanjm is not much open to changes to code that he wrote or touched. It seems to be ok, to add modules but changing existing code is a taboo? Also I am not doing heavy changes, only minor fixes for bugs that I observed. It is just no fun if at the end it doesn't boil down to objective arguments but to dictatorial decisions even for trivial bug fixes.

wolfmanjm:
When your "bug fixes" introduces bugs it will get rejected. These are not even fixing bugs these are making trivial, non essential changes which really serve no purpose at all. You presume that when you don't fully understand why I do something then I am at fault rather than simply your lack of understanding, so you insult me rather than try to understand the issues. If that is how you wish to proceed then really don't help anymore.
Not that it makes any difference but I have 45 years experience in embedded systems, I generally know what I am doing and why I do it.

minad:
No! I dont want to insult you. If I criticize this is not meant as an insult. Rather you are insulting me with what you say.

You see that I am capable of writing some more complex code as seen in the other pull requests. And now you say I don't try to understand, lack of understanding...

I think you agree that the m32 bug was a bug? This thing here however is undocumented behaviour, there it is not so clear.

I also see that you are capable since you wrote a lot here, independent of how long you work wherever on whatever. So let's stop this bullshit.

But please allow people to review your code, give hints for improvements, etc.

Am 24. Februar 2015 01:12:17 MEZ, schrieb Jim Morris <notifications@github.com>:

bgamari:
For what it's worth, it seems to me that the fact that the telnet interface emits a prompt --- and only one prompt --- despite prompting being disabled is a bit surprising. Admittedly it is a bit strong to claim the behavior is a bug but I also think it's a bit much to suggest that resolving this inconsistency will introduce a bug; to the contrary I think it makes the semantics of the project more clear. In my opinion a change that replaces inconsistent behavior with clear uniform semantics is desirable, especially when the breakage that it causes is so easily detected and fixed.

wolfmanjm:
Yea I have merged the bug fixes where you have fixed the bugs, and I appreciate those bug fixes.
But this is not a bug, it is a feature I specifically put in there for a reason.
If you need to know every detail I will will explain in this case, but I don't have time to explain every single detail everytime. In future if I say it is not a bug accept it.

In this specific case if you connect to the telnet port using python and the 
telnetlib, and you want to turn off prompts you have to wait for the first signin string then you turn off the prompts. I don't want to look for the entire signon string just the very last character which is '>'
There is a race condition where you cannot send the turn off prompts immediately. so you have to do this...

```
import getpass
import sys
import telnetlib

def write_raw_sequence(tn, seq):
    sock = tn.get_socket()
    if sock is not None:
        sock.send(seq)

HOST = "192.168.3.155"

tn = telnetlib.Telnet(HOST)
print tn.read_until("> ")

# turn off prompt
write_raw_sequence(tn, telnetlib.IAC + telnetlib.DONT + "\x55")

# get pos
tn.write("M114\n")
print tn.read_until("ok")

# turn on prompt
write_raw_sequence(tn, telnetlib.IAC + telnetlib.DO + "\x55")

tn.write("version\n")
print tn.read_until("> ")

tn.write("exit\n")
print tn.read_all()
```

wolfmanjm:
There are a lot of places in Smoothie where things could be made more consistent but for legacy reasons they stay the same.
This IMO is not inconsistent the > is part of the signon message not the prompt, it could be % for all i care, but for legacy reasons it is what it is now.
and FYI this is NOT easily fixed in the script for the race condition reasons I have mentioned.

minad:
@bgamari You said it in the right words. This is what I was thinking. I sought for consistency.

@wolfmanjm Thx for the long answer. I appreciate it.

* Changing something from > to % is not a bug and I would not suggest such a change, I can assure you. At least not after all these discussions that we had ;) Purely a matter of taste is clearly no reason. But changing > to % in between would seem even more inconsistent.

* Race condition: If the IAC is sent to late, then the prompt will be shown anyways. Is there a more serious problem than that? I also made some script which does exactly what you do, but I send the string immediately since I did not know. I haven't seen problems so far. Alternatively one could wait for the client to sent an IAC and sent the propmt only then or after receiving the first data (being IAC or not) from the client.

* Now my most important point. I think you should change your opinion concerning "I don't want to explain things, since I am doing those for a reason"! I tell you five reasons why you should explain!

1. Through explaining you documented an exisiting problem like a race condition. This should be documented either in code or here in an issue.

2. You help people understand. You have a lot of experience. I have maybe a third of your experience in years and if you know things that I don't, I would like to understand them. Helping others is probably already your interest if you are contributing to open source, but if you also help others becoming better you are doing even more, since you are enabling them.

3. You participate in a discussion which will lead to better understanding, critical review of the code and at the very end in something better for the project. I think you are interested in that.

4. If you don't explain you hinder other people in contributing. Not explaining might save time now, but in the long run the other strategy wins, since more people contribute and fix bugs for you.

5. You might improve your own coding through discussions and might gain new insights (Not in this issue I admit). For me this has happened (also not in this discussion), but if you allow people to criticize something interesting comes out at the end, maybe.

PR #628, opened 2015-02-26
Zgrid probing fixes

sorki:
Works without endstops module.

### DIFF
--- a/src/modules/tools/zprobe/ZGridStrategy.cpp
+++ b/src/modules/tools/zprobe/ZGridStrategy.cpp
@@ -106,13 +106,13 @@ ZGridStrategy::ZGridStrategy(ZProbe *zprobe) : LevelingStrategy(zprobe)
     this->cal[Z_AXIS] = 30.0f;
 
     this->in_cal = false;
+    this->pData = nullptr;
 }
 
 ZGridStrategy::~ZGridStrategy()
 {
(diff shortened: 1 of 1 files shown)


src/modules/tools/zprobe/ZGridStrategy.cpp
```diff
@@ -434,41 +442,41 @@ bool ZGridStrategy::doProbing(StreamOutput *stream)  // probed calibration
...
-                this->cal[Y_AXIS] = this->bed_y/2.0f;
-                this->cal[Z_AXIS] = this->bed_z/2.0f;    // Position head for probe removal
-            } else {
-                this->next_cal();                        // to not cause damage to machine due to Z-offset
-            }
-            this->move(this->cal, slow_rate);            // move to the next position
-
-            this->pData[pindex] = z ;                    // save the offset
+        if (probes == (probe_points-1)){
+            this->cal[X_AXIS] = this->bed_x/2.0f;    // Clear calibration position
```
wolfmanjm: I didn't notice befire but this does not support beds where 0,0 is center of bed

wolfmanjm:
Found several killer bugs in the current code...
1. needs to support bots that have 0,0 as center of bed
2.  while(!zprobe->getProbeStatus()); <---- This is a big NONO, must call ON_IDLE this will kill USB.

PR #632, opened 2015-02-28
Add option to oversample thermistor ADC values.

PetteriAimonen:
This improves the stability of PID control by reducing the effect of quantization
on the operation of the D term.

Pull request #564 rebased against newest edge.
See previous pull request for result graphs from both me and @minad, both of which show large improvement in stability.

I see no reason not to include this as an option. It is off by default and only takes 8 bytes ram per thermistor, no other overhead when disabled.

### DIFF
--- a/src/modules/tools/temperaturecontrol/Thermistor.cpp
+++ b/src/modules/tools/temperaturecontrol/Thermistor.cpp
@@ -15,6 +15,7 @@
 #include "libs/Median.h"
 #include "utils.h"
 #include "StreamOutputPool.h"
+#include "SlowTicker.h"
 
 // a const list of predefined thermistors
 #include "predefined_thermistors.h"
@@ -37,7 +38,7 @@
 #define rt_curve_checksum                  CHECKSUM("rt_curve")
 #define coefficients_checksum              CHECKSUM("coefficients")
 #define use_beta_table_checksum            CHECKSUM("use_beta_table")
-
+#define oversample_freq_checksum           CHECKSUM("oversample_freq")
 
 Thermistor::Thermistor()
 {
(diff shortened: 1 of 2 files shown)


minad:
@wolfmanjm @arthurwolf Any news on this pr?

arthurwolf:
Haven't had time to look into this in detail enough to answer, but it's on
my todo list.
Also I think there is some other code planned that would be an even more
elegant solution to the problem this adresses, so have to look into that
too before I can make an interesting answer :)

On Tue, Mar 3, 2015 at 1:32 PM, Daniel Mendler <notifications@github.com>
wrote:

> @wolfmanjm <https://github.com/wolfmanjm> @arthurwolf
> <https://github.com/arthurwolf> Any news on this pr?
>
> —
> <https://github.com/Smoothieware/Smoothieware/pull/632#issuecomment-76937082>
> .
>



-- 
Courage et bonne humeur.

wolfmanjm:
This is no different from the last one, which I already explained why we won't accept this. Oversampling is a good idea, this implementation is not a good one.

arthurwolf:
I know @minad is going to ask so I'll just answer with a bit more detail right away :)

In the same way we currently tend to reject modifications to the step generation code because a big refactor of it is on-going, this is not accepted mostly because a much better way to do the same thing has already been described and is planned to be added.

PetteriAimonen:
@wolfmanjm Yeah, pretty much what I expected. I would have appreciated if you had replied to my direct question earlier.

@arthurwolf Is the general policy of smoothie project "eventualistic", i.e. don't accept partial fix if eventually there will be a perfect one? When I bought the board, I expected it would work as promised when I bought it, not 5 years laters.

wolfmanjm:
You know you really should stop taking pull request rejections personally. As a maintainer of an open source project I feel it is my responsibility to accept and reject code based on what the project needs and what the future plans are. We are under no obligation whatsoever to accept pull requests. People like you who get all upset about rejections (which are not personal) make life really hard for maintainers. I see other maintainers having similar issues with certain people.

I already explained my objections in the previous pull request, this one was identical to that one with no changes whatsoever. What did you expect?

Once a pull request is accepted we become responsible for maintaining that code, and it also becomes legacy so if we want to do it a different way later on it becomes much harder especially if people are using the legacy code.
I did not feel that I wanted to maintain or fix this code, and I have discussed with the other primary contributors several alternatives which we feel is much better than this, but incompatible with it.

As this is open source you are free to maintain your own branch and do whatever you wish in that branch.
Complaining to the world about how mistreated you are in this project is just pathetic in my opinion.
I would certainly appreciate it if you did not contribute anymore. the other maintainers may not feel that way, but you have caused me enough irritation, and I really do not want to have to deal with you anymore.

PetteriAimonen:
@wolfmanjm Where have I taken it personally? Only thing I admit I have sometimes been unprofessional about is discussions with you, because "someone is wrong on the internet!" effect - I've tried to avoid them on IRC, and I think I've behaved myself on pull requests since.

Only reason I made this pull request was because @minad asked, as you can see in the previous pull request. I hoped that easily mergeable immediate fix for an issue that affects multiple users would be accepted.

But ok, I guess I have no hope of contributing anymore. Hopefully I won't ruin my relations with MachineKit project as badly.

PR #633, opened 2015-02-28
Trigger your camera (IRemitter module)

lautr3k:
Hi,
Here is a module to trigger your camera remotely, the goal is to create fluid timelapse...
An online tool to add trigger commands (Mxxx) is available here : http://www.onlfait.ch/GCodeTimeLapse.js/

Supported device : canon, canonwldc100, minolta, nikon, olympus, pentax, sony.
Tested on canon D5100, but I need you to test other models and improve the parser.

Peace.


minad:
thats cool :+1: however it seems to me there is a lot of code duplication. wouldn't it be possible to write this trigger sequence and timings somehow in the config file?

src/modules/tools/iremitter/Iremitter.cpp
```diff
@@ -0,0 +1,294 @@
...
+*/
+
+// pwm at freqency for x microseconds
+void Iremitter::high_us(unsigned int us, int freq)
+{
+	uint32_t start = us_ticker_read();
+	int pause = (1000 / freq / 2) - 4;
+
+	while((us_ticker_read() - start) < us)
+	{
```
wolfmanjm: I think you need to use a timer for this... busy waits are not very friendly to the rest of the code.

wolfmanjm:
You have to remove the busy waits, they are not allowed in the code for anything over say 1us
It locks up the entire system when you do that, and USB will disconnect. At a minimum you have to call on_idle_loop in the busy wait. But really to do it properly you need to use a timer to get the pulse width correct.

wolfmanjm:
Also you should not have merges in the patch, you need to do a clean rebase. and your branch should be called feature/iremitter or something as per the wiki...  http://smoothieware.org/github

PR #634, opened 2015-03-03
IRemitter module V2

lautr3k:
In response to the discussion on #633 I have made a new, more flexible and less blocking module.
Hope it is better, peace.


wolfmanjm:
This is out of date and needs to be updated. If it is fixed I'll reopen

PR #641, opened 2015-03-15
Master

mercurywaters:


wolfmanjm:
needs a description, and needs to be only the patch not the merges. this will not merge cleanly.
Fix it and resubmit please. Please follow the guidelines for submitting patches in the wiki.

wolfmanjm:
this also needs to be in a branch from edge not master.

PR #643, opened 2015-03-20
support of configurable slave motors added. 

ytkachov:
any of primary motor (alpha, beta or gamma) can have slave motor which does exactly the same movements as master motor. may be used in shapeoko CNC for Y axes for example.

### DIFF
--- a/ConfigSamples/Snippets/beta_slave_motor.config
+++ b/ConfigSamples/Snippets/beta_slave_motor.config
@@ -1,0 +1,5 @@
+#beta (Y) motor has slave motor controlled by delta pins
+beta_has_slave                               true             # Y axis second motor controlled by gamma driver
+beta_slave_step_pin                          2.3              # Pin for second Y stepper step signal
+beta_slave_dir_pin                           0.22             # Pin for second Y stepper direction
+beta_slave_en_pin                            0.21             # Pin for second Y enable
--- a/src/libs/StepperMotor.cpp
+++ b/src/libs/StepperMotor.cpp
@@ -17,12 +17,19 @@ float StepperMotor::default_minimum_actuator_rate= 20.0F;
 
 // A StepperMotor represents an actual stepper motor. It is used to generate steps that move the actual motor at a given speed
 
-StepperMotor::StepperMotor()
+StepperMotor::StepperMotor() : has_slave(false)
 {
     init();
 }
 
-StepperMotor::StepperMotor(Pin &step, Pin &dir, Pin &en) : step_pin(step), dir_pin(dir), en_pin(en)
+StepperMotor::StepperMotor(Pin &step, Pin &dir, Pin &en) : step_pin(step), dir_pin(dir), en_pin(en), has_slave(false)
+{
+    init();
+    enable(false);
+    set_high_on_debug(en.port_number, en.pin);
+}
+
+StepperMotor::StepperMotor(Pin &step, Pin &dir, Pin &en, Pin &slave_step, Pin &slave_dir, Pin &slave_en) : step_pin(step), dir_pin(dir), en_pin(en), slave_step_pin(slave_step), slave_dir_pin(slave_dir), slave_en_pin(slave_en), has_slave(true)
 {
     init();
     enable(false);
@@ -68,6 +75,7 @@ void StepperMotor::step()
 
     // output to pins 37t
     this->step_pin.set( 1 );
+    if(has_slave) this->slave_step_pin.set( 1 );
 
     // move counter back 11t
     this->fx_counter -= this->fx_ticks_per_step;
@@ -131,6 +139,7 @@ void StepperMotor::update_exit_tick()
 StepperMotor* StepperMotor::move( bool direction, unsigned int steps, float initial_speed)
 {
     this->dir_pin.set(direction);
+    if(has_slave) this->slave_dir_pin.set(direction);
     this->direction = direction;
 
     // How many steps we have to move until the move is done
--- a/src/libs/StepperMotor.h
+++ b/src/libs/StepperMotor.h
@@ -20,12 +20,21 @@ class StepperMotor {
     public:
         StepperMotor();
         StepperMotor(Pin& step, Pin& dir, Pin& en);
+
+        // slave motor pins present
+        StepperMotor(Pin& step, Pin& dir, Pin& en, Pin& slave_step, Pin& slave_dir, Pin& slave_en);
         ~StepperMotor();
 
         void step();
-        inline void unstep() { step_pin.set(0); };
+        inline void unstep() { 
+            step_pin.set(0); 
+            if(has_slave) slave_step_pin.set(0);
+        };
 
-        inline void enable(bool state) { en_pin.set(!state); };
+        inline void enable(bool state) { 
+            en_pin.set(!state); 
+            if(has_slave) slave_en_pin.set(!state);
+        };
 
         bool is_moving() { return moving; }
         void move_finished();
@@ -66,12 +75,16 @@ class StepperMotor {
     private:
         void init();
 
+        bool has_slave;
         int index;
         Hook* end_hook;
 
         Pin step_pin;
         Pin dir_pin;
         Pin en_pin;
+        Pin slave_step_pin;
+        Pin slave_dir_pin;
+        Pin slave_en_pin;
 
         float steps_per_second;
         float steps_per_mm;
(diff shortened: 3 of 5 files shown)


wolfmanjm:
Sorry we are not accepting any pull requests for the core kernel at the moment due to an upcoming major refactor.

Please add a link to your fork and a description to this page so others who may need this can use it if they wish..
http://smoothieware.org/third-party-branches

Although I personally prefer the hardware method of accomplishing this particular hack and feel a S/W method is not needed, one of the other core developers may consider merging this at a later date.

Thank you.

wolfmanjm:
As way of discussion in case someone else is interested in this patch I'd also like to add some comments about it.

I do not like the extra memory this uses regardless of whether any motor has slaves.. there are three extra pins per stepper motor whether they are needed or not, this memory usage adds up.
I am also uneasy about the additional work required in a  time critical ISR function to test for, and potentially step, a slave motor. Again the test is required whether or not any slave motors are defined, which in the majority of cases they will not be.

The H/W hack currently used requires no extra memory or processing time.
I am not sure if this could be done in S/W without using extra memory when not in use or extra processing time, but if it could it would be far preferable.

qharley:
I second rather using hardware to do the same.  Missing config could case
serious damage to the system for instance.


  [image: photo]
*Quentin Harley*
Founder, Reprap Morgan 3D Printers
 p:012 004 1883 | m:082 782 3902 | e:quentin@reprapmorgan.com | w:
www.reprapmorgan.com | a: 4 Burger Ave, Lyttelton Manor, Centurion, South
Africa

ytkachov:
Thank all of you for your comments. There were at least two reasons why I decided to do it in software. First of all I am a programmer and do believe that everything which could be done in software must be done in software. The second was that there is such improvement suggested in Smoothie contest page. 

I agree that we could avoid using extra memory and unnecessary test whether or not any slave motors are defined. It is easy. We just need virtual methods and derived class for PairedStepperMotor.

arthurwolf:
Hey !

There is actually a more general way to do this that is useful not only for
axes but also for all other modules : the multipin features, which allows
you to configure pins as "1.31+1.30" for example

Somebody coded it here :
https://github.com/Smoothieware/Smoothieware/compare/edge...gecgooden:fix/slaveaxis

But he doesn't have a Smoothieboard to test it, and when I tested it it did
not work.
If somebody wants to test it and debug it to find what's wrong, it'd be
awesome !

Cheers !


On Sun, Mar 22, 2015 at 12:40 PM, ytkachov <notifications@github.com> wrote:

> Thank all of you for your comments. There were at least two reasons why I
> decided to do it in software. First of all I am a programmer and do believe
> that everything which could be done in software must be done in software.
> The second was that there is such improvement suggested in Smoothie contest
> page.
>
> I agree that we could avoid using extra memory and unnecessary test
> whether or not any slave motors are defined. It is easy. We just need
> virtual methods and derived class for PairedStepperMotor.
>
> —
> <https://github.com/Smoothieware/Smoothieware/pull/643#issuecomment-84589676>
> .
>



-- 
Courage et bonne humeur.

ytkachov:
I can try to understand what is wrong with that solution. It really looks nice and elegant.

ytkachov:
BTW, Arthur what is going on with Smoothie Contest? Looks like page was not updated since long ago...

arthurwolf:
I got many more entries for the contest than I expected, so it's taking me
a long time to process them all, there should be news soon.

On Sun, Mar 22, 2015 at 1:06 PM, ytkachov <notifications@github.com> wrote:

> BTW, Arthur what is going on with Smoothie Contest? Looks like page was
> not updated since long ago...
>
> —
> <https://github.com/Smoothieware/Smoothieware/pull/643#issuecomment-84593278>
> .
>



-- 
Courage et bonne humeur.

ytkachov:
Hi Arthur,

I have investigated source codes of 

https://github.com/Smoothieware/Smoothieware/compare/edge...gecgooden:fix/slaveaxis

Though idea is very elegant its implementation is completely incorrect. It appears that this approach just can't work as it was supposed to work. Besides strange things like this:

    Pin* Pin::from_string(std::string value) {
        SinglePin* s_pin = new SinglePin();  // new SinglePin gets created
        s_pin->from_string(value); 	         // and initialized from string
                                             // and... nothing...
        return this;
    }

the main problem is that in many classes (StepperMotor, Extruder, ReprapDiscountGLCD and others) Pin is used as an *object*, not a *pointer to object*. As a result it can't be "virtualized" that is we can't use DoublePin as Pin. There are three possible solutions:

1) each Pin has information about two physical pins and some property which tell us if it is Single or Double. It is easy to implement but leads to waste of memory in all cases while Double Pins will be used quite rare.

2) Replace all Pin object to pointers to Pin. Though it requires changes in many places this approach is most robust. Additional memory is used when it is really required, correspond to general coding rules (for example in the Robot class there is pointer to StepperMotor objects while Pins are present as objects which means that header must be included in header). Drawbacks of this way is: a) using of inline functions becomes impossible b) call of virtual method requires additional processor time which may be critical.

3) Pin object is thin wrapper around internal pin_impl which is stored as pointer. pin_impl could be SinglePin or DoublePin depending on initialization string. In this case no changes required in those classes which use Pin obects but again more processor time waste on virtual method calls and additinal pointer is stored in Pin object.

_Which approach is preferable? Sorry for asking but having 20+ years of C++ Desktop and server development I am quite new in embedded programming so I need some advice to make it correct._

Thanks

arthurwolf:
Hey !

I think the best method here would be number 2.
But it needs to be tested for performance once implemented, as we can't
afford to loose too much time on some operations.

Cheers !


On Wed, Mar 25, 2015 at 1:20 PM, ytkachov <notifications@github.com> wrote:

> Hi Arthur,
>
> I have investigated source codes of
>
> edge...gecgooden:fix/slaveaxis
> <https://github.com/Smoothieware/Smoothieware/compare/edge...gecgooden:fix/slaveaxis>
>
> Though idea is very elegant its implementation is completely incorrect. It
> appears that this approach just can't work as it was supposed to work.
> Besides strange things like this:
>
> Pin* Pin::from_string(std::string value) {
>     SinglePin* s_pin = new SinglePin();  // new SinglePin gets created
>     s_pin->from_string(value);           // and initialized from string
>                                          // and... nothing...
>     return this;
> }
>
> the main problem is that in many classes (StepperMotor, Extruder,
> ReprapDiscountGLCD and others) Pin is used as an *object*, not a *pointer
> to object*. As a result it can't be "virtualized" that is we can't use
> DoublePin as Pin. There are three possible solutions:
>
> 1) each Pin has information about two physical pins and some property
> which tell us if it is Single or Double. It is easy to implement but leads
> to waste of memory in all cases while Double Pins will be used quite rare.
>
> 2) Replace all Pin object to pointers to Pin. Though it requires changes
> in many places this approach is most robust. Additional memory is used when
> it is really required, correspond to general coding rules (for example in
> the Robot class there is pointer to StepperMotor objects while Pins are
> present as objects which means that header must be included in header).
> Drawbacks of this way is: a) using of inline functions becomes impossible
> b) call of virtual method requires additional processor time which may be
> critical.
>
> 3) Pin object is thin wrapper around internal pin_impl which is stored as
> pointer. pin_impl could be SinglePin or DoublePin depending on
> initialization string. In this case no changes required in those classes
> which use Pin obects but again more processor time waste on virtual method
> calls and additinal pointer is stored in Pin object.
>
> *Which approach is preferable? Sorry for asking but having 20+ years of
> C++ Desktop and server development I am quite new in embedded programming
> so I need some advice to make it correct.*
>
> Thanks
>
> —
> <https://github.com/Smoothieware/Smoothieware/pull/643#issuecomment-86002549>
> .
>



-- 
Courage et bonne humeur.

wolfmanjm:
The trouble is making pins virtual will also add more memory overhead as that will make the vtable bigger for everything that uses a pin. Plus there will be a pointer to a pin plus the actual pin itself.
Why do we need this when there is a perfectly acceptable way to do it by wiring?

arthurwolf:
On Wed, Mar 25, 2015 at 7:59 PM, Jim Morris <notifications@github.com>
wrote:

> The trouble is making pins virtual will also add more memory overhead as
> that will make the vtable bigger for everything that uses a pin. Plus there
> will be a pointer to a pin plus the actual pin itself.
> Why do we need this when there is a perfectly acceptable way to do it by
> wiring?
>

It's a feature lots of people have been asking for.
But if it can't be done without a big RAM cost, it shouldn't be done.


> —
> <https://github.com/Smoothieware/Smoothieware/pull/643#issuecomment-86173592>
> .
>



-- 
Courage et bonne humeur.

PR #648, opened 2015-03-30
Update predefined_thermistors.h

Fourmi:
Add "Hisens 104F400 " thermistor steinhart value for hotend "Simple ONE" et "All In ONE"


wolfmanjm:
I think this is misnamed, does not appear to be hisens specific. I also do not fond this to be common,y available. please provide a link to a common reseller.
 did find reference to a 104F4000 but does not mention hisens.

Fourmi:
Ok,i see why Hisens is not appear because there are changed name .
Now it's Musor 
You can read the specs of thermistor here >https://drive.google.com/file/d/0B0eJVP0Swfx5WXczQ1QwZ09Kd2s/view?usp=sharing

Thx arthur

wolfmanjm:
then rename it it NTC104F4000

Fourmi:
Ok,i renamed it
Thx

arthurwolf:
Is this clean enough to be merged ? If so tell me so I add the thermistor to the documentation.

wolfmanjm:
as far as i can see it is a very rare thermistor and no one uses it, so it didn't seem worth adding. Especially as anyone actually having one and using it could simply add the data in config.

PR #650, opened 2015-04-06
Not showing hidden files

JesperKrogPoulsen:
OS X leaves a lot of hidden files on the SD card, whether files are written directly to SD card or through USB. This is cluttering the panel and since the hidden files are of no use for Smoothie, I have added code to filter them out. This includes directories like .Trashes. I can't think of a reason why people would want to name their gcode files with a '.' in the beginning, so I don't think it will harm anyone. If it does, it can be customised using the config file, but this will add to the size of the code with little value.

### DIFF
--- a/src/modules/utils/panel/screens/FileScreen.cpp
+++ b/src/modules/utils/panel/screens/FileScreen.cpp
@@ -116,13 +116,14 @@ void FileScreen::clicked_line(uint16_t line)
     }
 }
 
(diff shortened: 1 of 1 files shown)


WZ9V:
There is some software for OSX (DriveCleaner I think) that will mop up that stuff on eject on from the menu.  I use it for my SD Cards with the Mac.

PR #651, opened 2015-04-07
Scara transtrim

qharley:
This addition to the Scara code will translate the calibration home trims to the actual homing offsets.  This is useful where the initial guess was not good, or where the initial steps per degree are too far out.

The effect is that the arm remains at the home position after homing, and do not potentially move beyond the homing safe zone.


wolfmanjm:
somthign went very wrong with this, 111 commits? I think you need to do a rebase and try again.

src/modules/robot/Robot.cpp
```diff
@@ -383,6 +384,9 @@ void Robot::on_gcode_received(void *argument)
         }
     } else if( gcode->has_m) {
         switch( gcode->m ) {
+            case 0:  // M0 - Pause until pause button pressed again
```
wolfmanjm: Absolutely not :)

PR #662, opened 2015-05-01
Switch momentary output

691175002:
This is a small addition to the switch module.

A new configuration parameter has been added: output_pin_behavior, which is analogous to input_pin_behavior.

Setting "output_pin_behavior = toggle" preserves existing behavior, and is the default value.
Setting "output_pin_behavior = momentary" will instead output a pulse when the "on" g-code is received.

The pulse duration is configurable using the "output_pulse_duration" parameter and by default is 0.2 seconds.  This feature is useful for performing all sorts of miscellaneous tasks as the pulse duration can be configured to last minutes or even hours.

I have also implemented a change to how Switch.cpp was reading the configuration.  Details here:
https://groups.google.com/d/msg/smoothie-dev/hbjUP5UwMeM/L2SlhbxYPGIJ

I have tested the momentary output feature and fixed configuration, but have not tested the input modes.

This is my first time contributing to an open source project so I have no idea what I'm doing.

### DIFF
--- a/src/modules/tools/switch/Switch.cpp
+++ b/src/modules/tools/switch/Switch.cpp
@@ -37,8 +37,7 @@
 #define    max_pwm_checksum             CHECKSUM("max_pwm")
 #define    output_on_command_checksum   CHECKSUM("output_on_command")
 #define    output_off_command_checksum  CHECKSUM("output_off_command")
-#define    output_momentary_delay_checksum CHECKSUM("output_momentary_delay")
-
+#define    output_pulse_duration_checksum CHECKSUM("output_pulse_duration")
 Switch::Switch() {}
 
 Switch::Switch(uint16_t name)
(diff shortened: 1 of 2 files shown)


logxen:
I haven't fully reviewed the code ... but I would recommend changing the setting to "pulse" rather than "momentary". (momentary has a slightly different meaning in context)

wolfmanjm:
I am not too keen on it using 10 bytes per switch whether or not this is enabled. This seems of marginal general use and wasting that kind of memory is not a good trade off.

691175002:
Since there are only two output modes and two input modes they can be packed into the bitfield.  The counter/delay can be changed to uint16 which puts the maximum pulse duration at 18 hours and saves an additional 4 bytes.

logxen:
O also you could use the switch's value as the duration. That would save one of the int16s and and make it so the output duration would be set with an Scode. :)

691175002:
That seems like a pretty good idea - I'll give it a go.

WZ9V:
Not sure what I would use the function for but if it can be added with minimal impact it might come in handy.

WZ9V:
Walking the dogs, I thought of something it would be very useful for.  You could have Slic3r insert an on layer change G-Code to flip this switch which could be routed to an input pin on a BBB or RPi with a camera to take pictures every so often on layer changes.  In that situation you would really just want it to be a pulse that triggers the input.  I agree that pulse would be better than momentary for a name.

691175002:
I've implemented both suggestions.

The mode has been renamed to "pulse" and included as an output_type (along with digital and pwm) instead of introducing a new configuration parameter.  

The pulse output_type will set the output high for 0.2s when enabled.  A custom pulse duration can be specified using an Scode in seconds.
I tried to implement reasonable behaviour when the switch is configured as both an input and pulse output.  Such a configuration is unlikely to ever be useful, but it works.

The switch configuration now uses boolean flags instead of storing checksums.  This also fixes a related issue in the edge branch where "input_pin_behavior momentary" will actually configure the switch in toggle mode.

wolfmanjm:
Good job at saving memory :)

Have you  tested all the other states? as this pretty much touches everything you need to test pwm toggle etc for input and output. not just the mode you implemented. - Thanks

691175002:
I've tested every combination of input_pin_behavior and output_type.  So far everything works as it should.

wolfmanjm:
cannot be merged as it has conflicts. Reopen when fixed

Issue #669, opened 2015-06-07
Crash when issuing commands from panel while another command is executing


JesperKrogPoulsen:
When running a long command, like homing, from the panel, Smoothieware will crash if other commands are selected from the panel while the homing is executed.

The problem is, that PanelScreen::on_main_loop iterates over command_queue, which is a vector. When  new commands are queued from the panel, the iterator is invalidated.

wolfmanjm:
more generally the problem is that panel commands are issued during on_idle, but executed in on_main_loop, under normal conditions new commands cannot be issued until the previous ones have been executed. Homing issues on_idle() while it waits, hence this bug. This may be a more general bug that needs to be looked into when commands are issued from other sources such as network or USB serial.

PR #670, opened 2015-06-07
Fix/invalid iterator

JesperKrogPoulsen:
The command_queue can get modified if commands are added during a long-running command, like G28. This invalidates the iterator and results in a crash. Switched to a deck as agreed with Wolfmanjm


wolfmanjm:
fixes #669 
Thanks

PR #673, opened 2015-06-10
Added RotatableDelta robot arm solution

DouglasPearless:
new file:   ConfigSamples/FirePick.delta/config.FirePickSmoothieOnABreadboard
	modified:   src/modules/robot/Robot.cpp
	new file:   src/modules/robot/arm_solutions/RotatableDeltaSolution.cpp
	new file:   src/modules/robot/arm_solutions/RotatableDeltaSolution.h


src/modules/robot/arm_solutions/RotatableDeltaSolution.cpp
```diff
@@ -0,0 +1,305 @@
...
+	//the same, BUT NOT ZERO.
+
+	init();
+}
+
+// inverse kinematics
+// helper functions, calculates angle theta1 (for YZ-pane)
+int RotatableDeltaSolution::delta_calcAngleYZ(float x0, float y0, float z0, float &theta)
+{
+    float y1 = -0.5 * tan30 * delta_f; // f/2 * tan 30
```
wolfmanjm: All constants need to be appended with F otherwise all calculations are done in double as constants without F are double. Thi sneeds to be done in several lines following this one.

src/modules/robot/arm_solutions/RotatableDeltaSolution.cpp
```diff
@@ -0,0 +1,305 @@
...
+    y0 -= 0.5 * tan30       * delta_e; // shift center to edge
+    // z = a + b*y
+    float a = (x0*x0 + y0*y0 + z0*z0 + delta_rf*delta_rf - delta_re*delta_re - y1*y1) / (2.0*z0);
+    float b = (y1-y0)/z0;
+
+    float d = -(a+b*y1)*(a+b*y1) + delta_rf*(b*b*delta_rf+delta_rf); // discriminant
+    if (d < 0) return -1;                                            // non-existing point
+
+    float yj = (y1 - a*b - sqrt(d))/(b*b + 1.0);                     // choosing outer point
```
wolfmanjm: needs to be sqrtf()

src/modules/robot/arm_solutions/RotatableDeltaSolution.cpp
```diff
@@ -0,0 +1,305 @@
...
+    float zj = a + b*yj;
+
+    theta = 180.0F*atan(-zj/(y1 - yj))/pi + ((yj>y1)?180.0:0.0);
```
wolfmanjm: needs to be atanf()

src/modules/robot/arm_solutions/RotatableDeltaSolution.cpp
```diff
@@ -0,0 +1,305 @@
...
+int RotatableDeltaSolution::delta_calcForward(float theta1, float theta2, float theta3, float &x0, float &y0, float &z0)
+{
+     float t = (delta_f-delta_e)*tan30/2.0F;
+     float degrees_to_radians = pi/180.0F;
+
+     theta1 *= degrees_to_radians;
+     theta2 *= degrees_to_radians;
+     theta3 *= degrees_to_radians;
+
+     float y1 = -(t + delta_rf*cos(theta1));
```
wolfmanjm: needs to be cosf(), ditto for all the following functions

src/modules/robot/arm_solutions/RotatableDeltaSolution.h
```diff
@@ -0,0 +1,53 @@
...
+        //float tool_offset;	// Distance between end effector ball joint plane and tip of tool (Z probe)
+        float tool_offset;		// Distance between end effector ball joint plane and tip of tool (PnP)
+        float z_calc_offset;
+
+        float z_home_angle;		// This is the angle where the arms hit the endstop sensor
+        float z_home_offs;		// This is calculated from the above angle, after applying forward
+        						// kinematics, and adding the Z calc offset to it.
+
+        // Print surface diameter/2 minus unreachable space (avoid collisions with vertical towers).
+        float delta_printable_radius;
```
wolfmanjm: what are all these? please remove any unused variables.

wolfmanjm:
Several issues with optimizations, please fix and then we can merge. You just push to the same branch on your github and it will automatically be added here do not push a new pull request.

wolfmanjm:
I think you forgot to add the selection of the arm solution to robot.cpp, did you test this? I cannot merge something that is not thoroughly tested.

DouglasPearless:
I think the robot.cpp file was not picked up not the README file in the config directory which I make sure is sent when I fix the coding issues

wolfmanjm:
robot.cpp was there it just had the #include but nothing in the arm solution selection switch statement

src/modules/robot/arm_solutions/RotatableDeltaSolution.cpp
DouglasPearless: OK, looks like I copied the wrong file.

I will do a clean compile after your recommended changes, load the firmware onto the FPD and check that everything works before I update git.

Cheers
Douglas

DouglasPearless:
Hi Jim,

I think I have fixed everything, I have done a clean compile, installed the firmware and it all works properly.

I assume by updating my repo, it will have also sent you an updated request (or several)?

If not, I can submit again.

Cheers
Douglas

wolfmanjm:
yes just push to your github btsnch

DouglasPearless:
I think I have managed to push my branch.

src/modules/robot/Robot.cpp
```diff
@@ -174,16 +180,16 @@ void Robot::on_config_reload(void *argument)
     }
 
 
-    this->feed_rate           = THEKERNEL->config->value(default_feed_rate_checksum   )->by_default(  100.0F)->as_number();
-    this->seek_rate           = THEKERNEL->config->value(default_seek_rate_checksum   )->by_default(  100.0F)->as_number();
-    this->mm_per_line_segment = THEKERNEL->config->value(mm_per_line_segment_checksum )->by_default(    0.0F)->as_number();
-    this->delta_segments_per_second = THEKERNEL->config->value(delta_segments_per_second_checksum )->by_default(0.0f   )->as_number();
-    this->mm_per_arc_segment  = THEKERNEL->config->value(mm_per_arc_segment_checksum  )->by_default(    0.5f)->as_number();
-    this->arc_correction      = THEKERNEL->config->value(arc_correction_checksum      )->by_default(    5   )->as_number();
+    this->feed_rate           = THEKERNEL->config->value((unsigned int)default_feed_rate_checksum   )->by_default(  100.0F)->as_number();
```
wolfmanjm: please do nopt make arbitrary uneeded changes to the source code that is not related to your checkins. These need to be reverted.

wolfmanjm:
For some reason you added unsigned int casts to a bunch of lines which is not needed or wanted.

src/modules/robot/Robot.cpp
```diff
@@ -356,6 +362,7 @@ void Robot::on_gcode_received(void *argument)
                 }
                 gcode->mark_as_taken();
             }
+            break; //added by DHP to suppress warning 20150324
```
wolfmanjm: what is this needed for? again do not make unwanted changes.

wolfmanjm:
I am sorry but I reject pull requests that arbitrarily add things like this. they are not required and just confuse the pull request. If you were getting warnings then you were not using the required tool chain.

DouglasPearless:
The compiler warned that there was no break at the end of the statement.

(From memory) The unsigned int additions were again to remove compiler warnings.

I will remove them and resubmit :-)

src/modules/robot/arm_solutions/RotatableDeltaSolution.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+#ifndef beta_steps_per_mm_checksum
+  #define beta_steps_per_mm_checksum		CHECKSUM("beta_steps_per_mm_checksum");
+#endif
+
+#ifndef gamma_steps_per_mm_checksum
+  #define gamma_steps_per_mm_checksum		CHECKSUM("gamma_steps_per_mm_checksum");
+#endif
+
+const float pi     = 3.14159265358979323846;    // PI
```
wolfmanjm: these need to be const static otherwise they take up memory.

src/modules/robot/arm_solutions/RotatableDeltaSolution.cpp
```diff
@@ -0,0 +1,255 @@
...
+	float y0 = cartesian_mm[Y_AXIS];
+	float z_with_offset = cartesian_mm[Z_AXIS] + z_calc_offset; //The delta calculation below places zero at the top.  Subtract the Z offset to make zero at the bottom.
+
+	int status =              delta_calcAngleYZ(x0,                    y0,                  z_with_offset, alpha_theta);
+	if (status == 0) status = delta_calcAngleYZ(x0*cos120 + y0*sin120, y0*cos120-x0*sin120, z_with_offset, beta_theta);  // rotate co-ordinates to +120 deg
+	if (status == 0) status = delta_calcAngleYZ(x0*cos120 - y0*sin120, y0*cos120+x0*sin120, z_with_offset, gamma_theta);  // rotate co-ordinates to -120 deg
+
+	if (status == -1)  //something went wrong
+	{
+	    THEKERNEL->streams->printf("ERROR: Delta calculation fail!  Unable to move to:\n");
```
wolfmanjm: This is not allowed, an arm solution is not allowed to report errors like this, it will kill smoothie.

src/modules/robot/arm_solutions/RotatableDeltaSolution.h
```diff
@@ -0,0 +1,48 @@
...
+        //float tool_offset;	// No offset
+        //float tool_offset;	// Distance between end effector ball joint plane and tip of tool (Z probe)
+        float tool_offset;	// Distance between end effector ball joint plane and tip of tool (PnP)
+        float z_calc_offset;
+
+        float z_home_angle;	// This is the angle where the arms hit the endstop sensor
+        float z_home_offs;	// This is calculated from the above angle, after applying forward
+        						// kinematics, and adding the Z calc offset to it.
+
+        // Print surface diameter/2 minus unreachable space (avoid collisions with vertical towers).
```
wolfmanjm: more things that are not used.

wolfmanjm:
what tool chain are you using? those warnings do not appear with the tool chain we use.

src/modules/robot/Robot.cpp
wolfmanjm: this is actually a good catch and is a bug. Not sure why the tool chain does not return this as an error

wolfmanjm:
the thing is it is incorrect, as CHECKSUM is actually a uint16_t not an unsigned int, hence you should not make changes to core code unless you fully understand what you are doing and check with me first.
So if this required a cast (which is does not) the cast would be uint16_t.

DouglasPearless:
Was using the mac_install on my Mac, though I found there may have been some old toolchain stuff on that machine.

wolfmanjm:
did you do a ./BuildShell becuase if not it would use a different tool chain which would be very bad.

Make sure you build with out toolchain otherwise there maybe nasty surprises.

If you were getting warnings on those casts then you were not using our toolchain.

DouglasPearless:
Yep, I have moved my development onto a linux box so that change was not undone.

I have undone it and I am not getting the warning on linux :-)

DouglasPearless:
I have (hopefully) fixed it all, let me know if I have missed anything or you find anything else :-)

DouglasPearless:
If you are now happy with the request, then I intend to move onto the Z-Probe for calibration.

I have looked as the current routines and I am not sure any of them are suitable.

While the FPD rotational delta uses 3 arms like a linear delta, it has 4 towers (and while it does use a square bed, it arguably has a circular build volume like a delta).

I believe that a version of the grid Z-probe which works for a X=0, Y=0, Z=0 at the lower centre would be suitable?

What are your views on this given your understanding of how the existing Z-Probe code works?

Cheers
Douglas

wolfmanjm:
I don;t see any changes since my last comment. There were still casts to unsigned int which are incorrect, and unused variables defined in the header.

src/modules/robot/Robot.cpp
```diff
@@ -195,20 +201,20 @@ void Robot::on_config_reload(void *argument)
...
-    alpha_en_pin.from_string(   THEKERNEL->config->value(alpha_en_pin_checksum   )->by_default("0.4"  )->as_string())->as_output();
-    beta_step_pin.from_string(  THEKERNEL->config->value(beta_step_pin_checksum  )->by_default("2.1"  )->as_string())->as_output();
-    beta_dir_pin.from_string(   THEKERNEL->config->value(beta_dir_pin_checksum   )->by_default("0.11" )->as_string())->as_output();
-    beta_en_pin.from_string(    THEKERNEL->config->value(beta_en_pin_checksum    )->by_default("0.10" )->as_string())->as_output();
-    gamma_step_pin.from_string( THEKERNEL->config->value(gamma_step_pin_checksum )->by_default("2.2"  )->as_string())->as_output();
-    gamma_dir_pin.from_string(  THEKERNEL->config->value(gamma_dir_pin_checksum  )->by_default("0.20" )->as_string())->as_output();
-    gamma_en_pin.from_string(   THEKERNEL->config->value(gamma_en_pin_checksum   )->by_default("0.19" )->as_string())->as_output();
+    alpha_step_pin.from_string( THEKERNEL->config->value((unsigned int)alpha_step_pin_checksum )->by_default("2.0"  )->as_string())->as_output();
```
wolfmanjm: remove casts...

src/modules/robot/arm_solutions/RotatableDeltaSolution.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+    for(auto &i : options) {
+        switch(i.first) {
+            case 'A': delta_e			= i.second; break;
+            case 'B': delta_f			= i.second; break;
+            case 'C': delta_re			= i.second; break;
+            case 'D': delta_rf			= i.second; break;
+            case 'E': delta_z_offset		= i.second; break;
+            case 'F': delta_ee_offs		= i.second; break;
+            case 'G': tool_offset		= i.second; break;
```
wolfmanjm: G is not a valid parameter (neither is T or M or N)

src/modules/robot/arm_solutions/RotatableDeltaSolution.cpp
```diff
@@ -0,0 +1,251 @@
...
+	float y0 = cartesian_mm[Y_AXIS];
+	float z_with_offset = cartesian_mm[Z_AXIS] + z_calc_offset; //The delta calculation below places zero at the top.  Subtract the Z offset to make zero at the bottom.
+
+	int status =              delta_calcAngleYZ(x0,                    y0,                  z_with_offset, alpha_theta);
+	if (status == 0) status = delta_calcAngleYZ(x0*cos120 + y0*sin120, y0*cos120-x0*sin120, z_with_offset, beta_theta);  // rotate co-ordinates to +120 deg
+	if (status == 0) status = delta_calcAngleYZ(x0*cos120 - y0*sin120, y0*cos120+x0*sin120, z_with_offset, gamma_theta);  // rotate co-ordinates to -120 deg
+
+	if (status == -1)  //something went wrong
+	{
+//DEBUG CODE, uncomment the following to help determine what may be happening if you are trying to adapt this to your own different roational delta.
```
wolfmanjm: you can't do nothing here you at least need to set actuator_mm to a reasonable value.

src/modules/robot/Robot.cpp
DouglasPearless: I was under the assumption that not changing the actuator mm in this routine meant that the calling routine will receive back whatever the actuator mm was prior to calling, i.e. no change if the code cannot return a valid value?

src/modules/robot/arm_solutions/RotatableDeltaSolution.cpp
DouglasPearless: Ah, I see why, as these are reserved for the G T M & N codes :-)

I have fixed that.

src/modules/robot/arm_solutions/RotatableDeltaSolution.cpp
```diff
@@ -0,0 +1,253 @@
...
+#define delta_z_offset_checksum 		CHECKSUM("delta_z_offset_checksum")
+
+#define delta_ee_offs_checksum 		  	CHECKSUM("delta_ee_offs_checksum")
+#define tool_offset_checksum			CHECKSUM("tool_offset_checksum")
+
+#define z_home_angle_checksum        		CHECKSUM("z_home_angle_checksum")
+
+#define delta_printable_radius_checksum    	CHECKSUM("delta_printable_radius_checksum")
+
+#ifndef alpha_steps_per_mm_checksum
```
wolfmanjm: Are these needed? I don;t see them used so should be removed.

src/modules/robot/arm_solutions/RotatableDeltaSolution.cpp
```diff
@@ -0,0 +1,253 @@
...
+const static float sin120 = 0.86602540378443864676372317075294; //sqrt3/2.0
+const static float cos120 = -0.5;
+const static float tan60  = 1.7320508075688772935274463415059; //sqrt3;
+const static float sin30  = 0.5;
+const static float tan30  = 0.57735026918962576450914878050196; //1/sqrt3
+
+RotatableDeltaSolution::RotatableDeltaSolution(Config* config)
+{
+	// End effector length
+	delta_e = config->value((unsigned int)delta_e_checksum)->by_default(131.636F)->as_number();
```
wolfmanjm: cast to unisigned int is incorrect.

src/modules/robot/arm_solutions/RotatableDeltaSolution.h
```diff
@@ -0,0 +1,44 @@
...
+        float delta_ee_offs;		// Ball joint plane to bottom of end effector surface
+        float tool_offset;		// Distance between end effector ball joint plane and tip of tool
+        float z_calc_offset;
+
+        float z_home_angle;		// This is the angle where the arms hit the endstop sensor
+        float z_home_offs;		// This is calculated from the above angle, after applying forward
+        				// kinematics, and adding the Z calc offset to it.
+
+        
+        float delta_printable_radius;	// Print surface diameter/2 minus unreachable space (avoid collisions with vertical towers).
```
wolfmanjm: unused

src/modules/robot/arm_solutions/RotatableDeltaSolution.h
```diff
@@ -0,0 +1,44 @@
...
+
+        float xyz_full_steps_per_rotation;	// stepper motor steps per 360 full rotation
```
wolfmanjm: unused

src/modules/robot/arm_solutions/RotatableDeltaSolution.h
```diff
@@ -0,0 +1,44 @@
...
+
+        float delta[3] = {0.0, 0.0, 0.0};
```
wolfmanjm: unused

ConfigSamples/FirePick.delta/config.FirePickSmoothieOnABreadboard
```diff
@@ -0,0 +1,331 @@
+# NOTE Lines must not exceed 132 characters
```
wolfmanjm: I feel this file should not be included.

src/modules/robot/arm_solutions/RotatableDeltaSolution.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+	int status =              delta_calcAngleYZ(x0,                    y0,                  z_with_offset, alpha_theta);
+	if (status == 0) status = delta_calcAngleYZ(x0*cos120 + y0*sin120, y0*cos120-x0*sin120, z_with_offset, beta_theta);  // rotate co-ordinates to +120 deg
+	if (status == 0) status = delta_calcAngleYZ(x0*cos120 - y0*sin120, y0*cos120+x0*sin120, z_with_offset, gamma_theta);  // rotate co-ordinates to -120 deg
+
+	if (status == -1)  //something went wrong
+	{
+	    //Do nothing so the actuator_mm is unchanged by this routine.
+            // as we have no idea what it should be
+//DEBUG CODE, uncomment the following to help determine what may be happening if you are trying to adapt this to your own different roational delta.
```
wolfmanjm: actuator_mm is not set to anything on call so needs to be written to otherwise it will have random data in it.

src/modules/robot/Robot.cpp
DouglasPearless: Hmm,

On re-reading the Robot.cpp I realise that all methods except "void Robot::reset_position_from_current_actuator_position()" pass actuator_pos as an unitialised array (which depending on the compiler may initialise as [0,0,0] ).

There does not appear to be a way that the 

        void cartesian_to_actuator( float[], float[] );
        void actuator_to_cartesian( float[], float[] );

methods can inform the calling routine that it was unable to calculate a value, what are your thoughts on this use-case?

Cheers
Douglas

ConfigSamples/FirePick.delta/config.FirePickSmoothieOnABreadboard
DouglasPearless: OK, I will remove that, I am not fussed about it :-)

src/modules/robot/Robot.cpp
wolfmanjm: there should always be a valid answer, there is no case where it is ok to not return something.

you would need to move further up the chain to stop a print if an invalid gcode or position were 
given and there currently is no way to detect invalid gcode.

src/modules/robot/arm_solutions/RotatableDeltaSolution.cpp
DouglasPearless: The following are used in method delta_calcAngleYZ:
delta_e
delta_f
delta_re
delta_rf

and in the init():
z_calc_offset is calculate from
delta_z_offset
tool_offset
delta_ee_offs

however, the z_home_angle is not currently used, will be used in the following formula:

	z_home_offs    = (((delta_z_offset - tool_offset - delta_ee_offs) - 182.002) - 0.5);

as I have hardwired the 182.002 in the mean time to get it working, so while z_home_angle is not currently used, I’d like to keep in in there until we sort that one piece out (please!); if we don’t actually need it and that value is true for all builds, then I will remove it.  I am waiting on a reply from Neil Jensen who is rather busy trying to get the beta programme sorted.

The:
> delta_printable_radius
is not currently used but is planned for a release to enable more sophisticated checking that hot-end tool head stays within the printable area, as the PnP toolhead can move outside of it.

My intention with the ToolManager rewrite is to parameterise the allowable travel limits for tool heads.

If you want I can remove it for now and put it back later.

Cheers
Douglas

wolfmanjm: well if this is not the final version then please hold off pushing it until it is final.

Just let me know when you are done.

I'm sorry to be pushy but we have a very high standard of code review, and it is very time consuming 
to review every pull request, hence I really do not want to do this piece meal.

Thank you

On 06/10/2015 05:06 PM, Douglas Pearless wrote

DouglasPearless: Hi Jim,

Good point.

I will remove all the code intended for future feature sets and make it a clean implementation and later create a new branch for the next feature set.

It will be the final code for a Rotational Delta arm solution, munis future to-be-implemented features.

Cheers
Douglas

DouglasPearless: Hi Jim,

Thanks for your patience.

I have removed all the extraneous code and variables that are not required to implement a generic Rotatable Delta mechanism, including the Smoothie-on-a-breadboard config file

As this is now a generic solution (with an example config for the FPD), hopefully this now meets your requirements and can be accepted.

If I have missed something, please let me know as I am rather keen to get this into Edge for everyone to use.

Cheers
Douglas

src/modules/robot/arm_solutions/RotatableDeltaSolution.cpp
```diff
@@ -0,0 +1,242 @@
...
+#define delta_e_checksum       			CHECKSUM("delta_e_checksum")
+#define delta_f_checksum   	    		CHECKSUM("delta_f_checksum")
+#define delta_re_checksum       		CHECKSUM("delta_re_checksum")
+#define delta_rf_checksum       		CHECKSUM("delta_rf_checksum")
+#define delta_z_offset_checksum 		CHECKSUM("delta_z_offset_checksum")
+
+#define delta_ee_offs_checksum 		  	CHECKSUM("delta_ee_offs_checksum")
+#define tool_offset_checksum			CHECKSUM("tool_offset_checksum")
+
+#ifndef alpha_steps_per_mm_checksum
```
wolfmanjm: please remove these three as we do not like #ifdefs in the code is makes it harder to read and these appear to not be needed anywhere

src/modules/robot/arm_solutions/RotatableDeltaSolution.h
```diff
@@ -0,0 +1,40 @@
...
+			       		// NOTE: For OpenPnP, set the zero to be about 25mm above the bed
+
+        float delta_ee_offs;		// Ball joint plane to bottom of end effector surface
+        float tool_offset;		// Distance between end effector ball joint plane and tip of tool
+        float z_calc_offset;
+
+        float z_home_offs;		// This is calculated from the above angle, after applying forward
+        				// kinematics, and adding the Z calc offset to it.
+
+        float xyz_full_steps_per_rotation;	// stepper motor steps per 360 full rotation
```
wolfmanjm: still unused

src/modules/robot/arm_solutions/RotatableDeltaSolution.h
```diff
@@ -0,0 +1,40 @@
...
+
+        float delta[3] = {0.0, 0.0, 0.0};
```
wolfmanjm: I cannot see where this is used at all. Please note we are very short on memory so using any memory when it is not required is not good.

src/modules/robot/arm_solutions/RotatableDeltaSolution.h
DouglasPearless: I have high lighted in green the method that uses the variable, and if red if it should have been removed (and now has been).

wolfmanjm:
float xyz_full_steps_per_rotation;
  float delta[3] = {0.0, 0.0, 0.0};

are not referenced anywhere in the code

DouglasPearless:
Fixed

DouglasPearless:
Hi Jim,

Thank you, my first contribution!

Cheers
Douglas

triffid:
@DouglasPearless congrats, and thanks for sticking with us through this process!

Some find it too discouraging and quit, however it's the only reasonable way we can maintain code quality.

DouglasPearless:
Hi Michael,

No problems, I have been a coder for many a year and like nice clean readable code.

I intend to re-write the ToolManager to make it behave more like the G and M code routines as a foundation to allow auto tool head and tool changing for the FirePick delta bot we are working on.

Cheers
Douglas

logxen:
Umm ... this actually just needs a small change still. "RotatableDelta" is not the correct name for this. The module and its config options should be called "RotaryDelta"

wolfmanjm:
i can fix that.

DouglasPearless:
I also am happy to do that so I can clean compile and test the changes on real hardware.

DouglasPearless:
I have made all the changes and compiled and check it, and pushed them to you.
NOTE: I have not renamed the branch, do you want that too?

DouglasPearless:
I have made all the changes and compiled and check it, and pushed them to you.
NOTE: I have not renamed the branch, do you want that too?

wolfmanjm:
Yea it needs to be a different pull request I cannot merge this as it is already merged.

DouglasPearless:
oK, is it sufficient for me to rename the branch from feature/rotatable-delta to feature/rotary-delta and will this achieve the same thing, then we can delete the feature/rotatable-delta branch ?

wolfmanjm:
no you need to create a new branch from current edge, make the changes, then push the branch to your 
github then do a pull request

DouglasPearless:
OK, I will do that :-)

DouglasPearless:
I think I have managed to create a new branch feature/rotary-delta with all the changes and send a pull request to your repo.

Cheers
Douglas

wolfmanjm:
I never saw the new PR? Are yo still working on this? as I am building a delta which is similar and would use the arm solution. (I already fixed some typos )

DouglasPearless:
Hi,

Everything should work just nicely so I have not needed to send any more pull requests; 

I have pressed pause on this while I complete another project; in the meantime have a look at the FPD website http://tinwhiskers.io <http://tinwhiskers.io/> and  http://firepickdelta.dozuki.com <http://firepickdelta.dozuki.com/>.

Be aware we are having issues with what is called the ‘bowl’ effect where movement across the Z-Plane is not flat and we are having to write some crazy maths to fix it; I would not recommend this for a 3D printer for anything except small print jobs (50x50x50mm) as the error gets up to 3-4mm out towards the edge of the build plate. I was getting some very fast print speeds with this arm solution, but again, only small items.

I note that  https://github.com/626Pilot/Smoothieware.git <https://github.com/626Pilot/Smoothieware.git> has some interesting stuff that I will try to adapt for the FPD.

Cheers
Douglas

wolfmanjm:
ok, so I looked at your branch and notice you did rename to rotarydelta, but did not submit that. So I Have a PR that renames accordingly. I didn't immediately see if you fixed any other things except two small changes to the FK.

I was considering a zgrid like compensation like Johann did for the kossel. I am not a fan of the techniques that 626pilot and others have done. There are at least three different attempts to do the same thing just using different least error math. I am rather taken with the offline calculator that dc did for his firmware...  http://escher3d.com/pages/wizards/wizarddelta.php

wolfmanjm:
seems you made some critical sign changes in your branch which were not in the smoothie code. I just updated based on the diffs I found between your version and the version you did a PR for, and now FK works properly, which probably means the new trim will work. Your branch is terribly out of date though wrt to smoothie edge.

DouglasPearless:
Hi

Yes I do need to revise the code and put my rotary delta back together to test it!

wolfmanjm:
I wanted to give you a heads up, I have overhauled the rotary delta code in smoothieware edge, it 
still seems to work ok ;)

The biggest changes are introducing a is_rdelta to the endstop.cpp as it appeared that normal 
endstop code won't work for rotary delta homing.

Basically the changes are that trims are disabled as trim does not appear to work with rotary 
deltas. (at least not the new way they are done).

home_offsets (set with M206) are the theta offsets for the homing angle.

The alpha_max, beta_max, gamma_max are now the angle that the actuators are at when homed.

When a rotary delta homes it uses the angle given in the alpha_max+home_offset[0] and uses the FK 
(actuator_to_cartesian) to input the angles and get the cartesian XYZ for those angles, I added a 
new call Robot::reset_actuator_position(a, b, c) where a b c are the theta angles for each actuator.

This will set the actuators to the exact angle (with theta offsets if needed) and set the cartesian 
positions accordingly.

M306 and trim (M666) is disabled for rotary delta. (unless someone can explain to me how trim would 
actually work in this case).

I have a different design to firepick and it is easy to make a calibration tool to determine the 
exact angle each actuator homes to and adjust the theta offset accordingly, that way every actuator 
homes to the correct angle, and they do not all need to home to the exact same angle.

My plans are to implement the delta grid levelling compensation that Johann Rocholl used for the 
Kossel, I think this will compensate for the bowl shapes and the bed being not tram with the head 
movements (thus making what we use trim for in a linear delta obsolete).

Comments and feedback are welcome.

Here is a video of my rotary deltas first moves (yes it is an upside down one and the bed moves and 
the extruder is stationary).

https://youtu.be/IL7GPGmQZSA

DouglasPearless:
Hi Jim,

I like your “upside-down” rotary delta!

My RotaryDelta is currently in pieces as I experiment with new arms (etc).

As soon as I put it back together I will try your updated code.

BTW, we have great difficultly with keeping movement on the X&Y level on the Z Plane (it tends to have a bowl effect toward the outer limits ox X and Y where the Z can be out as much as 3mm!), are you also experiencing that?

Cheers
Douglas

DouglasPearless:
Hi,

I am looking at restarting work on the Smoothie firmware and was wondering why you were not a fan on 626Pilot’s Annealing method, it is working rather well for me on a linear delta and I am about to port it to the Rotary Delta ARM solution.  What in your opinion is a better method?

Cheers
Douglas

wolfmanjm:
I do not believe it will work on a rotary delta due to the non linear nature of everything. Also there are no parameters to tweak, it needs the current parameters to be exact, and just measuring the Z deltas will not allow you to compensate adequately, what you will find is your XY dimensions will be way out even if your heights are ok.
Anyway the best method is the delta grid I just implemented. This was Johanns original method to compensate for bed issues, and IMO still the best, it also seems to work pretty well for rotary deltas, although not as well as it works for linear delta.
Feel free to port 626pilots stuff, I will look at your attempts with interest. 
However do not expect it to be accepted into the smoothie mainline code.

DouglasPearless:
Thanks for the information, I will try your updated Edge branch and see how the updated delta grid goes with my Rotary Delta (need to put it back together again!)

I will probably also play with 626Pilot’s code to see how it works vs the other (accuracy).

I need to sit down and perform some empirical testing so I can confirm what any changes to the code (e.g. 686Pilots’s code) does to affect the accuracy.

Cheers
Douglas

wolfmanjm:
I understand his code, it basically randomly adjusts the tower offsets and angles and trim and rod length until the z height errors are minimized. There are at least two other things that do basically the same thing, just using a different least errors method. The trouble with rotary deltas is that there are no tower angles and offsets to adjust, and trim does not work for rotary delta either so is disabled for them. So there are no parameters to tweak on the rotary delta. If you want to use a least error technique and can add the parameters it needs to tweak, I prefer the one here...   http://escher3d.com/pages/wizards/wizarddelta.php

It is offline so uses no smoothie resources.

Issue #649, opened 2015-03-31
[ZGrid] Several bugs/Todos


wolfmanjm:
1. Support circular beds
2. Initial move after homing needs to be fast rate not slow rate (deltas have a long way to move).
https://github.com/Smoothieware/Smoothieware/blob/edge/src/modules/tools/zprobe/ZGridStrategy.cpp#L502
3. There needs to be an option to tell M500 to load the grid data or not.. Currently it always inserts a M375 which is particularly a problem if it has not been saved with M374. Default can be to restore default grid.
https://github.com/Smoothieware/Smoothieware/blob/edge/src/modules/tools/zprobe/ZGridStrategy.cpp#L366

sorki:
> while(!zprobe->getProbeStatus()); <---- This is a big NONO, must call ON_IDLE this will kill USB.

This doesn't even work for me for some reason and the probe attach/detach routine should be implemented elsewhere - preferably configurable like  "Pause when out of filament" is implemented.

Is there a G-code that would wait for user pressing a specified button?

> Support circular beds

This should be relatively easy to implement with masking/skipping parts of the grid that are outside of the delta radius.

I'll add one more issue - this feature would be extremely useful for PCB milling but it has to support small dimensions (e.g. 3x3cm). According to testing of one user on the IRC this doesn't work as expected and results in weird behaviour. For this to be convenient for users it should be possible to specify dimensions to be probed with G-code params.

Leveling in this case also seems to depend on line segments per second which is not enabled on cartesians by default.

qharley:
The list is growing fast! ;-)

qharley:
I am working on some of these issues, but I do not have a Delta.  Round bed and middle zero will be a best guess fix if I attempt it.

wolfmanjm:
I think there are several people who can test on a delta. I certainly can.

qharley:
I will take you up on that offer.

barsamgr:
Hello
I noticed to bugs in the Zgrid code. I have corrected them in my own fork, but the correction is not in a separate brach.
1. (Not definitely a bug, but my opinion) Z should move to zprobe->getProbeHeight() and not to std::get<Z_AXIS>(this->probe_offsets) + zprobe->getProbeHeight(). It is supposed that Probe Height is defined as a safe distance to clear the probe movement.
[...]

PR #679, opened 2015-07-01
zgrid: no homing prior probing

sorki:
Redundant and dangerously unexpected
- user can issue G28 manually if required

Related to #649.

### DIFF
--- a/src/modules/tools/zprobe/ZGridStrategy.cpp
+++ b/src/modules/tools/zprobe/ZGridStrategy.cpp
@@ -447,9 +447,6 @@ void ZGridStrategy::setZoffset(float zval)
 
 bool ZGridStrategy::doProbing(StreamOutput *stream)  // probed calibration
 {
-    // home first
-    this->homexyz();
-
     // deactivate correction during moves
     this->setAdjustFunction(false);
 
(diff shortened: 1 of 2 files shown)


qharley:
Zgrid is an automated grid system, and if not homed will wreck havoc on
most systems (not excluding my own).
If you like I can add a config option for "No Home" for users that want to
home the machine themselves.


  [image: photo]
*Quentin Harley*
Founder, Reprap Morgan 3D

wolfmanjm:
Yes it should be an option default set to home

qharley:
Will do so tomorrow. I am working on the firmware at the moment, so timing
is good...

wolfmanjm:
actually it should be configurable to home in XY only or XYZ. Some systems may not want to home in Z until the probe is done, but there is no system I can think of that could do a probe without at least homing in XY first.

qharley:
Ok. Will send a pull request when done

sorki:
> but there is no system I can think of that could do a probe without at least homing in XY first.

Any CNC without endstops. I wanted to use Z-Grid for PCB probing but it seems easier to implement it in host rather than in firmware

qharley:
I did a pull request for this: https://github.com/Smoothieware/Smoothieware/pull/681

wolfmanjm:
@sorki if there are no endstops then the endstop module would be disabled and G28 would be ignored, so there is still no case where homing in at least XY would not be required before a probe that requires to know where XY are.

Issue #500, opened 2014-08-25
G3 Counter Clockwise Arcs


Veloce7:
The G2 command works, but G3 does nothing.

sorki:
G0 X0 Y0
G2 X10 Y10 I5 J5
G2 X0 Y0 I5 J5

works properly. Same g-code with G3 ends on X10 Y10 (last command is ignored).

Not sure if this makes sense but it seems there's some inconsistency between G2/G3.

arthurwolf:
Yeah this has been reported several times now, but I looked at the code and can't find anything that could cause this.
I tried debugging this but ended up burning my usb port and had to move to doing other stuff unfortunately ...

GeorgeIoak:
Been doing some testing myself and I've found that some G3 commands seem to work. I'm testing with [this file](http://pastebin.com/7xRef0B7) and it stopped at line 1165 but if you search the file you'll see that there are other G3 command sent before this line. Maybe this help debug the code.

sorki:
@GeorgeIoak there are modal G3 arcs you have in your g-code

```
G3 X41.089 Y-28.164 I0.8 J0.015 F1256
Y-28.151 I-0.8 J-0.002
```

and these are not supported by smoothieware - you need either post process such lines so each of them is prefixed with G3 like this:

```
G3 X41.089 Y-28.164 I0.8 J0.015 F1256
G3 Y-28.151 I-0.8 J-0.002
```

or fix the output of your CAM processor.

It might be a good idea to enable modal G3 codes in smoothie if it doesn't break command parsing.

GeorgeIoak:
Thanks @sorki for looking at that file. I used Autodesk Fusion 360 to generate that file. I think it has an option for post processing but it seems since Smoothie allows other g-codes to be modal that G3 could be "upgraded" to be modal as well?

wolfmanjm:
G0 and G1 are currently semi-modal, the caveat is there needs to be a preceding space on the X or Y.
G2 or G3 can also be easily made modal with the same caveat.
eg...
```
G3 X41.089 Y-28.164 I0.8 J0.015 F1256
   Y-28.151 I-0.8 J-0.002
```

GeorgeIoak:
Is that something that is "supposed" to be done or just a requirement for Smoothie? I'm new to g-code and post processing but it seems odd that the file needs to have a certain format. Coming from PCB design, when you generate gerbers you never have to worry if the file is properly formatted.

[...]

PR #682, opened 2015-07-03
fixed arc problem with G03 - patched from grbl project

st0ne:
No movement is performed when G03 is used and a full circle (end point is coincident with start point) is programmed. Problem found by Christoph Baumann and me.


wolfmanjm:
fixes #500

arthurwolf:
Thank you so much !!!

PR #684, opened 2015-07-08
Probe improvements

sorki:
- G30 Fxyz support (custom probing feedrate)
- M670 support for live probe configuration and config-override saving support
- zprobe.return_feedrate config option
- zprobe.max_z is not divided by two
- safe default feedrates for cartesians

```
    ;Probe feedrates slow/fast/return (mm/sec):
    M670 S0.50 F2.00 R2.50
    ;Probe max_z (mm):
    M670 Z200.00
    ;Probe height (mm):
    M670 H2.00
```

New options/commands are not documented nor included in config samples yet - I'll add them to this pull request & wiki when this is reviewed.

### DIFF
--- a/src/modules/tools/zprobe/ZProbe.cpp
+++ b/src/modules/tools/zprobe/ZProbe.cpp
@@ -123,8 +123,8 @@ void ZProbe::on_config_reload(void *argument)
     }
 
     this->probe_height  = THEKERNEL->config->value(zprobe_checksum, probe_height_checksum)->by_default(5.0F)->as_number();
-    this->slow_feedrate = THEKERNEL->config->value(zprobe_checksum, slow_feedrate_checksum)->by_default(5)->as_number(); // feedrate in mm/sec
-    this->fast_feedrate = THEKERNEL->config->value(zprobe_checksum, fast_feedrate_checksum)->by_default(100)->as_number(); // feedrate in mm/sec
+    this->slow_feedrate = THEKERNEL->config->value(zprobe_checksum, slow_feedrate_checksum)->by_default(1)->as_number(); // feedrate in mm/sec
+    this->fast_feedrate = THEKERNEL->config->value(zprobe_checksum, fast_feedrate_checksum)->by_default(2)->as_number(); // feedrate in mm/sec
     this->max_z         = THEKERNEL->config->value(gamma_max_checksum)->by_default(500)->as_number(); // maximum zprobe distance
 }
 
--- a/src/modules/tools/zprobe/ZProbe.cpp
+++ b/src/modules/tools/zprobe/ZProbe.cpp
@@ -37,6 +37,7 @@
 #define debounce_count_checksum  CHECKSUM("debounce_count")
 #define slow_feedrate_checksum   CHECKSUM("slow_feedrate")
 #define fast_feedrate_checksum   CHECKSUM("fast_feedrate")
+#define return_feedrate_checksum CHECKSUM("return_feedrate")
 #define probe_height_checksum    CHECKSUM("probe_height")
 #define gamma_max_checksum       CHECKSUM("gamma_max")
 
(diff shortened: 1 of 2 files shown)


arthurwolf:
Hey :) Thanks for the code.

Here : https://github.com/Smoothieware/Smoothieware/pull/684/files#diff-17c0c771d21d822323322904d7a5e6f1L126 you are changing a default value, that's going to distrub people, could you fix that ?

Thanks.

src/modules/tools/zprobe/ZProbe.cpp
```diff
@@ -176,15 +178,14 @@ bool ZProbe::run_probe(int& steps, bool fast)
 
     // Enable the motors
     THEKERNEL->stepper->turn_enable_pins_on();
-    this->current_feedrate = (fast ? this->fast_feedrate : this->slow_feedrate) * Z_STEPS_PER_MM; // steps/sec
-    float maxz= this->max_z*2;
```
wolfmanjm: This was done for a reason, so should not be changed. an alternative is to set a maxz independent config for zprobe and not use the endstop maxz

src/modules/tools/zprobe/ZProbe.cpp
```diff
@@ -123,9 +124,10 @@ void ZProbe::on_config_reload(void *argument)
     }
 
     this->probe_height  = THEKERNEL->config->value(zprobe_checksum, probe_height_checksum)->by_default(5.0F)->as_number();
-    this->slow_feedrate = THEKERNEL->config->value(zprobe_checksum, slow_feedrate_checksum)->by_default(5)->as_number(); // feedrate in mm/sec
-    this->fast_feedrate = THEKERNEL->config->value(zprobe_checksum, fast_feedrate_checksum)->by_default(100)->as_number(); // feedrate in mm/sec
+    this->slow_feedrate = THEKERNEL->config->value(zprobe_checksum, slow_feedrate_checksum)->by_default(1)->as_number(); // feedrate in mm/sec
```
wolfmanjm: cannot change the default values here, they must be left as they were for backward compatibility. Breaking existing users use case is not nice.

sorki:
I've did that on purpose - see "zprobe: lower default feedrates" commit. For cartesian machines with leadscrew on Z 100 mm/sec is way too much. I can leave out this commit if you think it's going to cause trouble - I think it shouldn't as config samples include both values so people probably already have them in their config files

wolfmanjm:
Nope, old configs did not have these defaults, so you cannot change them.

src/modules/tools/zprobe/ZProbe.cpp
```diff
@@ -289,19 +288,40 @@ void ZProbe::on_gcode_received(void *argument)
...
+            case 670:
+                if (gcode->has_letter('S')) this->slow_feedrate = gcode->get_value('S');
+                if (gcode->has_letter('F')) this->fast_feedrate = gcode->get_value('F');
+                if (gcode->has_letter('R')) this->return_feedrate = gcode->get_value('R');
+                if (gcode->has_letter('Z')) this->max_z = gcode->get_value('Z');
+                if (gcode->has_letter('H')) this->probe_height = gcode->get_value('H');
+                gcode->mark_as_taken();
+                break;
+
+            default:
```
wolfmanjm: This breaks all the strategies gcode handling.

src/modules/tools/zprobe/ZProbe.cpp
sorki: So this needs probe.max_z ? or max_travel?

src/modules/tools/zprobe/ZProbe.cpp
sorki: ah, I see. they need 500/503

wolfmanjm:
I am sorry but you changed too much and most of it breaks existing functionality.
I am rejecting this pull request.
Please resubmit each one separately.

So the main one that seems to not be broken is setting specific feedrates for the return probe, and a m code to set these and save them.
However the case statement does not allow M500/M503 to be passed onto the strategies which is breaking their ability to save stuff. All M codes need to be sent on to the levelstrategy as the original code allowed.

Changing the defaults is not a good idea as that will upset people who currently have things working well.

I have commented on the individual lines that I feel should not have been changed.

maxz could probably be set with a specific maxzprobe config, but the original legacy code of default maxz*2 needs to be done if it is not done, as it was not in existing configs. Backward compatibility here is key. maxz*2 was done on purpose to avoid issue with deltas. if specifically set then the *2 can be left out.

There maybe other problems I have missed. But resubmitting one feature change at a time will make it easier to review the problems and fix them.

Small changes are better than doing it all at once!
Oh and all leveling strategies will then need to be tested to make sure nothing else broke.

Thanks and sorry for the critical assessment.

src/modules/tools/zprobe/ZProbe.cpp
wolfmanjm: yea i would suggest a probe.max_travel (or max_z) but then the code needs to allow for legacy configs that do not have that specified and default to exactly what it was before. the rule of least surprises to legacy users is needed here.
and as I said each of these changes needs its own PR, it makes it easier to review and see what else may break. Pay particular attention to the levelling strategies that rely on zprobe functionality to not change.

src/modules/tools/zprobe/ZProbe.cpp
wolfmanjm: Yes and also other mcodes, do not presume if you process an mcode that the strategies do not also.

Issue #685, opened 2015-07-18
Laser module Minimum and Maximum power settings


thinkl33t:
Currently, the `laser_module_max_power` sets the default power that is used when a `G1`/`G2`/`G3` command is issued with no `S` modifier.  This value is then auto-adjusted in the laser module to make it proportional to the head speed, before it is then output to the laser PWM pin.

Reading the documentation, I got the mistaken impression this config variable would set the limit value and scale the `S` values using it. 

I plan on replacing the config variables:
* `laser_module_max_power`
* `laser_module_tickle_power`

with:
* `laser_module_max_power` - New config variable, used to set the maximum PWM frequency that will be sent to the laser.
* `laser_module_min_power` - Replace old `laser_module_tickle_power`, this would be the minimum power sent to the laser, and would normally be set to a value *just under* the value at which your laser fires.  On `G0` moves, the laser would be set to this power.
* `laser_module_default_power` - The power level to use if a value hasn't been set by an `S` modifier.

Any S values sent, along with the default power, would be scaled automatically to be on a scale between min and max power.  for example, if the config is set to:
 
```
laser_module_max_power = 0.8 
laser_module_min_power = 0.2
laser_module_default_power = 0.6
```

* If the user sends a `G0` command, the laser PWM output would be set to `0.2`.
* If the user sends a `G1` command, the laser PWM would be set to `((0.8-0.2)*0.6)+0.2 = 0.56`
* If the user sends a `G1 S0.1` command, the laser PWM would be `((0.8-0.2)*0.1)+0.2 = 0.26`
* If the user sends a `G1 S1.0` command, the laser PWM would be `((0.8-0.2)*1.0)+0.2 = 0.8`

This preserves existing functionality, and provides a consistent range for scaling the laser power, and allows laser users who may want to preserve tube life to set a maximum limit to laser power.  

Also, as a laser tube ages, the power needed to make it turn on gets higher, so the minimum limit allows the user to adjust up, and continue to use the same gcode files with no changes.

wolfmanjm:
you need to maintain backward compatibility with old configs. So rather than replace you would add new settings, and the old settings would remain the same.

arthurwolf:
+wolfmanjm in this specific case, users actually expect it to behave the way he is describing, and I've gotten several reports of people thinking this is a bug, and having to do dirty host-side hacks.
For the few users for which this will break backwards compatibility, it will do so in a way that's not dangerous ( ie it'll be lower power than they expect ).
[...]

PR #686, opened 2015-07-19
Changes to laser module PWM config.

thinkl33t:
This changes the config to work as described in issue #685.  It makes
the laser_module_max_power work as expected / documented, introduces
a minimum power setting, and scales the output between the two.

It should be backwards-compatible with existing configs, but I've not been
able to test it on a live machine yet.  Hopefully I'll be able to get it on a scope 
over the next couple of days to confirm operation.

### DIFF
--- a/ConfigSamples/AzteegX5Mini.delta/config
+++ b/ConfigSamples/AzteegX5Mini.delta/config
@@ -66,10 +66,16 @@ extruder_en_pin                              0.4              # Pin for extruder
 delta_current                                0.7              # Extruder stepper motor current
 
 # Laser module configuration
-laser_module_enable                          false            # Whether to activate the laser module at all. All configuration is ignored if false.
-#laser_module_pin                             2.7              # this pin will be PWMed to control the laser
-#laser_module_max_power                       0.8              # this is the maximum duty cycle that will be applied to the laser
-#laser_module_tickle_power                    0.0              # this duty cycle will be used for travel moves to keep the laser active without actually burning
+laser_module_enable                          false            # Whether to activate the laser module at all. All configuration is
+                                                              # ignored if false.
+#laser_module_pin                             2.5             # this pin will be PWMed to control the laser. Only P2.0 - P2.5, P1.18, P1.20, P1.21, P1.23, P1.24, P1.26, P3.25, P3.26
+                                                              # can be used since laser requires hardware PWM
+#laser_module_maximum_power                   1.0             # this is the maximum duty cycle that will be applied to the laser
+#laser_module_minimum_power                   0.0             # This is a value just below the minimum duty cycle that keeps the laser
+                                                              # active without actually burning.
+#laser_module_default_power                   0.8             # This is the default laser power that will be used for cuts if a power has not been specified.  The value is a scale between
+                                                              # the maximum and minimum power levels specified above
+#laser_module_pwm_period                      20              # this sets the pwm frequency as the period in microseconds
 
 # Hotend temperature control configuration
 temperature_control.hotend.enable            true             # Whether to activate this ( "hotend" ) module at all. All configuration is ignored if false.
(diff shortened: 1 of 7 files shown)


src/modules/tools/laser/Laser.cpp
```diff
@@ -59,8 +61,14 @@ void Laser::on_module_loaded() {
     this->laser_pin->period_us(THEKERNEL->config->value(laser_module_pwm_period_checksum)->by_default(20)->as_number());
     this->laser_pin->write(this->laser_inverting ? 1 : 0);
 
-    this->laser_max_power =    THEKERNEL->config->value(laser_module_max_power_checksum   )->by_default(0.8f)->as_number() ;
-    this->laser_tickle_power = THEKERNEL->config->value(laser_module_tickle_power_checksum)->by_default(0   )->as_number() ;
+    this->laser_max_power = THEKERNEL->config->value(laser_module_max_power_checksum   )->by_default(1.0f)->as_number() ;
```
wolfmanjm: This changes the defualt behavior and therefore breaks backwards compatibility with legacy config files.

logxen:
The change of default max_power from 0.8 to 1.0 is not ok ... full power will break a generic 40W blue box laser because they only come with 40W psu ... the laser tubes are really only 30W. So the default is 0.8. It's easy to turn it all the way up if you know it won't hurt your machine.

This also changes the name of the existing tickle_power config setting which would obviously break existing configs. You could add min_power if you really want but you can't remove tickle_power. (Read them both and use the one as the default value when you read the second one)

This also changes the meaning of max_power ... but in this case it is a good change I expect most people are confused about what this does. In removing this setting you should also add a default_power that does what max_power used to do.

As to the actual functional code change to ::set_proportional_power ... nice catch. That's a good improvement. :)

logxen:
O one more thing ... as you are changing and adding config settings you also need to update the ConfigSamples

thinkl33t:
I have changed the config variables used as suggested by arthur in #685

ConfigSamples/AzteegX5Mini.delta/config
```diff
@@ -66,10 +66,16 @@ extruder_en_pin                              0.4              # Pin for extruder
 delta_current                                0.7              # Extruder stepper motor current
 
 # Laser module configuration
-laser_module_enable                          false            # Whether to activate the laser module at all. All configuration is ignored if false.
-#laser_module_pin                             2.7              # this pin will be PWMed to control the laser
-#laser_module_max_power                       0.8              # this is the maximum duty cycle that will be applied to the laser
-#laser_module_tickle_power                    0.0              # this duty cycle will be used for travel moves to keep the laser active without actually burning
+laser_module_enable                          false            # Whether to activate the laser module at all. All configuration is
+                                                              # ignored if false.
+#laser_module_pin                             2.5             # this pin will be PWMed to control the laser. Only P2.0 - P2.5, P1.18, P1.20, P1.21, P1.23, P1.24, P1.26, P3.25, P3.26
```
wolfmanjm: wrong pin

ConfigSamples/AzteegX5Mini/config
```diff
@@ -61,10 +61,16 @@ extruder_en_pin                              0.4              # Pin for extruder
 delta_current                                1.0              # Extruder stepper motor current
 
 # Laser module configuration
-laser_module_enable                          false            # Whether to activate the laser module at all. All configuration is ignored if false.
-#laser_module_pin                             2.7              # this pin will be PWMed to control the laser
-#laser_module_max_power                       0.8              # this is the maximum duty cycle that will be applied to the laser
-#laser_module_tickle_power                    0.0              # this duty cycle will be used for travel moves to keep the laser active without actually burning
+laser_module_enable                          false            # Whether to activate the laser module at all. All configuration is
+                                                              # ignored if false.
+#laser_module_pin                             2.5             # this pin will be PWMed to control the laser. Only P2.0 - P2.5, P1.18, P1.20, P1.21, P1.23, P1.24, P1.26, P3.25, P3.26
```
wolfmanjm: wrong pin

arthurwolf:
Anybody has something against this being merged in ? Looks clean now.

wolfmanjm:
fine with me

PR #689, opened 2015-07-25
Set the ADC clock div to 8 so we actually get 1000 samples/sec on the…

wolfmanjm:
… ADC instead of 6127 samples/sec

Anyone see why not? We were running an ADC rate of around 6127, now it is 1000 as specified.


wolfmanjm:
Fixed the switch_type config to default to pwm so it is still backwards compatible with legacy config files.
In the future the default may be deprecated back to none.

PR #690, opened 2015-07-26
Add adc oversampling and filtering and averaging

wolfmanjm:
Refactored ADC to handle the filtering and averaging etc, removed from Thermistor.
Added better checks in Thermistor for open circuit thermistor, not just max ADC value.
Added oversampling to increase ADC resolution by 2 bits to 14bits.
Added better filtering for raw ADC values before they get oversampled.
Average last 4 oversampled readings when thermistor reads them, this flattens out the rate of change and improves the PID performance.
Removed the race condition on original ADC reads that could cause bad ADC values to randomly appear.

NOTE rerunning PID autotune is recommended after this change.


triffid:
looks awesome, nice work

PR #694, opened 2015-08-14
Allow auto pid tuning to display its output, but allow host to still get temps

wolfmanjm:


wolfmanjm:
Tested on current pronterface (and a really old one)
and master octoprint.

PR #695, opened 2015-08-16
Add speed limiting based on maximum extrusion rate in mm^3/sec

wolfmanjm:
Also adds speed limiting based on extruder max_speed in mm/sec.
Adds M203 Vxxx which sets the maximum Extruder volumetric rate to xxx mm³/sec

If a planned line exceeds the specified volumetric rate then the feedrate of the line is lowered to a point where the max volumetric setting in not exceeded, same with max_speed for the extruder.

M203 E0 will turn off the extruder speed check
M203 V0 will turn off volumetric limiting if in volumetric extrusion mode
defaults are M203 E50 (set in config) and V0
For now volumetric speed limit can only be set with M203 Vxxx (no config setting)


qharley:
Good idea

PR #697, opened 2015-08-20
ZGrid fine grid fix

qharley:
Added constant equal to half the multiplier to ensure the conditions are not foiled by floating point problems.


qharley:
That was quick!

[image: photo]
*Quentin Harley*
Founder, Reprap Morgan 3D

PR #702, opened 2015-08-25
Added max temp check, if exceeeded hotend will be turned off and HALT…

wolfmanjm:
… asserted

Added M143 S0 P300 which will set max temp for hotend 0 to 300 degrees C
This may be a breaking change as previously max_temp simply stopped one from entering a temp that was too high
default is 300

### DIFF
--- a/ConfigSamples/AzteegX5Mini.delta/config
+++ b/ConfigSamples/AzteegX5Mini.delta/config
@@ -87,6 +87,8 @@ temperature_control.hotend.thermistor        EPCOS100K        # see http://smoot
 temperature_control.hotend.set_m_code        104              #
 temperature_control.hotend.set_and_wait_m_code 109            #
 temperature_control.hotend.designator        T                #
+#temperature_control.hotend.max_temp         300              # Set maximum temperature - Will prevent heating above 300 by default
+#temperature_control.hotend.min_temp         0                # Set minimum temperature - Will prevent heating below 0 by default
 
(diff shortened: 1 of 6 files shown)


wolfmanjm:
Needs someone else to test

qharley:
This could be problematic for me... Is it not possible to keep the "stupid" check and add a new config setting for the max limit temp, defaulted at 300?

I found that limiting the hotend temperature setting to a high safe level makes the hotends last much longer.  As a group DIY enthusiast love breaking rules.  While it is a good thing when innovating new technologies, its no good if you run too close to a temperature that can now cause a print failure.

wolfmanjm:
Why is it problematic? if the temp goes above the setting it halts? that is a very important safety feature.
If you don;t like it set it to 1000

qharley:
No, not what I meant.

I simply propose that you can still set a soft limit at whatever
temperature, like 250 to curb enthusiasm.  One of my clients come to mind
immediately....
Previously it would simply not go there.  Now the print will fail.  See the
difference?


[image: photo]
*Quentin Harley*
Founder, Reprap Morgan 3D

wolfmanjm:
It was causing a lot of confusion, people expected it to shutoff at that temperature, and at least one person just lost a jhead because it did not.
It now does what most people think and expect it to do.

qharley:
I like the new safety feature a lot. I wanted it a long time ago. I'll add
the soft limit into my own branch. Thanks

PR #705, opened 2015-08-25
zprobe: add configuration via M670

sorki:
Also adds support for storing probe configuration
to config-override via M500

;Probe feedrates slow/fast/return (mm/sec):
M670 S0.50 K2.00 R2.50
;Probe max_z (mm):
M670 Z200.00
;Probe height (mm):
M670 H2.00

Depends on #704.

### DIFF
--- a/src/modules/tools/zprobe/ZProbe.cpp
+++ b/src/modules/tools/zprobe/ZProbe.cpp
@@ -313,17 +313,36 @@ void ZProbe::on_gcode_received(void *argument)
 
     } else if(gcode->has_m) {
         // M code processing here
-        if(gcode->m == 119) {
-            int c = this->pin.get();
-            gcode->stream->printf(" Probe: %d", c);
-            gcode->add_nl = true;
+        int c;
+        switch (gcode->m) {
+            case 119:
+                c = this->pin.get();
+                gcode->stream->printf(" Probe: %d", c);
+                gcode->add_nl = true;
+                break;
 
-        }else {
-            for(auto s : strategies){
-                if(s->handleGcode(gcode)) {
-                    return;
+            case 670:
+                if (gcode->has_letter('S')) this->slow_feedrate = gcode->get_value('S');
+                if (gcode->has_letter('K')) this->fast_feedrate = gcode->get_value('K');
+                if (gcode->has_letter('R')) this->return_feedrate = gcode->get_value('R');
+                if (gcode->has_letter('Z')) this->max_z = gcode->get_value('Z');
+                if (gcode->has_letter('H')) this->probe_height = gcode->get_value('H');
+                break;
+
+            case 500: // save settings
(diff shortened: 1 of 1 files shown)


sorki:
removed mark_as_taken stuff

src/modules/tools/zprobe/ZProbe.cpp
```diff
@@ -287,17 +287,35 @@ void ZProbe::on_gcode_received(void *argument)
...
+        switch (gcode->m) {
+            case 119:
+                c = this->pin.get();
+                gcode->stream->printf(" Probe: %d", c);
+                gcode->add_nl = true;
+                break;
+
+            case 670:
+                if (gcode->has_letter('S')) this->slow_feedrate = gcode->get_value('S');
+                if (gcode->has_letter('F')) this->fast_feedrate = gcode->get_value('F');
```
wolfmanjm: cannot use F here as it is globally reserved to mean G0/G1 feedrate which can affect other things.

src/modules/tools/zprobe/ZProbe.cpp
```diff
@@ -287,17 +287,35 @@ void ZProbe::on_gcode_received(void *argument)
...
+                if (gcode->has_letter('H')) this->probe_height = gcode->get_value('H');
+                break;
+
+            case 500: // save settings
+            case 503: // print settings
+                gcode->stream->printf(";Probe feedrates slow/fast/return (mm/sec):\nM670 S%1.2f F%1.2f R%1.2f\n",
+                    this->slow_feedrate, this->fast_feedrate, this->return_feedrate);
+                gcode->stream->printf(";Probe max_z (mm):\nM670 Z%1.2f\n", this->max_z);
+                gcode->stream->printf(";Probe height (mm):\nM670 H%1.2f\n", this->probe_height);
+
```
wolfmanjm: add comment here to indicate fall through is intended.

src/modules/tools/zprobe/ZProbe.cpp
sorki: Your suggestion? I can't come up with better fitting letter :)

sorki:
Now using T instead of F, added note.
"; Probe feedrates Slow/fasT/Return (mm/sec):\nM670 S%1.2f T%1.2f R%1.2f\n",

Tested with todays edge, ZGrid works too.

wolfmanjm:
T is also a reserved word.. it is tool change it is not a valid parameter. How about K or Q

sorki:
K sounds good!

wolfmanjm:
Is there a specific reason for splitting the M670 into three lines rather than one? 
I'd prefer to keep it to one line if possible unless there is an overriding reason otherwise.

wolfmanjm:
Never mind I'll change that myself unless you have a valid objection.

PR #715, opened 2015-09-06
Increase Z Home Ofs menu setting precision to 0.01 mm

OskarLinde:
Z Home Ofs is useful for fine adjustment of the "paper thickness" offset between head and bed, but the menu option needs higher precision for this to be usable from the panel. 10 µm seems like a reasonable level of precision for tweaking the first layer offset.


wolfmanjm:
How about 0.05 as a compromise? i think 0.01 is too fine and requires too many encoder turns if it needs to be changed by a few mm.

wolfmanjm:
NOTE I may remove this all togther and replace with either M306 Z0 or the baby step thing.

PR #723, opened 2015-09-09
Allow fractional ms to be specified for HW PWM 

wolfmanjm:


wolfmanjm:
oops included unrelated changes :)

PR #713, opened 2015-09-06
Generic calibration strategy

OskarLinde:
Here's a simple implementation of delta calibration that does a single probe pass and computes optimal model parameters to fit the probe results. It also supports bed leveling using a proper rotation transformation rather than the skew transform the other bed mapping strategies use.

The general idea is to store actuator positions for each probe trigger point and then optimize the model parameters until these actuator positions all correspond as closely as possible to z=0 in software. The model parameters are optimized without any knowledge of their meaning, each parameters influence on the solution is dynamically measured by nudging them one by one. (This means that the same method could be applied to other future arm solutions as well.) The model parameters are optimized using a iterative nonlinear solver using the [Levenberg-Marquardt](https://en.wikipedia.org/wiki/Levenberg–Marquardt_algorithm) method.

The general syntax is:

`G32 <variables to optimize> Pp Oo`

| Parameter | &nbsp; |
| --- | --- |
| P | number of probe points, default = 50, min = number of variables |
| O | number of repeats for each probe point (helps flush out false probe results), default = 1 |

Variables available to all arm solutions:

| Variable | &nbsp; |
| --- | --- |
| X,Y,Z | endstop trim |
| U,V | plane tilt |
| W | plane offset (mutually exclusive with X,Y,Z) |

For LinearDelta:

| Variable | &nbsp; |
| --- | --- |
| R | delta radius |
| A,B,C | tower offsets |
| D,E,H | tower angle |

The calibration respects Home Offset Z meaning one can keep Home Offset Z set to "paper thickness". The offset as defined unfortunately needs to be negative. Something to remember.
##### Examples:

| Gcode&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; | &nbsp; |
| --- | --- |
| `G32 X0 Y0 Z0` | Calibrate endstops, 50 probe points |
| `G32 U0 V0 W0 N7` | Bed leveling (plane orientation and offset), 7 probe points |
[...]

PR #725, opened 2015-09-10
Bugfix/robot transformed milestone

OskarLinde:
Basically, the way to think about the coordinate systems is that "last_milestone" is STL coordinates, and "transformed_last_milestone" is machine coordinates.

This bug can be triggered on a delta by applying a compensation transform that makes a large positive increase in Z and then do a G28 home followed by a G1 Z10. Upon G1 Z10 it will crash into the endstops.

NOTE: I can't guarantee that there are no additional places the two milestones get out of sync.

### DIFF
--- a/src/modules/robot/Robot.cpp
+++ b/src/modules/robot/Robot.cpp
@@ -245,6 +245,7 @@ void Robot::on_config_reload(void *argument)
     // so the first move can be correct if homing is not performed
     float actuator_pos[3];
     arm_solution->cartesian_to_actuator(last_milestone, actuator_pos);
+    // TODO: transformed_milestone???
     for (int i = 0; i < 3; i++)
         actuators[i]->change_last_milestone(actuator_pos[i]);
 
@@ -639,8 +640,14 @@ void Robot::reset_axis_position(float x, float y, float z)
     this->transformed_last_milestone[Y_AXIS] = y;
     this->transformed_last_milestone[Z_AXIS] = z;
 
+    // check function pointer and call if set to transform the target to compensate for bed
+    if(compensationTransform) {
+        // some compensation strategies can transform XYZ, some just change Z
+        compensationTransform(transformed_last_milestone);
+    }
+
     float actuator_pos[3];
-    arm_solution->cartesian_to_actuator(this->last_milestone, actuator_pos);
+    arm_solution->cartesian_to_actuator(this->transformed_last_milestone, actuator_pos);
     for (int i = 0; i < 3; i++)
         actuators[i]->change_last_milestone(actuator_pos[i]);
 }
@@ -649,10 +656,17 @@ void Robot::reset_axis_position(float x, float y, float z)
 void Robot::reset_axis_position(float position, int axis)
 {
     this->last_milestone[axis] = position;
-    this->transformed_last_milestone[axis] = position;
+    for (int i = 0; i < 3; i++)
+        this->transformed_last_milestone[i] = this->last_milestone[axis];
+
+    // check function pointer and call if set to transform the target to compensate for bed
+    if(compensationTransform) {
+        // some compensation strategies can transform XYZ, some just change Z
+        compensationTransform(transformed_last_milestone);
+    }
 
     float actuator_pos[3];
-    arm_solution->cartesian_to_actuator(this->last_milestone, actuator_pos);
+    arm_solution->cartesian_to_actuator(this->transformed_last_milestone, actuator_pos);
 
     for (int i = 0; i < 3; i++)
         actuators[i]->change_last_milestone(actuator_pos[i]);
@@ -662,11 +676,14 @@ void Robot::reset_axis_position(float position, int axis)
 void Robot::reset_position_from_current_actuator_position()
 {
     float actuator_pos[]= {actuators[X_AXIS]->get_current_position(), actuators[Y_AXIS]->get_current_position(), actuators[Z_AXIS]->get_current_position()};
-    arm_solution->actuator_to_cartesian(actuator_pos, this->last_milestone);
-    memcpy(this->transformed_last_milestone, this->last_milestone, sizeof(this->transformed_last_milestone));
+    arm_solution->actuator_to_cartesian(actuator_pos, this->transformed_last_milestone);
+    memcpy(this->last_milestone, this->transformed_last_milestone, sizeof(this->transformed_last_milestone));
+    if (compensationTransformInverse) {
+        compensationTransformInverse(last_milestone);
+    }
 
     // now reset actuator correctly, NOTE this may lose a little precision
-    arm_solution->cartesian_to_actuator(this->last_milestone, actuator_pos);
+    arm_solution->cartesian_to_actuator(this->transformed_last_milestone, actuator_pos);
     for (int i = 0; i < 3; i++)
         actuators[i]->change_last_milestone(actuator_pos[i]);
 }
--- a/src/modules/robot/Robot.h
+++ b/src/modules/robot/Robot.h
@@ -48,6 +48,7 @@ class Robot : public Module {
 
         // set by a leveling strategy to transform the target of a move according to the current plan
         std::function<void(float[3])> compensationTransform;
+        std::function<void(float[3])> compensationTransformInverse;
 
         struct {
             bool inch_mode:1;                                 // true for inch mode, false for millimeter mode ( default )
--- a/src/modules/tools/zprobe/ThreePointStrategy.cpp
+++ b/src/modules/tools/zprobe/ThreePointStrategy.cpp
@@ -319,10 +319,12 @@ void ThreePointStrategy::setAdjustFunction(bool on)
 {
     if(on) {
         // set the compensationTransform in robot
-        THEKERNEL->robot->compensationTransform= [this](float target[3]) { target[2] += this->plane->getz(target[0], target[1]); };
+        THEKERNEL->robot->compensationTransform        = [this](float target[3]) { target[2] += this->plane->getz(target[0], target[1]); };
+        THEKERNEL->robot->compensationTransformInverse = [this](float target[3]) { target[2] -= this->plane->getz(target[0], target[1]); };
     }else{
         // clear it
         THEKERNEL->robot->compensationTransform= nullptr;
+        THEKERNEL->robot->compensationTransformInverse= nullptr;
     }
 }
 
--- a/src/modules/tools/zprobe/ZGridStrategy.cpp
+++ b/src/modules/tools/zprobe/ZGridStrategy.cpp
@@ -632,10 +632,12 @@ void ZGridStrategy::setAdjustFunction(bool on)
 {
     if(on) {
         // set the compensationTransform in robot
-        THEKERNEL->robot->compensationTransform= [this](float target[3]) { target[2] += this->getZOffset(target[0], target[1]); };
+        THEKERNEL->robot->compensationTransform=        [this](float target[3]) { target[2] += this->getZOffset(target[0], target[1]); };
+        THEKERNEL->robot->compensationTransformInverse= [this](float target[3]) { target[2] -= this->getZOffset(target[0], target[1]); };
     }else{
         // clear it
         THEKERNEL->robot->compensationTransform= nullptr;
+        THEKERNEL->robot->compensationTransformInverse= nullptr;
     }
 }
 


wolfmanjm:
transformed_last_milestone IS NOT machine coordinates by any stretch. it is the last milestone compensated by any Z adjustment made by the levelling strategies.
Also I don't see why it is an inverse transform.. that also makes no sense to me.
You would do the same compensation as you would for a XYZ coming from gcode. IMO

OskarLinde:
The way one can consider transformed_last_milestone machine coordinates is the following: last_milestone is the coordinate given by the gcode command. The compensation transform is applied to this to get transformed_last_milestone. The robot is then moved to the position that corresponds to arm_solution->cartesian_to_actuator(transformed_last_milestone,...). I.e. actuators <=> transformed_last_milestone and last_milestone is the virtual gcode coordinate

OskarLinde:
We are homing to machine coordinates, so to calculate what those coordinates coorespond to int gcode coordinates, we have to apply the inverse transform

wolfmanjm:
The only difference is subtracting instead of adding right? that does not warrant adding another function pointer, this is extremely inefficient. I cannot accept it in this form. I'll have to think about a better way .

wolfmanjm:
I have no idea what you mean by machine coordinates, that term is meaningless. they are both cartesian coordinates, and one is compensated. there are cartesian coordinates and actuator coordinates. the compensated coordinates are a hack for ABL. the reason both are needed is to correctly calculate the distance the actuators need to move.

OskarLinde:
No, if we want to support generic compensationTransforms (which I assume since both the comment explicitly states that and the function takes and returns a float[3]) you can not in general assume that it is only a addition->subtraction.

Also, extremely inefficient is a bit hyperbolic. The inverse code is only called when you do reset_axis_position which is once per G28 while the normal transform is called for each line segment. I can not by any stretch of imagination understand how that would have any performance effect at all.

OskarLinde:
We have 3 coordinate systems in the code. 
1. gcode coordinates, that are given by the gcode,
2. transformed coordinates, that are used by the robot
3. actuator coordinates that are used by the motors

Since 2 directly maps to 3 I call them machine coordinates. You are free to give them any other name.

wolfmanjm:
Inefficient in terms of memory. Function pointers use a lot of memory they are not just a pointer.
I think adding a bool to the compensation transform lambda that indicates transform vs inverse transform will be more memory efficient.
Did I mention we are extremely low on runtime memory? Every byte literally counts.
Anyway I need more time to digest this as it changes quite a bit of somewhat complex interactions, and I suspect you have not tested this on a cartesian using threepoint levelling or on a scara using  zgrid levelling.

OskarLinde:
Yeah, a bool argument would work. And you are correct, it is not properly tested by any stretch of imagination. :-)

OskarLinde:
Without thinking too much about it, it feels that the only thing regular last_milestone variable is used for is to handle relative Gcode moves. Are there any other uses?

wolfmanjm:
M114 uses last_milestone and then search all the places that use get_axis_position(), it is used in a lot of places. I suspect some of those actually need transformed_last_milestone instead, so this is really a potentially breaking change IMO.
You are going to need to examine every place it is used and see which one needs to be used.
M114 will need to return last_milestone and not transforned_last_milestone though so that is one of your list :)

wolfmanjm:
I still do not see why an inverse is being used... on G92 (or reset_axis_position()) we set the last_milestone to 0,0,0 why is transformed_last_milestone not set to compensationTransform([0,0,0]);
It seems to me it is no different from doing G1 X0 Y0 Z0, which would also do compensationTransform([0,0,0]); please explain in more detail why inverse is used here?

OskarLinde:
You are correct, reset_axis_position() should use compensationTransform and not its inverse. I had in mind that home-position should be independent on compensation transform, but that is clearly not correct. (This has a number of other consequences, including the soft-endstop PR)

That leaves reset_position_from_current_actuator_position() as the place that needs the inverseCompensationTransform.

Re: get_axis_position(), of the different calls, most seem ok, but most places that uses this value to call cartesian_to_actuator() does the wrong thing:

SCARAcal.cpp, four different places that call get_axis_position() all need to apply compensationTransform
SCARAcal.cpp: 135 should apply compensationTransform to home_off
Touchprobe.cpp:132 calls cartesian_to_actuator without compensationTransform
Robot.cpp:327 forgets about transformed_last_milestone
Robot.cpp:247 ?

My suggestion is that we implement a ```Robot::axis_to_actuator()``` and ```Robot::actuator_to_axis()```
to complement ```Robot::get_axis_position()```, and make sure to call that instead of cartesian_to_actuator() in most places outside of Robot.

OskarLinde:
Or maybe my initial version was mostly, and home-position should be independent on compensationTransform. In which case I instead broke G92.

OskarLinde:
So one could chose to do it both ways, but this is the way I think makes most sense: In this case, a G28 needs to go via inverseCompensationTransform to populate last_milestone.

wolfmanjm:
I think this is all wrong sorry. There need be no  inverseCompensationTransform IMO. I think you are looking at this entirely wrong and making it way more complicated than it needs to be.

OskarLinde:
I would be very happy to be proven wrong, but last_milestone and transformed_last_milestone need to be kept in sync... reset_position_from_current_actuator_position() will need to do inverseCompensationTransform() to recover last_milestone, or how would you do it otherwise?

OskarLinde:
And what should G28 followed by M114 print if you have a compensation transform?

wolfmanjm:
well if you think it through you will see the error of your ways :)
M114 always reports the position it was told to go to. in the case of M28 it will be 0,0,0 ( or the G28 parameters) always. and G28 zeros both last_milestone and the transformed one too, it is not necessary to compensate that as it will be compensated on the first move. So basically this is entire PR is not really needed. Complicates everything and breaks many things. I think the original bug in your code you were trying to fix with this was a different bug altogether based on not using relative probes points as is required. Then you just complicated it more and more. At this point you need to step back and look at it in the context that has nothing to do with your ABL algorithm. I think you will see this is unnecessary.

OskarLinde:
Problem is, I have thought about it  :-) And I have everything working fine on my side with my fixes so I know they work. The question is if there is a simpler fix.

So, G28 M114 on my delta displays: X:0, Y:0, Z:170, is that wrong?

And, I can get the bug that this PR fixes to trigger without using my autocalibration code. See the first post:

>This bug can be triggered on a delta by applying a compensation transform that makes a large positive increase in Z and then do a G28 home followed by a G1 Z10. Upon G1 Z10 it will crash into the endstops.

So the compensation on the first move strategy clearly doesn't work.

Basically what you are saying is, let's ignore that last_milestone is incorrect after G28 and after Resume, it will be fixed after the first move. Two problems with that: 1) If the first move is relative, it will move to the wrong position (target = last_milestone + gcode_coord), and 2) If the first move gets segmented into more than 1 segment (i.e. a delta move), the first segment (and thus all following) is incorrectly computed. The latter bug can be also fixed by changing all computations to be done in transformed_coordinates.

Are you supposed to be allowed to do a relative move after a G28?

src/modules/robot/Robot.cpp
```diff
@@ -662,11 +676,14 @@ void Robot::reset_axis_position(float position, int axis)
 void Robot::reset_position_from_current_actuator_position()
 {
     float actuator_pos[]= {actuators[X_AXIS]->get_current_position(), actuators[Y_AXIS]->get_current_position(), actuators[Z_AXIS]->get_current_position()};
-    arm_solution->actuator_to_cartesian(actuator_pos, this->last_milestone);
-    memcpy(this->transformed_last_milestone, this->last_milestone, sizeof(this->transformed_last_milestone));
```
wolfmanjm: do this the other way around and you do not need the inverse transform.

wolfmanjm:
I still think you are wrong on most of this. I need to think it through and see where it breaks the other levelling strategies. The three point will break as it uses relative probes and it works fine now, so this can't be a huge issue. Just because it works for you does not mean it works, I do not think you have considered the other applications, especially the cartesian applications.

I will concede that maybe the transformed_last_milestone needs to be compensated after G28. however the inverse is never required and needs to be removed.
I think this will break the threepoint leveling but I need to think through how and why. as threepoint currently works this has to break it.

OskarLinde:
I have not yet made the necessary changes to make 3-point (and SCARACalibration, but I have no way of testing that) working properly.

I would love to get rid of the inverse transform if possible. Maybe it can be done by changing home_offs to compensate for whatever the compensation transform does at that position.

qharley:
Scaracal works exactly as intended.  Please do not change it.  The section
you refer to translates the trim offset into actual home offsets.  This
prevents the home and trim movement on the Morgan that can lead to trouble
when you home twice in a row..
The scara module already use forward transforms to calculate targets for
calibration.

OskarLinde:
@qharley, I don't think there is much risk of breakage. Basically, the bug this PR addresses only occurs when you have a ```compensationTransform``` installed. Without a compensation transform everything works as before. Basically this PR fixes:

* With a compensationTransform, G28 followed by a relative move moves to the wrong position.
* With a compensationTransform, G92 didn't work.
* With a compensationTransform, the first move on a delta (or other configurations that lead to segmentation) created a spurious move that could end up crashing into the endstops leading to stepper skipping and loss of position registration.

The *unrelated* and only potential bug in SCARACal is that it does get_axis_position() followed by cartesian_to_actuator() which gives the wrong result if a compensationTransform is installed. The lowest risk fix is to simply delete/disable compensationTransforms before SCARA calibration.

qharley:
Why would you like to transform Theta and Psi in the scara kinematics
anyway.  It does not influence Z in any way.
I am using the ZGrid module I wrote for the morgan.  How does using
compensationTransform make it better?

OskarLinde:
The ZGrid module implements a local skew transform as compensationTransform. That is, it preserves X and Y and only changes Z. Since the SCARA kinematic's theta and psi are invariant to Z, ZGrid probably didn't cause any issue. In #713 the compensationTransform does a proper rotation compensation which makes X' and Y' dependent on Z.

As for what makes it better, it rotates rather than skews the model to fit the plane.

The bug in SCARACal is basically that all

    THEKERNEL->robot->get_axis_position(cartesian);    // get actual position from robot
    THEKERNEL->robot->arm_solution->cartesian_to_actuator( cartesian, actuators );      // translate to get actuator position

Should be

    THEKERNEL->robot->get_axis_position(cartesian);    // get actual position from robot
    if (THEKERNEL->robot->compensationTransform)
        THEKERNEL->robot->compensationTransform(cartesian);
    THEKERNEL->robot->arm_solution->cartesian_to_actuator( cartesian, actuators );      // translate to get actuator position

BTW, I like the calibration procedure in SCARACal, we should add something similar for delta calibration of arm lengths and bed center.

Lodran:
There's a bug in void Robot::reset_axis_position(float position, int axis)

    this->last_milestone[axis] = position;
    for (int i = 0; i < 3; i++)
        this->transformed_last_milestone[i] = this->last_milestone[axis];

should be:

    this->last_milestone[axis] = position;
    for (int i = 0; i < 3; i++)
        this->transformed_last_milestone[i] = this->last_milestone[i];

arthurwolf:
@OskarLinde Nice find. Would you mind making this a separate pull request as this pull request is a more general discution about the transforms ?

Lodran:
@arthurwolf : Actually, I noted it here because the bug only exists in this pull request, not in the current Edge branch.

arthurwolf:
Yep sorry I missed that

On Tue, Oct 27, 2015 at 6:57 AM, Lodran <notifications@github.com> wrote:

> @arthurwolf <https://github.com/arthurwolf> : Actually, I noted it here
> because the bug only exists in this pull request, not in the current Edge
> branch.
>
> —
> <https://github.com/Smoothieware/Smoothieware/pull/725#issuecomment-151383125>
> .
>



-- 
Courage et bonne humeur.

OskarLinde:
@Lodran, well spotted! That is certainly a bug.

wolfmanjm:
Can you remove the inverse compensation stuff as it is only used for reset_position_from_current_actuator_position() which is only used for abort so is not critical. 
We can revisit the need for that in another PR.
Also fix the bug mentioned above.

If you do that I will merge this.

wolfmanjm:
This I think will be made obsolete by #804 
It is still debatable whether homing sets the machine position before the compensation transform, and that is the way I have done it. I then apply the compensation transform to the homed coordinates to get the machine position. IMO compensation transform should probably be turned off when homing anyway.
The only place left that would require the inverse compensation transform is the call made when aborting to recover the current position, and I think that is not that important.

As a side note I believe you were totally correct on most of these issues, sorry for being such a hard ass about it :)

src/modules/robot/Robot.cpp
```diff
@@ -639,8 +640,14 @@ void Robot::reset_axis_position(float x, float y, float z)
     this->transformed_last_milestone[Y_AXIS] = y;
     this->transformed_last_milestone[Z_AXIS] = z;
 
+    // check function pointer and call if set to transform the target to compensate for bed
+    if(compensationTransform) {
```
wolfmanjm: FWIW I still think this is incorrect... homing did not use compensation transform, it set the actuators to a known position before compensation so both last_milestone and transformed_milestone are the same thing. *Homing does not use robot so there is no way for it to apply compensation). You can't then apply compensation transform without actually moving the head to match what you just set the actuators to.

wolfmanjm:
FWIW the refactor I did to the motion control made this issue more important as homing and zprobes no longer used direct stepper drives but all go through the planner and thus the whole machine coordinate to compensated coordinate stuff.
Plus having added real workspace coordinate systems and making G92 do the correct thing rather than reset axis changed the whole structure of the code.
So in summary I eventually did add inverse compensation transform so when we display realtime positions (which were not available at the time of this issue) we can derive the actual position from the actuator all the way back to the workspace coordinates.
It also appears that Homing did require an inverse compensation transform to get the MCS position from the given actuator coordinates. So Oskar was correct yet again :) (and I was wrong, yet again).
I am still struggling with how to get M306 to work in this new world and what to call the different coordinate systems, as technically machine coordinates (as per linuxcnc) are pre compensation, but the code is currently confusing with its naming of these.

OskarLinde:
That's cool! It was too long since I looked at this, so I can't help much
with naming. The potentially noneuclidian frame you can end up with after a
compensation transform is especially tricky to define things relative to.

PR #726, opened 2015-09-10
Make soft endstops truly soft

OskarLinde:
This changes the physical motion of the actuators when applying endstop-trim into a software compensation. This makes positive endstop trim values not cause G28 to crash into the endstops.

### DIFF
--- a/src/modules/tools/endstops/Endstops.cpp
+++ b/src/modules/tools/endstops/Endstops.cpp
@@ -28,6 +28,7 @@
 #include "StreamOutputPool.h"
 #include "Pauser.h"
 #include "StepTicker.h"
+#include "BaseSolution.h"
 
 #include <ctype.h>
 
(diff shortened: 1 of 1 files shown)


wolfmanjm:
I like this, it was the way I originally wanted to do it, except we did not have Forward Kinematics so could not do it this way.

However this will break a few peoples configs as they (mis-)use trim to move the carriages off the endstops so they can enable limits.

I am torn about this as I have warned and documented to NOT use limits on a delta this way. OTH I don't like to break peoples configs. Most changes must be backwards compatible, so I need to think about this a bit. It is tempting to break it for those that ignored my warnings :) but probably not a great idea.
However it is not backwards compatible which is a bit of an issue.
@arthurwolf any comments on that?

OskarLinde:
Isn't that something that Endstops::back_off_home() already does?

wolfmanjm:
Endstops::back_off_home() does not work for deltas at the moment, if you look it moves X Y Z cartesian, for a delta it would just need to move in Z.

I'd like to merge this PR, but need to fix Endstops::back_off_home() for deltas first. which requires a separate PR and testing.

wolfmanjm:
Ok I have fixed Endstops::back_off_home() so it works for deltas. so now it should be backward compatible.

PR #734, opened 2015-09-13
Oskar linde feature/soft soft endstop fixed limits

wolfmanjm:
Also fixed limit switches working with Deltas. No extra trim needed now.


wolfmanjm:
I think one thing people are not going to like with this is that after homing X and Y are not 0 with M114.

PR #738, opened 2015-09-21
Improved ZGridStrategy:

RTimothyEdwards:
(1) Removed a redundant and incorrect move in the Z-probe.  If
cal_offset and/or probe_offset non-zero, then the calibration
loop first moves the extruder to the grid position, then makes
the correct move.  The first (incorrect) move has been removed.

(2) Corrected the getZOffset() routine, which would incorrectly
compute xIndex and yIndex if cal_offset_x or cal_offset_y is
negative ((int) is not the floor of a negative number).

(3) Improved getZOffset() by extrapolating points outside the
grid using the nearest grid square, such that for purposes of
correction, the plane of each outermost grid square effectively
extends to infinity.  The previous code returned zero for all
areas outside the grid, creating a discontinuity at the grid
boundary.

These changes were made especially for using the ZGridStrategy
with a cartesian printer and the recently popular inductive
proximity sensor.  The proximity sensor must have a metal surface
underneath at all times, so the probe calibration must be set to
a smaller area than the bed plate.  The cal_offset_x and
cal_offset_y plus the bed_x and bed_y serve to create an interior
area for the probe grid.  The new getZOffset() will extrapolate
values to first order from the probe grid out to the edge of the
bed plate.

Changes should not interfere with any existing use of
ZGridStrategy.

### DIFF
--- a/src/modules/tools/zprobe/ZGridStrategy.cpp
+++ b/src/modules/tools/zprobe/ZGridStrategy.cpp
@@ -522,10 +522,6 @@ bool ZGridStrategy::doProbing(StreamOutput *stream)  // probed calibration
         } else {
             this->next_cal();                                     // to not cause damage to machine due to Z-offset
         }
-	// probeDistance() moves to the correct point, so this
-	// move is not only redundant, but wrong.
-        // this->move(this->cal, slow_rate);                         // move to the next position
-
         this->pData[pindex] = z ;                                 // save the offset
     }
 
@@ -662,14 +658,6 @@ float ZGridStrategy::getZOffset(float X, float Y)
     int xIndex = (int)(floorf(xdiff));	// Get the current sector (X)
     int yIndex = (int)(floorf(ydiff));	// Get the current sector (Y)
     
-    // * Care taken for table outside boundary
-    // * Returns zero output when values are outside table boundary
-    // if(xIndex < 0 || xIndex > (this->numRows - 1) || yIndex < 0
-    //    || yIndex > (this->numCols - 1))
-    // {
-    //   return (0);
-    // }
-
     // Index bounds limited to be inside the table
     if (xIndex < 0) xIndex = 0;
     else if (xIndex > (this->numRows - 2)) xIndex = this->numRows - 2;
@@ -677,14 +665,7 @@ float ZGridStrategy::getZOffset(float X, float Y)
     if (yIndex < 0) yIndex = 0;
     else if (yIndex > (this->numCols - 2)) yIndex = this->numCols - 2;
 
-    // if (xIndex == (this->numRows - 1))
-    //     xIndex2 = xIndex;
-    // else
     xIndex2 = xIndex+1;
-
-    // if (yIndex == (this->numCols - 1))
-    //     yIndex2 = yIndex;
-    // else
     yIndex2 = yIndex+1;
 
     xdiff -= xIndex;                    // Find floating point
--- a/src/modules/tools/zprobe/ZGridStrategy.cpp
+++ b/src/modules/tools/zprobe/ZGridStrategy.cpp
@@ -522,7 +522,9 @@ bool ZGridStrategy::doProbing(StreamOutput *stream)  // probed calibration
         } else {
             this->next_cal();                                     // to not cause damage to machine due to Z-offset
         }
-        this->move(this->cal, slow_rate);                         // move to the next position
+	// probeDistance() moves to the correct point, so this
+	// move is not only redundant, but wrong.
+        // this->move(this->cal, slow_rate);                         // move to the next position
 
         this->pData[pindex] = z ;                                 // save the offset
     }
@@ -654,26 +656,36 @@ float ZGridStrategy::getZOffset(float X, float Y)
 
     float dCartX1, dCartX2;
 
-    int xIndex = (int) xdiff;                  // Get the current sector (X)
-    int yIndex = (int) ydiff;                  // Get the current sector (Y)
+    // Get floor of xdiff.  Note that (int) of a negative number is its
+    // ceiling, not its floor.
 
+    int xIndex = (int)(floorf(xdiff));	// Get the current sector (X)
+    int yIndex = (int)(floorf(ydiff));	// Get the current sector (Y)
+    
     // * Care taken for table outside boundary
     // * Returns zero output when values are outside table boundary
-    if(xIndex < 0 || xIndex > (this->numRows - 1) || yIndex < 0
-       || yIndex > (this->numCols - 1))
-    {
-      return (0);
-    }
+    // if(xIndex < 0 || xIndex > (this->numRows - 1) || yIndex < 0
+    //    || yIndex > (this->numCols - 1))
+    // {
+    //   return (0);
+    // }
 
-    if (xIndex == (this->numRows - 1))
-        xIndex2 = xIndex;
-    else
-        xIndex2 = xIndex+1;
+    // Index bounds limited to be inside the table
+    if (xIndex < 0) xIndex = 0;
+    else if (xIndex > (this->numRows - 2)) xIndex = this->numRows - 2;
 
-    if (yIndex == (this->numCols - 1))
-        yIndex2 = yIndex;
-    else
-        yIndex2 = yIndex+1;
+    if (yIndex < 0) yIndex = 0;
+    else if (yIndex > (this->numCols - 2)) yIndex = this->numCols - 2;
+
+    // if (xIndex == (this->numRows - 1))
+    //     xIndex2 = xIndex;
+    // else
+    xIndex2 = xIndex+1;
+
+    // if (yIndex == (this->numCols - 1))
+    //     yIndex2 = yIndex;
+    // else
+    yIndex2 = yIndex+1;
 
     xdiff -= xIndex;                    // Find floating point
     ydiff -= yIndex;                    // Find floating point


arthurwolf:
I only see one commit and one file change, are you sure this includes all the changes you describe in the description ?

RTimothyEdwards:
Yes, the fix is pretty simple in spite of the lengthy commit message.  The change from a bounded grid correction to one that extrapolates out to infinity is surprisingly trivial.  If you prefer cleaned-up code, I'll get rid of the commented-out lines.

wolfmanjm:
@qharley can you check this? make sure it doesn't break anything on your end. If you give a thumbs up I'll merge it. - Thanks

qharley:
I'll try it on a machine tonight or tomorrow morning.  Buried under a
mountain of paperwork...

RTimothyEdwards:
Hello Jim et al.,

> @qharley <https://github.com/qharley> can you check this? make sure it
> doesn't break anything on your end. If you give a thumbs up I'll merge it.

The plane extrapolation outside the grid boundary should be okay;  I've
already checked it on my MakerGear M2.  The removal of the extra "move"
for each grid point check probably could use a bit of close scrutiny.
Since it's a move that happens after the calibration move, it will also
be the last move of the calibration.  this->next_cal() will go to (0, 0)
after the last point, so removing the move() causes the extruder to
stay at the last grid point checked rather than return to (X, Y) home.

I started checking the Z-grid calibration on my printer last night, and
I found that I will need another change to get it working on the M2.  The
main thing I found was that the "z" value returned from
zprobe->probeDistance is equal to the total distance traveled, which
includes the "probe_height".  I set my probe_height to, say, 5mm during
testing to watch the calibration better.  But the "probe_height" is not
subtracted out of the final value in this->pData[], so I end up getting
a Z home offset of ~5mm after calibration, which is wrong.  The Z home
offset after calibration should be the difference between my Z min switch
and the ideal extruder height above the bed, plus any extra adjustment at
the (X, Y) origin caused by normalize_grid().

This may have something to do with my use of the Z-min switch in addition
to the probe, and homing to X, Y, Z first to get a point of reference.  I
don't have much choice with my M2, given that the proximity sensor Z-probe
is sitting out in front of the extruders and so is sitting over air when
the extruder is homed to X and Y, preventing the probe from working.  I
considered other options such as a more complicated homing routine, but
keeping the Z-min switch seemed like the easiest to implement and most
robust.

The result of this observation is that to get a working calibration,
I need to change

	this->pData[pindex] = z;
to
	this->pData[pindex] = z + zprobe->getProbeHeight();

So am I missing something obvious about the way the Z-grid calibration
works, or is this just the difference between the way my M2 and the
proximity sensor works vs. whatever setup the Z-grid algorithm was
originally designed for?

If the latter, is there a way to
differentiate between the types of setup in the calibration code?  If
a difference is caused by the use of Z homing before calibration, for
example, I could check for the use of XYZ homing, or check that pins
are independently defined for both the Z-probe and Z-min switch.  Any
advice would be appreciated.
					Regards,
					Tim

+--------------------------------+-------------------------------------+
| R. Timothy Edwards (Tim)       | email: tim@opencircuitdesign.com    |
| Open Circuit Design            | web:   http://opencircuitdesign.com |
| 19412 Cissel Manor Drive       | phone: (301) 528-5030               |
| Poolesville, MD 20837          | cell:  (408) 828-8212               |
+--------------------------------+-------------------------------------+

wolfmanjm:
@RTimothyEdwards Actually no this is wrong... the probes MUST be relative not absolute as the probe offset is not always know accurately. If you make the probe offset part of the calibration then this will have to remain in your fork.
The use case I want to keep is to have removable probes, which means the z offset is variable and unknown, so all probes are relative to the first probe point.
The way I do this with three point is that after the level is finished and saved you zero z to the first probe point using M306 Z0 and the min z endstop. As smoothie saves the probes result it makes permanent zprobes and probing every print optional.

RTimothyEdwards:
I'm not quite sure which part of my comment you are referring to when you say "no this is wrong".  I've read your response several times and came up with a different interpretation every time.  So bear in mind that I may be completely missing the point.  But I think that the issue here is the difference between a mechanical Z-probe like the one shown in the photo on the Z-probe page of the Smoothieware site, and the inductive proximity sensor I'm using, and which has become popular on a number of 3D printers.  The inductive proximity sensor is bolted to the extruder assembly, so it has an absolute and known offset from the extruder nozzle (with the caveat that because it's not bolted to the nozzle itself, that offset can be changed whenever I remove the nozzle for repair;  but otherwise, it is a long-term fixed value).  On the other hand, I thought that I was preserving the relative nature of the measurement, in that at the end, you still may want to zero to a probe point.

Are you taking issue with my "correction":

   this->pData[pindex] = z + zprobe->getProbeHeight()

?  The probe_height is an arbitrary distance above the bed to start the probe movement, but I see this value ending up added to the Z home offset, and the modified line above removes it.  There is a possibility that the way you are doing calibrations, you are wiping out that Z home offset by re-zeroing Z to the first probe point.  In that case, for your method, any constant value added to all the pData values should be irrelevant, and my code change above won't make any difference to it.  But if you know that the Z probe is fixed relative to the nozzle, then the value that normalize_grid sets the Z home offset to is significant, and correct, and the value of Z after probing is exactly where you want it to be.

I don't think that this interferes with your mechanical/removable probe method, but if it does, then I would ask for an additional parameter to ZGridStrategy to specify which kind of probe is being used (fixed vs. removable;  or inductive vs. mechanical), and modify the routine accordingly.

I did a few prints today with my modified algorithm and it is very cool to see the Z motor tracking the bed whereever it goes.

wolfmanjm:
@RTimothyEdwards sorry I misread your comment, yes that is correct the probe height needs to be subtracted from the probe result to get the relative bed height. I do that in the three point leveling too. https://github.com/Smoothieware/Smoothieware/blob/edge/src/modules/tools/zprobe/ThreePointStrategy.cpp#L289

the probe offset though should not be used in the calculation which is what I thought you were suggesting..

RTimothyEdwards:
Ah, then the subtraction of the probe height is missing from the Z grid calculations.  Should I add that to my commit?  I'm hesitant to add anything that would change the result of the Z grid method, since it claims to have been tested an proven on at least one device.

wolfmanjm:
i think it should be added, but @qharley needs to chime in here

qharley:
It should really be added and if it's not it might be a regression that
crept in with one of the latest rounds of bug fixes.

qharley:
It seems to work very well.  I have a small change in terms of it handling the probe removal position (currently it does nothing) that will also optimise the main loop of the code a tiny bit.

@wolfmanjm  if you pull this I will do a pull request on the additional change.

@RTimothyEdwards Nice work!

qharley:
In terms of the probe height, it is compensated for in line 505.
this->cal[Z_AXIS] = std::get<Z_AXIS>(this->probe_offsets) + zprobe->getProbeHeight();

Z will drive to a height of probe_ofsets + getProbeHeight, but will only record the delta beween that position and the probe tip.  The actual probe height is then effectively mechanically removed from the equation because of the relative measurement.

wolfmanjm:
I think he still needs to add ```this->pData[pindex] = z + zprobe->getProbeHeight();``` then I'll merge.

wolfmanjm:
Ok I didn't see the last comment.

wolfmanjm:
@qharley I never noticed you add in the probe offset.. that makes it pretty much unusable for me. Anyway to remove that? the probe s need to be relative and the probe offset should not be required for Z at all.

qharley:
The probe offset should not influence you if your offset settings are zero for X and Y.  I only added it to ensure that if we mount a probe to the side of the nozzle the machine can still probe the rights points on the bed if you require it.  default is zero offset.

We use it in Morgan to specify the position of the probe switch, 14.5mm in front of the nozzle.  I might want to add a switch in the hot-end mount, which would change the offset to zero again.

wolfmanjm:
my offsets will not always be 0 for X and Y but the Z probe offset is meaningless when using a temporary probe.

qharley:
I also use a temporary probe but set the Z distance to be slightly more
than the offset.  That way you won't crash the bed by mistake while moving
in to do M306 Z0

RTimothyEdwards:
Thanks for the merge.  Sorry I wandered off for a week without responding.

> In terms of the probe height, it is compensated for in line 505.
> this->cal[Z_AXIS] = std::get(this->probe_offsets) + zprobe->getProbeHeight();
>
> Z will drive to a height of probe_offsets + getProbeHeight, but will only record the delta beween that position and the probe tip. The actual probe height is then effectively mechanically removed from the equation because of the relative measurement.

I don't quite get that, and it's not what I see happening in practice.  What I see happening is that I set the probe_height to, say, 5mm, and the line above will always move to a position with the probe about 5mm above the target.  Then "the delta between that position and the probe tip" will be about -5mm;  that is, the probe_height value plus some small error offset which is the thing I want to measure.  So all my recorded values are around -5mm.  Then, the routine calls normalize_grid(), which subtracts off that -5mm value from all the grid points, and then adds the negative of that value (5mm) to the Z home offset to get the new Z home offset.  So my Z home offset, which used to be close to zero, is now about 5mm.

I think that the Z home offset has no effect until "home Z" is called, so that if you don't home the Z axis you might never notice that the Z home offset is now a large positive value.  But in my setup, I am defining "HOMEXYZ"  for the homing at the beginning of the "G32" command.  So if I do "G32" twice in a row, then the second time, it homes Z, then adds a 5mm offset to Z.  At that point, it thinks that it is at Z = 5mm above the bed when in fact it is sitting right on the bed already.  It proceeds to discover that the probe is already tripped at every probe point and records a zero, effectively erasing the Z grid.

So I think that maybe this only shows up for me because (1) I still have my Z min stop in place, and (2) I am using HOMEXYZ for the homing at the beginning of G32.  Changing line 520 from

      this->pData[pindex] = z ;
to
     this->pData[pindex] = z + zprobe->getProbeHeight();

solves the problem for me, and I can't see any reason why it would cause problems in another setup (but please correct me if I'm wrong).  If you agree, I will update and make another pull request for the single change.

I would like to avoid the HOMEXYZ altogether, but I don't think I can do that due to the way my inductive probe sits in front of the plate, hanging over air, when the printer has homed X and Y.  It might work with a more complicated homing option that would (1) home X and Y, (2) move to center, (3) home Z (using the Z probe), (4) home X and Y.  But that's an experiment for another day.

qharley:
Ok. I'll test that on machines this morning. We also need homexyz every
time because of the SCARA implementation.

PR #741, opened 2015-09-23
all

rickenbroc:


wolfmanjm:
what is this? what is it for? this will need a lot of explanation.
Thank you

wolfmanjm:
If this is an STM32 port then it needs to stay in your fork. We can link to it on the wiki in third party branches.

There is no way to merge this into smoothie.

PR #746, opened 2015-09-27
Print actuator position adjustments after homing

OskarLinde:
***Calibration 101 Lesson 1:*** Make sure your endstops are working properly.

I had suspected my endstops where acting-up sporadically, so I implemented the attached patch to investigate. A large number of ```G28```s (with 100 mm of moving back in between each) later, I got the following results:

image

These are histograms showing the positions of each endstop trigger position. The x-axis unit is in steps. 80 steps/mm make each step 12.5 µm. Slow homing rate was 5 mm/sec.

Look at that alpha endstop(!). I replaced all three microswitch based endstops with optical ones:

image

Better – now all three act like the beta-endstop did initially – but there are a number of -75 µm trigger points that look suspicious. Doubling the homing rate to 10 mm/sec:

image

Interestingly, those outlier trigger points move twice as far away – their cause seems to be related to time rather than position. Could there be something that the CPU starts doing at the wrong time making it miss the endstop signal by a few milliseconds?

Reducing the homing rate to 1 mm/sec made that problem mostly go away. All endstops are now repeatable to within ± 1 steps:

image

Anyway, I thought the attached patch, or even better a modified version of it that applies to single-axis homing as well, could help other users make sure their endstops are working as expected.

I also implemented a fix making all my homes end on exactly the same single microstep (>100 trials) by automatically repeating homing. If that sounds interesting, I'll wrap it up with config variables etc...

A few remaining questions:
* What causes the time-related outliers?
* Is it a good idea to lower the default slow homing rate?
* What kind of repeatability are others getting?

### DIFF
--- a/src/modules/tools/endstops/Endstops.cpp
+++ b/src/modules/tools/endstops/Endstops.cpp
@@ -659,6 +659,10 @@ void Endstops::on_gcode_received(void *argument)
 
             if(home_all) {
                 // for deltas this may be important rather than setting each individually
+                float actuators_before[3];
+                for (int i = 0; i < 3; i++) {
+                    actuators_before[i] = THEKERNEL->robot->actuators[i]->get_current_position();
+                }
 
                 // Here's where we would have been if the endstops were perfectly trimmed
                 float ideal_position[3] = {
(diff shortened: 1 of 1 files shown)


wolfmanjm:
what causes the problems are delays in on_idle, (or bad endstops). Basically if you have a host connected and it decides to look at the sdcard it will delay on_idle for a long time. additionally if you have panel and it decides it is time to update the panel it will delay on_idle. and any interrupts that come in may delay on_idle too. Any delays in on_idle will cause the zprobe or endstops to return inconsistent results.
The real solution to this is to have the endstops trigger an interrupt that stops the movement immediately regardless of what else is going on.

OskarLinde:
* You were awfully quick to close this, does that mean that you think it is a bad idea to add that printout on homing? Why?
* What would it entail to implement interrupt routines for the endstops?
* Would it be prudent to change the default homing rate or at least update the documentation?
* If it is hard to implement interrupt routines for the endstops, would an automatically repeated endstop homing make sense (as it solves my endstop repeatability issues perfectly)?

wolfmanjm:
It is debugging info and is not needed, it does not fix anything. It is also not optional and we don't want that printed everytime we home.

Adding interrupts is always very tricky. Most programmers are inexperienced with concurrent programming and how it can have unseen effects. This one is particularly hard to do and maintain the optional module loading. It is also complicated by debounce logic which is required for mechanical switches. However I am closing in on a solution to most of these issues and will add them once I have it all figured out. (Been trying to figure out how to make endstops interrupt driven for quite some while now and maintain smoothie structure).

Homing rates are already default to very slow and can be made slower if needed. The current rate works for most people.
Not having the MSD mounted while homing or probing (and indeed even printing)  is already well documented.

PR #748, opened 2015-10-01
CoreXZ Support

jnjackins:
I've added support for CoreXZ style 3D printers (http://forums.reprap.org/read.php?2,377858)

I''d be happy to contribute the changes upstream. Please have a look and tell me what you think of my approach.

Another approach would be to integrate the CoreXZSolution code with the HBotSolution code with a configuration option such as hbot_motion_plane, letting users choose an XY configuration for HBot or CoreXY style machines, and XZ for CoreXZ style machines. This would reduce duplicate/similar code, but also add complexity.


wolfmanjm:
This look fine except rather than provide a whole new config file, just provide the few lines of config and put it in the Snippets directory.

wolfmanjm:
Also maybe you should add options settable by M665 like the delta solution does to give an alternative for setting the reduction configuration.

jnjackins:
Re: M665 - I'm not sure there's much point. These reduction values aren't something that ever needs calibration or adjustment; they should always be round numbers that are fundamental to the design of the machine. If you feel strongly about it, though, I can add it.

wolfmanjm:
ok no problem, just remove the sample config, and create a snippet instead. 
Thanks

jnjackins:
Already done :)

PR #749, opened 2015-10-02
Removed old reference to max_jerk

justinledwards:


arthurwolf:
Thanks :) Nice catch.

PR #751, opened 2015-10-04
for the change with support for tower scaling.

eeuse1:
Hi

yesterday i upload the newest versio ( with support for tower scaling

then  X axis & Z axis is out of control ，no matter which axis i move，there just Y axis is moving

but i still can do home all axis


wolfmanjm:
This does not merge. I think you commited a pull request to the wrong repo.

626Pilot:
I must have typed the wrong thing. Sorry for the trouble.

PR #759, opened 2015-11-05
Feature/nmotors

trayracing:
*FOR CODE REVIEW* Pull request of the feature/nmotors branch please. (Hopefully it will be ready for merging without too much ado.) This branch adds the abilty for an arm solution to control more than three motors given an X/Y/Z position. It also adds a new arm solution that drives mirroring motors for a cartesian. (I have wilder plans for it, but this was an easy test.)

The code delta is less intimidating than it looks:
* A new type, ActuatorCoordinates, is created and used in place of float[3] for the actuator postions. This let static typing do the work of locating code that needed to be changed.
* Several places that that hard-coded 3 actuators were refactored. (Including Robot::on_config_reload, where steppers are allocated and configured)
* A new arm solution, multicartesian, is introduced. It can be configured to mirror/reorder any of the axes.
These end up with mostly small changes repeated in a fair number of places.

Things I didn't do:
* Change endstops or z-probe code to usefully handle more than 3 actuators. I didn't have a setup to test such changes.
* Augment the M92/114 commands - (setting steps per mm) They use X/Y/Z notation, rather than the A/B/C... notation that would be easier to extend. A little discussion is in order first. (Maybe just accept either notation?)
Neither of those changes need to block merging the code to edge.

Additional notes:
* Arm solutions needing more than 3 motors need to override get_actuator_count().
* ActuatorCoordinates is an std::array<float, 6>. The portion of the array in use is robot->actuators.size(). This is slightly klunky compared to a vector, but keeps a fast, heap-free implementation for coordinates.
* Anyone needing more than 6 actuators will need to bump up the size of ActuatorCoordinates, add code to configure the additional steppers, and recompile.


triffid:
If we're going to finally get this change in, consider extending arm
solution input to accept both position and rotation so we can control stuff
like robot arms and Stewart platforms that can rotate as well as translate

trayracing:
triffid, that sounds like a great followup project. Are you thinking we should create a new type AxisCoordinates to pass into the arm solution now, so that the arm solution function signatures only changes once? (We can add an get_axis_count() method later, when the other supporting work such as motion planning is done, without forcing further source level changes.)

trayracing:
Another point for discussion - the multicartesian solution could replace the regular cartesian solution, if we change the multicartesian default to drive only 3 motors. Yes? No?

wolfmanjm:
This is great work.

My main concern is it uses a lot of extra memory even if you only have 3 axis as 99% of users do. We are extremely short on memory and using this extra memory when it is not required is simply not a good idea..

My next concern is the increase in processing time in the step ticker ISR. Have you hooked this to a logic analyzer and timed the increase it processing time? especially when nearing 100KHz steps. also the block handover time is currently too long anyway and this will almost double that I think.

Lastly testing is an issue, this affects every arm solution so who is going to test every arm solution and make sure it isn't breaking something?

Other than that it is one of the better attempts to add multi axis I have seen so far.

My suggestion is fix it so that 3 axis does not use any more memory than it currently does, and the extra axis only use memory if they are configured..

Get agreement from all the major arm solution maintainers to test this (they really may not want to as it does not really benefit them at all).

Maybe consider maintaining this as a separate fork for people who need more than 3 axis.

trayracing:
Hi wolfmanjm. Perhaps you misread the description. This doesn't add extra axes, just extra actuators - but ONLY when you ask for them. As such, it shouldn't affect the step ticker in the 3 actuator case at all. I don't, however, own a logic analyzer to confirm that. 

You mention a lot of extra memory would be used (presumably in the 3 actuator case). Where is that? On the heap? I don't see a actuator coordinate sets being stored - they are generally stack objects used in converting from axis coordinates to feed the steppers. Is there a good way to test peak memory usage? It's easy enough to compile with k_max_actuators set to 3 instead of 6.

I would be happy to contact your list of major arm solution maintainers, if you would provide it. I expect many will immediately recognize this is simply a type change, with no changes to their logic or algorithms.

On a larger scale, this change is not what I would call localized, so I would like to see some regression testing. My project is currently a stack of 5 motors, which limits what I can test.

A separate fork would be a poor choice - it fragments the user base, and it's too easy for 3 actuator assumptions to creep in. A simpler route would be merge with k_max_actuators set to 3 instead of 6, if experiments prove it's needed.

wolfmanjm:
use the mem command, load with edge check mem, then load with your branch and 3 axis and select mem, the difference in free memory is what you have used extra.

I see many cases in your PR that allocate a larger array than the three actuators currently allocated, therefore using more memory.

Just to confirm you expect the number of actuators to be selected at compile time not at config time correct? in which case the default needs to be set to 3 not 6 as it currently is.

src/modules/robot/Robot.cpp
```diff
@@ -396,22 +396,20 @@ void Robot::on_gcode_received(void *argument)
...
                     this->max_speeds[Z_AXIS] = gcode->get_value('Z');
-                if (gcode->has_letter('A'))
-                    alpha_stepper_motor->set_max_rate(gcode->get_value('A'));
-                if (gcode->has_letter('B'))
-                    beta_stepper_motor->set_max_rate(gcode->get_value('B'));
-                if (gcode->has_letter('C'))
-                    gamma_stepper_motor->set_max_rate(gcode->get_value('C'));
-
+                for (size_t i = 0; i < actuators.size(); i++) {
+                    if (gcode->has_letter('A' + i))
```
wolfmanjm: this is a problem you cannot have actuators named E or F or G, so you cannot simply do 'A' + i for more than four actuators

src/modules/robot/Robot.cpp
```diff
@@ -194,58 +208,54 @@ void Robot::on_config_reload(void *argument)
...
+        uint16_t per_mm_checksum;
+        uint16_t per_sec_checksum;
+        char const *pin_defaults[3];
+        float per_mm_default;
+    } p[] =
+    {
+        { ACTUATOR_CHECKSUMS("alpha"),   {"2.0", "0.5",  "0.4" },   80.0F },
+        { ACTUATOR_CHECKSUMS("beta"),    {"2.1", "0.11", "0.10"},   80.0F },
+        { ACTUATOR_CHECKSUMS("gamma"),   {"2.2", "0.20", "0.19"}, 2560.0F },
+        { ACTUATOR_CHECKSUMS("delta"),   { "nc",   "nc",   "nc"},   80.0F },
```
wolfmanjm: this is a problem as delta and epsilon are used for setting extruder currents so will lead to confusion if this is used in addition to two extruders.

src/modules/tools/endstops/Endstops.cpp
```diff
@@ -669,15 +670,14 @@ void Endstops::on_gcode_received(void *argument)
...
 
                     // We are actually not at the ideal position, but a trim away
-                    float real_actuator_position[3] = {
-                        ideal_actuator_position[X_AXIS] - this->trim_mm[X_AXIS],
-                        ideal_actuator_position[Y_AXIS] - this->trim_mm[Y_AXIS],
-                        ideal_actuator_position[Z_AXIS] - this->trim_mm[Z_AXIS]
-                    };
+                    ActuatorCoordinates real_actuator_position;
+                    for (size_t i = THEKERNEL->robot->actuators.size() + 1; i--;) {
+                        real_actuator_position[i] = ideal_actuator_position[i] - this->trim_mm[i];
```
wolfmanjm: trim_mm is hardcoded as 3 this will crash if there are more than three actuators.

wolfmanjm:
Ok so I did misunderstand a bit, this only adds more actuators and not any more axis... so my question is how is this useful in the standard use case? what are the use cases for this? why is this useful for a wider audience?

src/modules/robot/Robot.cpp
arthurwolf: Well for current, there is a direct mapping between greek letter and on-board motor, so it doesn't matter if we are setting current for an actuator or an extruder.

This *does* make things a bit harder to understand for beginners, but that's a job for the documentation, not for the code.

arthurwolf:
@wolfmanjm This is useful for any arm solution that requires more than 3 actuators, like this one https://www.youtube.com/watch?v=lNC9tqnVhwg , robot arms, and a few others people have been asking about/trying to implement.
All in all, it's also a great step towards 6-axis control, and moving extruder movement into robot.

If this doesn't have too big of a cost in terms of ram and flash, and the code can be made clean enough, I really really want this to get in.

src/modules/robot/Robot.cpp
trayracing: Good point. Actually, adding rotational position like triffid wants also uses A, B and C, so the existing three are begging for change too. Unfortunately, there isn't a nice long string contiguous string of available letters. Anyone have favorite triples that have easy-to-rember greek letter nmes? RGB and STP (from computer graphics) are already spoken for.

arthurwolf: It makes little sense keeping up with this format as we add more actuators/axes.
I'd much rather we didn't modify this gcode to accomodate more axes/actuators, and made a *new* gcode that doesn't use letters, but rather answers something like 

axes:0,10,0,0,5,0 actuators:10,30,10,5,0,5

trayracing:
@wolfmanjm, the number of actuators allocated is controlled at run time by the arm solution's get_actuator_count() method. The default implementation returns 3. Overriding this allows multicartesian to allocate only the motors requested by the config file. BTW, multicartesian is a functional superset of cartesian. In the interest of space, should I replace the guts of cartesian with multicartesian?

src/modules/robot/Robot.cpp
trayracing: Is this table worth splitting up to shink it? Around 52 bytes savings, less whatever the additional code takes up.

src/modules/tools/endstops/Endstops.cpp
trayracing: Thanks for the catch. Fixed committed. I didn't try to extend trim to the extra actuators - I don't have a setup to test that.

src/modules/robot/Robot.cpp
trayracing: That's good for readback, what should we do for entry? Specify an input/output : index and value? Like M203 I0 25 I2 15 O4 97? (OK, poor example. :) ) If this is a larger question, I'm OK backing out the g-code mods from this pull request and doing them separately as a follow-on once there's an agreed-on format. Also, are there guidelines somewhere for pinking new M-codes?

trayracing: Another option for setting many steppers or positions would be letter pair codes. Like M203 PA12 PB33 SA400 RD1500, where the first letter of the pair would be "P"  for position, "S" actuator, or "R" for raw step count. The second letter would be the channel: "A","B","C",... We could make the second character a digit but that make it hard to read. Perhaps this wants to be a separate pull request.

arthurwolf: You should leave getting/setting values of actuators via gcodes alone for
now, leave it out of this pull request, and we'll work on it later.
There is already a lot to do on the more basic stuff.

On Thu, Nov 5, 2015 at 11:56 PM, trayracing <notifications@github.com>
wrote:

> In src/modules/robot/Robot.cpp
> <https://github.com/Smoothieware/Smoothieware/pull/759#discussion_r44082349>
> :
>
> > @@ -396,22 +396,20 @@ void Robot::on_gcode_received(void *argument)
> >                      this->max_speeds[Y_AXIS] = gcode->get_value('Y');
> >                  if (gcode->has_letter('Z'))
> >                      this->max_speeds[Z_AXIS] = gcode->get_value('Z');
> > -                if (gcode->has_letter('A'))
> > -                    alpha_stepper_motor->set_max_rate(gcode->get_value('A'));
> > -                if (gcode->has_letter('B'))
> > -                    beta_stepper_motor->set_max_rate(gcode->get_value('B'));
> > -                if (gcode->has_letter('C'))
> > -                    gamma_stepper_motor->set_max_rate(gcode->get_value('C'));
> > -
> > +                for (size_t i = 0; i < actuators.size(); i++) {
> > +                    if (gcode->has_letter('A' + i))
>
> Another option for setting many steppers or positions would be letter pair
> codes. Like M203 PA12 PB33 SA400 RD1500, where the first letter of the pair
> would be "P" for position, "S" actuator, or "R" for raw step count. The
> second letter would be the channel: "A","B","C",... We could make the
> second character a digit but that make it hard to read. Perhaps this wants
> to be a separate pull request.
>
> —
> <https://github.com/Smoothieware/Smoothieware/pull/759/files#r44082349>.
>



-- 
Courage et bonne humeur.

wolfmanjm:
So this needs to be a compile time option only, and set to 3 by default.

Changing the definition of gcode is not a good idea IMO. so M203 PA12 is not only illegal gcode but a really bad idea.

src/modules/robot/Block.h
```diff
@@ -41,14 +42,14 @@ class Block {
 
         std::vector<Gcode> gcodes;
 
-        unsigned int   steps[3];           // Number of steps for each axis for this block
+        std::array<int, k_max_actuators>   steps; // Number of steps for each axis for this block
```
wolfmanjm: This array will be 12 bytes bigger for every block in the queue, that is 384 bytes. This is why the default needs to be 3 not 6, unless someone actually needs 6 motors.

src/modules/tools/endstops/Endstops.cpp
```diff
@@ -146,6 +147,10 @@ void Endstops::on_module_loaded()
 // Get config
 void Endstops::on_config_reload(void *argument)
 {
+    if (THEKERNEL->robot->actuators.size() > 3) {
+        THEKERNEL->streams->printf("Please implement trim for >3 actuators");
```
wolfmanjm: Note this only gets printed out to the uart which no one will see unless debugging. This should do something more drastic. but really until there is a case for endstops for more than three motors, nothing should be changed here, and remain hard coded for 3.

src/modules/tools/endstops/Endstops.cpp
```diff
@@ -669,15 +674,14 @@ void Endstops::on_gcode_received(void *argument)
...
                     THEKERNEL->robot->arm_solution->cartesian_to_actuator(ideal_position, ideal_actuator_position);
 
                     // We are actually not at the ideal position, but a trim away
-                    float real_actuator_position[3] = {
-                        ideal_actuator_position[X_AXIS] - this->trim_mm[X_AXIS],
-                        ideal_actuator_position[Y_AXIS] - this->trim_mm[Y_AXIS],
-                        ideal_actuator_position[Z_AXIS] - this->trim_mm[Z_AXIS]
-                    };
+                    ActuatorCoordinates real_actuator_position;
+                    for (size_t i = THEKERNEL->robot->actuators.size() + 1; i--;) {
```
wolfmanjm: this should be remain hard coded to 3, no changes until it is understood how to handle more than 3 motors for endstops. FWIW this is mostly used for deltas which would haeve no reason to have more than 3 motors.

trayracing:
How does it follow that it needs to be a compile time option set to 3?

trayracing:
Wait - missed your earlier comment. I suppose >3 isn't the common case (is that true arthur - what's the demand for slave axes?), and it's easy enough to build for more.

src/modules/tools/endstops/Endstops.cpp
trayracing: What would be more drastic but safe with unmodified endstop code?

wolfmanjm:
Not only is this an uncommon case, it is a far edge case. Lacking any concrete use cases I see not much use in this at the moment. slaving is done with wiring and takes no memory or flash or time. The one case is when there is an arm solution that slaves an actuator without a one to one relationship (not that I have ever seen that case). If the default is set to 3, the sample arm solution is removed from this PR, and the endstops are left unchanged from what they were. Then I see no reason to not merge this as it does clean up a somewhat messy current implementation,, and actually saves some memory by eliminating the redundant StepperMotor pointers. It also paves the way for the more interesting use case of n-axis.

src/modules/tools/endstops/Endstops.cpp
wolfmanjm: Right now I'd leave endstops unchanged from what it was (other than using the new typedef). Handling multiple endstops can be left for when that use case actually evolves.
left as it was it would continue to work regardless of how many motors were defined.

trayracing:
Slaving isn't that weird - it was a suggested project for smoothiecontest. However, I don't need it and can pull it and set max actuators to 3. I'm fine pulling the current endstop changes and instead disable endstop trim when > 3 actuators to avoid buffer overwrite on real_position. I'll shoot for getting updates in at the end of the weekend.

arthurwolf:
On Fri, Nov 6, 2015 at 1:42 PM, trayracing <notifications@github.com> wrote:

> Slaving isn't that weird - it was a suggested project for smoothiecontest.
>

It's since been abandonned due to it's high RAM cost, when the same thing
can be achieved by simply wiring things in parralel.

> However, I don't need it and can pull it and set max actuators to 3. I'm
> fine pulling the current endstop changes and instead disable endstop trim
> when > 3 actuators to avoid buffer overwrite on real_position. I'll shoot
> for getting updates in at the end of the weekend.
>

This is great. While as-is it doesn't add any new functionality, it's
something we want as it makes it possible to add the more exotic arm
solutions.

> —
> <https://github.com/Smoothieware/Smoothieware/pull/759#issuecomment-154399206>
> .
>



-- 
Courage et bonne humeur.

trayracing:
@wolfmanjm, @arthurwolf Where does documentation about using more than 3 actuators go? I'd been using multicartesian for part of that. I can write a few lines of text, if you want to stick something on the web site.

arthurwolf:
We probably don't need documentation at this point, but we'll be pointing
people at this when they want to add exotic arm solutions.

On Fri, Nov 6, 2015 at 5:32 PM, trayracing <notifications@github.com> wrote:

> @wolfmanjm <https://github.com/wolfmanjm>, @arthurwolf
> <https://github.com/arthurwolf> Where does documentation about using more
> than 3 actuators go? I'd been using multicartesian for part of that. I can
> write a few lines of text, if you want to stick something on the web site.
>
> —
> <https://github.com/Smoothieware/Smoothieware/pull/759#issuecomment-154461385>
> .
>



-- 
Courage et bonne humeur.

trayracing:
OK, I've removed the extra actuators from the Mcodes, removed the endstop trim changes, removed multicartesian, made the zeta default steps/mm more consistent, and hardcoded the max actuator count to 3. Did I miss any requested changes? 

Memory-wise, it looks like total free RAM is very slightly up, and firmware size a bit down.
Here's the stats, old ones first:
---old-------------------------------
Unused Heap: 9108 bytes
Used Heap Size: 17220
Allocated: 8980, Free: 7200
Total Free RAM: 16308 bytes
Free AHB0: 14256, AHB1: 10440

Extracting disassembly to ../LPC1768/main.disasm
   text    data     bss     dec     hex filename
 344748     408    9872  355028   56ad4 ../LPC1768/main.elf
---new-------------------------------
Unused Heap: 8844 bytes
Used Heap Size: 17484
Allocated: 8964, Free: 7504
Total Free RAM: 16348 bytes
Free AHB0: 14256, AHB1: 10440

Extracting disassembly to ../LPC1768/main.disasm
   text    data     bss     dec     hex filename
 343340     408    9872  353620   56554 ../LPC1768/main.elf

wolfmanjm:
great job, thanks

wolfmanjm:
someone want to test this before it is merged? it looks ok, but I may have missed something.

src/modules/robot/Robot.cpp
```diff
@@ -194,58 +205,53 @@ void Robot::on_config_reload(void *argument)
...
-    gamma_en_pin.from_string(   THEKERNEL->config->value(gamma_en_pin_checksum   )->by_default("0.19" )->as_string())->as_output();
-
-    float steps_per_mm[3] = {
-        THEKERNEL->config->value(alpha_steps_per_mm_checksum)->by_default(  80.0F)->as_number(),
-        THEKERNEL->config->value(beta_steps_per_mm_checksum )->by_default(  80.0F)->as_number(),
-        THEKERNEL->config->value(gamma_steps_per_mm_checksum)->by_default(2560.0F)->as_number(),
-    };
-
     // TODO: delete or detect old steppermotors
+#if 0 //TODO: Add unregistering to StepperMotor/StepTicker so we can delete old motors
```
wolfmanjm: please remove this. configs cannot be reloaded (despite the archaic name)

src/modules/robot/Robot.cpp
```diff
@@ -194,58 +205,53 @@ void Robot::on_config_reload(void *argument)
...
-    this->gamma_stepper_motor  = new StepperMotor(gamma_step_pin, gamma_dir_pin, gamma_en_pin);
+    struct
+    {
+        uint16_t pin_checksums[3];
+        uint16_t per_mm_checksum;
+        uint16_t per_sec_checksum;
+        char const *pin_defaults[3];
+    } const p[] =
+    {
+        { ACTUATOR_CHECKSUMS("alpha"),   {"2.0", "0.5",  "0.4" } },
```
wolfmanjm: I'm not too fond of this construct, however I can live with it. But I suggest removing the default pins, and just set them to "nc" if undefined. It should be ok as every config file has always had them specifically defined. This will save some flash and make this construct a bit more tolerable. - Thanks

src/modules/robot/Robot.cpp
```diff
@@ -194,58 +205,53 @@ void Robot::on_config_reload(void *argument)
...
+        { ACTUATOR_CHECKSUMS("delta"),   { "nc",   "nc",   "nc"} },
+        { ACTUATOR_CHECKSUMS("epsilon"), { "nc",   "nc",   "nc"} },
+        { ACTUATOR_CHECKSUMS("zeta"),    { "nc",   "nc",   "nc"} }
+    };
 
-    alpha_stepper_motor->change_steps_per_mm(steps_per_mm[0]);
-    beta_stepper_motor->change_steps_per_mm(steps_per_mm[1]);
-    gamma_stepper_motor->change_steps_per_mm(steps_per_mm[2]);
+    if (this->arm_solution->get_actuator_count() > k_max_actuators) {
+        THEKERNEL->streams->printf("Too many actuators, %d. Increase k_max_actuators and rebuild", this->arm_solution->get_actuator_count());
```
wolfmanjm: This is really not good enough. the message will not be seen unless a uart ftdi is hooked up, and will leave the configuration in an undefined state. I'll need to think about what to do here. but for now just iterate over the smaller of the two values.

src/modules/robot/Robot.cpp
trayracing: Without the defaults, I was able to flatten the table. Also compiled out the extra rows when k_max_actuators == 3.Still not ideal. I would suggest smoothie2 use the dot style configuation for actuators, which would make this table unnecessary and the firmware more compact.

trayracing:
All suggested changes have been made. Is there further action needed on my part?

wolfmanjm:
It needs to be tested by several people with the different types of printers to make sure nothing broke.

arthurwolf:
@trayracing I recommend you make an email to the dev and support smoothie mailing lists, asking for people to help you test this. Provide them with a pre-compiled .bin file.

trayracing:
Is there a normal staging place on for the .bin file?

arthurwolf:
Staging place ?

On Thu, Nov 19, 2015 at 2:35 PM, trayracing <notifications@github.com>
wrote:

> Is there a normal staging place on for the .bin file?
>
> —
> <https://github.com/Smoothieware/Smoothieware/pull/759#issuecomment-158058291>
> .
>



-- 
Courage et bonne humeur.

trayracing:
Where can I put a .bin file so that people can try it?

Sent from Yahoo Mail on Android 
 
 On Thu, Nov 19, 2015 at 5:40 AM, Arthur Wolf<notifications@github.com> wrote:  Staging place ?

On Thu, Nov 19, 2015 at 2:35 PM, trayracing <notifications@github.com>
wrote:

> Is there a normal staging place on for the .bin file?
>
> —
> <https://github.com/Smoothieware/Smoothieware/pull/759#issuecomment-158058291>
> .
>



-- 
Courage et bonne humeur.

arthurwolf:
You can put it in your repository.

On Fri, Nov 20, 2015 at 1:27 AM, trayracing <notifications@github.com>
wrote:

> Where can I put a .bin file so that people can try it?
>
> Sent from Yahoo Mail on Android
>
> On Thu, Nov 19, 2015 at 5:40 AM, Arthur Wolf<notifications@github.com>
> wrote: Staging place ?
>
> On Thu, Nov 19, 2015 at 2:35 PM, trayracing <notifications@github.com>
> wrote:
>
> > Is there a normal staging place on for the .bin file?
> >
> > —
> > <
> https://github.com/Smoothieware/Smoothieware/pull/759#issuecomment-158058291
> >
> > .
> >
>
>
>
> --
> Courage et bonne humeur.
>
> —
> Reply to this email directly or view it on GitHub.
>
> —
> <https://github.com/Smoothieware/Smoothieware/pull/759#issuecomment-158242495>
> .
>



-- 
Courage et bonne humeur.

wolfmanjm:
FYI I'll be incorporating this PR into my WCS work branch, so it'll get merged with that once it is fully tested.

wolfmanjm:
merged into PR #804

src/modules/robot/Block.h
wolfmanjm: this needs to be unsigned, fixed in my branch

PR #761, opened 2015-11-05
Added an panel.init property to the ST7565 module to specify a custom display init sequence

errolt:
Hi Arthur,

I don't know if this will be useful to others but I have added an option to the ST7565 module so that the user can specify the initial display init bytes. I have a display that has a different row/column config and is upside down and reversed with the default init code.

Sample:
panel.lcd                                    st7565_glcd
panel.init                                    e2,a6,a4,a0,ae,c0,a2,2f,21,25,81,2f,af

If there are any violated standards then please tell me.

Thank you,
Errol

### DIFF
--- a/src/modules/utils/panel/panels/ST7565.cpp
+++ b/src/modules/utils/panel/panels/ST7565.cpp
@@ -42,6 +42,7 @@
 #define a0_pin_checksum            CHECKSUM("a0_pin")
 #define red_led_checksum           CHECKSUM("red_led_pin")
 #define blue_led_checksum          CHECKSUM("blue_led_pin")
+#define init_checksum          	   CHECKSUM("init")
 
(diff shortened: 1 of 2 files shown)


wolfmanjm:
this adds 52 bytes of RAM use whether it is used or not. That is not acceptable. If you can make it use no extra RAM when not used I will reconsider.
Thank you

errolt:
Would it have less impact if I created a new panel module, duplicating the current ST7565 module, that has that option added? I'm not sure how unused classes affect memory usage, as I normally develop in plain C.

wolfmanjm:
I'd do whatever you feel is easiest, and keep it in your fork. This is not something I see as generally useful therefore not really needed in the main branch.
If I were you I'd just #ifdef around the setup in the current one and set it to whatever you want, that way it uses no extra memory.

PR #762, opened 2015-11-09
Reset Watchdog introduced to capture especially heating problems afte…

risip:
…r firmware crash.

This is especially important for firmware crashes in the heating environment.
Since after a firmware crash there may be no control over the heating mechanism anymore!
The feature itself is used in our 3d printer, but with some code modifications, because of other dependencies.

### DIFF
--- a/src/modules/utils/resetWatchdog/ResetWatchdog.cpp
+++ b/src/modules/utils/resetWatchdog/ResetWatchdog.cpp
@@ -1,0 +1,40 @@
+#include "ResetWatchdog.h"
+#include "Kernel.h"
+#include "SlowTicker.h"
+#include "StreamOutputPool.h"
+#include "SerialMessage.h"
+
+ResetWatchdog::ResetWatchdog(){
+   if(WDT_ReadTimeOutFlag() == SET){
+      awaked_from_reset = true;
+      WDT_ClrTimeOutFlag();
+      connected = false;
+   }
+}
+
+void ResetWatchdog::on_module_loaded(){
+   WDT_Init(WDT_CLKSRC_PCLK,WDT_MODE_RESET);
+   WDT_Start(100000);   // in micrseconds
+   THEKERNEL->slow_ticker->attach( (unsigned int)11, this, &ResetWatchdog::watchdog_tick ); // readings per second must be appropriate to restart the watchdog before timeout
+   this->register_for_event(ON_GCODE_RECEIVED);
+}
+
+uint32_t ResetWatchdog::watchdog_tick(uint32_t dummy){
+   WDT_Feed();
+   return 0;
+}
+
+// G-Code can be only received if connected.
+// Needed to ensure that the user sees the message
+void ResetWatchdog::on_gcode_received( void *argument ) {
+   //Gcode *gcode = static_cast<Gcode *>(argument);
+
+   if(!connected){
+      connected = true;
+      if(awaked_from_reset){
+         THEKERNEL->streams->printf("// Booted after internal Reset trigger, because of a firmware crash.\n");
+         awaked_from_reset = false;
+      }
+   }
+}
+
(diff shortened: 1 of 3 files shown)


src/modules/utils/resetWatchdog/ResetWatchdog.cpp
```diff
@@ -0,0 +1,40 @@
+#include "ResetWatchdog.h"
+#include "Kernel.h"
+#include "SlowTicker.h"
+#include "StreamOutputPool.h"
+#include "SerialMessage.h"
+
+ResetWatchdog::ResetWatchdog(){
+   if(WDT_ReadTimeOutFlag() == SET){
+      awaked_from_reset = true;
```
arthurwolf: I believe it's "awaken".

arthurwolf:
@risip I think the watchdog is already being used in something related to DFU flashing and the bootloader.

Did you look into this, and is your code compatible with that function ?

wolfmanjm:
This is not compatible and breaks the DFU. the bootloader would also need to be changed. I have already actually done this, and a module is not required to do it. If you look more closely you will see there is already a watchdog library implemented.

wolfmanjm:
FWIW this is also totally the wrong way to handle watchdog. feeding the watchdog from an interrupt is really the wrong thing to do, as a stall in on_idle which would affect heaters will not trigger it as the interrupts will continue.
Also when the firmware crashes MRI captures the crash and turns off all mosfets already, so that is already done.
Also the flag you set to indicate a watch dog reset (which probably will never happen) will not work as it does not survive a reset.

Basically this method is incorrect in so many ways.

arthurwolf:
Hey !

Thanks a lot for offering to help with this.

As @wolfmanjm mentionned this is incompatible with existing stuff unfortunately.

I wonder if there is a solution that'd allow us not to break DFU, but to add watchdog-triggered-MRI, without having to change the bootloader ( which we can't ) ... not sure there is though.

triffid:
There's no reason we couldn't reflash the bootloader with a special
firmware update or from within smoothie itself, I don't believe there's
flash area protection like AVRs. The big caveat is that they strongly
suggest disabling interrupts while flashing, so handling it from within
smoothie may be problematic.

As for bootloader using the watchdog flag, considering how rare it is for
smoothie to lock up solid without crashing, is it really such a problem to
need a manual reset afterwards because it's in DFU mode?

If someone is printing, the print is likely lost anyway and those of us
interested in debugging would have this feature disabled so we can press
ISP and go digging.

Ultimately it's simply a safety feature to disable heaters in the unlikely
event of a non-crashing lockup - we already disable heaters on crash.

wolfmanjm:
I already added watchdog handling, as @triffid already did 99% of the work in the libs/Watchdog/cpp file.
https://github.com/Smoothieware/Smoothieware/pull/763
I just needed to test it before merging.

wolfmanjm:
Thank you for attempting to add this, I am sorry I was harsh, but your PR would never have worked, and testing is required for any future PRs you may wish to submit.
Anyway it nudged me to merge the code I had already added, so Watchdog is added now and this has been tested.
In future please make sure you totally understand the system before hacking on it, and if something is not clear please ask first.
In this case you were unaware that the watchdog flag was used in the bootloader to force DFU upload mode, and that interrupts continue to fire even if the system has hung somewhere, and that on a crash all mosfets are already turned off.

Thank you for helping..

risip:
Hello all,


this code is tested and used in our environment.

I didn't see the DFU interaction, since we do not use it currently. Thank you for the hint I have to take this into account!

But it was necessary to have this, because our heaters were not disabled when the firmware crashes.


As I can see within DFU.cpp it is only used for detaching DFU, so I have to take care to disable the other watchdog at this point!


Best regards

wolfmanjm:
actually your version will not work at all. I'd discard it and use what we just pushed to edge. That does work. The DFU is not in the smoothie code it is in the bootloader which you cannot change.

PR #764, opened 2015-11-10
SCARA solution updates

qharley:
Improved calibration movements: configurable Z movement
Included D and E parameters to get and set optional: set min and max
singularity protection values on the fly and save to config-override

Calibration example added in snippets.


src/config.default.new
```diff
@@ -0,0 +1,9 @@
+# This is a place for the user to specify their own 'firm' defaults at compile time
```
wolfmanjm: what is this? please remove it.

qharley: will do.

qharley: Done.  Don't know how that crept in...

PR #768, opened 2015-11-13
Remove play/pause

wolfmanjm:
rename pause to kill
use kill functionality if old pause button enabled
refactor to rename pause to kill


qharley:
How should I proceed if I want to keep the pause button on my machine?

arthurwolf:
You probably want to setup a switch module to trigger the suspend command.
See http://smoothieware.org/stopping-smoothie and
http://smoothieware.org/switch

On Fri, Nov 13, 2015 at 12:04 PM, Quentin Harley <notifications@github.com>
wrote:

> How should I proceed if I want to keep the pause button on my machine?
>
>
> On Fri, Nov 13, 2015 at 8:27 AM Jim Morris <notifications@github.com>
> wrote:
>
> > Merged #768 <https://github.com/Smoothieware/Smoothieware/pull/768>.
> >
> > —
> > <https://github.com/Smoothieware/Smoothieware/pull/768#event-463287076>.
>
> >
>
> —
> <https://github.com/Smoothieware/Smoothieware/pull/768#issuecomment-156398589>
> .
>



-- 
Courage et bonne humeur.

wolfmanjm:
You just leave everything as it is, pause is actually kill. Pause never worked reliably so it was replaced by kill, it pretty much does the same thing except you cannot resume, which you could't do before anyway.
If you actually want a pause/resume then as arthur says you can use suspend (M600) and resume (M601), or simply pause the stream on the host with the pause button on the host.

qharley:
With my Morgans I actually could pause and resume reliably, but i agree
that suspend and resume is a better choice.

I am just looking for a way to keep using a very handy button I added to
the machine that will do pause/resume: Users found very creative ways to
make use of it to improve their workflow.
Kill is a bit of a waste to me since the very handy reset button does
exactly that...

MrRocketman:
Hello,

I just updated my CNC mill and am trying to get the kill functional working. The problem is that my machine has a belt driven z axis, so when I kill the machine, the spindle drops like a rock and crashes into the workpiece. How can I utilize the new kill functionality yet maintain power to the stepper motor to prevent a crash? 

Thanks,
James

wolfmanjm:
that is simply not an option. kill should really kill the power (and can be configured to do so), but the soft kill turns everything off that it can. leaving power on the motors would not really be a kill function.
I cannot think of a solution for your use case I am afraid other than fixing your Z to have a little more friction.

wolfmanjm:
Having said that one solution would be to disable the enable pin for that motor and force it low, that way it will always have power regardless of the halt state. However I would not recommend that solution unless you have no other option.

WZ9V:
I can see a case for both concepts.  An emergency stop which is kill all the power because something has gone horribly wrong and a kill this job which basicly stops the tool and moves the machine to a predetermined safe position.  On my 3D printer I'd use the second one for the oh crap my print is messed up for some reason and I just want to stop the job and clear the mess.

wolfmanjm:
you can already do the second case, use pause or stop in pronterface, or abort if printing from sdcard

arthurwolf:
you can also control that enable pin with a gcode using a switch module

On Tue, Nov 24, 2015 at 1:41 AM, Jim Morris <notifications@github.com>
wrote:

> Having said that one solution would be to disable the enable pin for that
> motor and force it low, that way it will always have power regardless of
> the halt state. However I would not recommend that solution unless you have
> no other option.
>
> —
> <https://github.com/Smoothieware/Smoothieware/pull/768#issuecomment-159115061>
> .
>



-- 
Courage et bonne humeur.

MrRocketman:
Good idea with the switch module. I tried adding this:

switch.zPermanent.enable                     	true
switch.zPermanent.output_pin			0.21 // Z axis enable pin
switch.zPermanent.fail_safe_set_to		1

But it still temporarily disables the z for a second after hitting the kill button. Interestingly the z stays enabled if I hit the kill button while the machine is not moving. Only when the machine is moving and I hit kill does it disable for a second causing it to drop. It doesn't quite seem like fail_safe_set_to is doing it's job. Or I don't understand it all the way. Any more hints? Thanks

arthurwolf:
You also want to set the enable pin to nc so smoothie doesn't disable it
when it disables the others.

On Tue, Nov 24, 2015 at 10:37 PM, James Adams <notifications@github.com>
wrote:

> Good idea with the switch module. I tried adding this:
>
> switch.zPermanent.enable true
> switch.zPermanent.output_pin 0.21 // Z axis enable pin
> switch.zPermanent.fail_safe_set_to 1
>
> But it still temporarily disables the z for a second after hitting the
> kill button. Interestingly the z stays enabled if I hit the kill button
> while the machine is not moving. Only when the machine is moving and I hit
> kill does it disable for a second causing it to drop. It doesn't quite seem
> like fail_safe_set_to is doing it's job. Or I don't understand it all the
> way. Any more hints? Thanks
>
> —
> <https://github.com/Smoothieware/Smoothieware/pull/768#issuecomment-159412432>
> .
>



-- 
Courage et bonne humeur.

wolfmanjm:
You want
```
switch.zPermanent.fail_safe_set_to 0
```

otherwise it will get set to 1 on halt or crash. which is exactly what enable does.

PR #770, opened 2015-11-14
PlayLed feature

qharley:
added back slow flicker function to indicate system suspend status

### DIFF
--- a/src/modules/utils/PlayLed/PlayLed.cpp
+++ b/src/modules/utils/PlayLed/PlayLed.cpp
@@ -3,12 +3,15 @@
 /*
  * LED indicator:
  * off   = not paused, nothing to do
+ * slow flash = suspended
  * fast flash = halted
  * on    = a block is being executed
  */
 
 #include "KillButton.h"
 #include "modules/robot/Conveyor.h"
+#include "PublicData.h"
+#include "modules/utils/player/PlayerPublicAccess.h"
 #include "SlowTicker.h"
 #include "Config.h"
 #include "checksumm.h"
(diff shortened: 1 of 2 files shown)


src/modules/utils/PlayLed/PlayLed.cpp
```diff
@@ -54,8 +57,25 @@ uint32_t PlayLed::led_tick(uint32_t)
 
     if(++cnt >= 6) { // 6 ticks ~ 500ms
         cnt= 0;
-        led.set(!THEKERNEL->conveyor->is_queue_empty());
```
qharley: will add back...

wolfmanjm:
Sorry but suspend is not a system state like pause was so this is not appropriate IMO.

arthurwolf:
@wolfmanjm maybe we could have an optional output pin that would turn on when suspend is issued and off when resume is issued ?

wolfmanjm:
I'd reconsider if this was optional, and defaults to off. and if it did not use public request to get the suspend status. basically that means moving Player to Kernel, and querying it directly.
I still do not consider suspend a system status though, just as pause from the host (or for sdcard) is not a system status. and playled is supposed to indicate system status IMO.

qharley:
Interesting...

No problem.  At least the Morgans will keep behaving the same way even
though the pause function has been removed.

I used code borrowed from the panel module (public request).  Thought it
would be acceptable since it is used already...

wolfmanjm:
Well pause was removed becuase it was totally broken. should never have been used. We had been talking about that for a while, it was actually deprecated over a year ago.

As for suspend, it is a totally different thing in my view. Halt (and pause) were system states and nothing could be done until they were out of that state. suspend is not a system state just as pausing a print from pronterface or stopping a print. there is no change in the system state and the print can be restarted or not and it can be resumed or not, the system itself is not stopped, just the current print is suspended.

qharley:
Ok.  I suppose it has a lot to do with perception. Even though the state in
the software did not change, the machine is waiting for the user, and I
want to indicate that it is.

No problem.  I will maintain it in my own branch.

PR #772, opened 2015-11-18
Pay attention to G2/G3 in extruder module. Fix #771.

### LINKED ISSUE #771: No extrusion in G2/G3
It seems that "E" parameter to G2/G3 (arc moves) is ignored - no extrusion happens during move.
For example:
G92 E0
G2 X-100 Y-100 I20 J20 E5
generates no extruder motor movement at all.
It seems that Extruder.cpp module doesn't care about G2/3 at all:
https://github.com/Smoothieware/Smoothieware/blob/edge/src/modules/tools/extruder/Extruder.cpp#L547

Is this intended?

mdebski:


arthurwolf:
Hey.

Was this tested and confirmed to work ?

Cheers.

mdebski:
Just wrote a comment in #771. Yes.

arthurwolf:
Extruder already knows how to properly handle segments due to the fact we support delta printers, so i'd expect arcs not to be different from "segments per second", and this should work without any bad side effects.

@wolfmanjm do you agree ?

src/modules/tools/extruder/Extruder.cpp
```diff
@@ -544,7 +544,7 @@ void Extruder::on_gcode_execute(void *argument)
             this->target_position += this->travel_distance;
             this->en_pin.set(0);
 
-        } else if (gcode->g == 0 || gcode->g == 1) {
+        } else if (gcode->g == 0 || gcode->g == 1 || gcode->g == 2 || gcode->g = 3) {
```
wolfmanjm: this is better as ``` if(gcode->g <= 3)```

wolfmanjm:
It is hard to say things are so complicated now, but seeing as no slicers generate G2/G3 it shouldn't really be an issue as no one will see it.
One issue is that G2/G3 do not handle the speed limiting added for extruders, but given this is an extreme edge case again it is probably irrelevant.

src/modules/tools/extruder/Extruder.cpp
triffid: Also watch for assignment instead of comparison at the end there! gcode->g
= 3... Compiler should spit a warning about that

mdebski:
Changed to "<= 3".
No, the compiler didn't issue any warnings :/

wolfmanjm:
then you are not using the compiler that is recommended. What tool chain are you using?

mdebski:
Well, now trying incorrect code I get
modules/tools/extruder/Extruder.cpp:547:46: error: lvalue required as left operand of assignment
         } else if (gcode->g <= 3 || gcode->g = 3) {
So even error, not warning.

dunno, probably did sth wrong then.

mdebski:
Oh, fine, the bug was introduced by moving the change to my fork ("just one line, no need to git format-path..."). So the compiled and tested firmware was ok.

PR #775, opened 2015-11-22
Add offset option for AD8495

errolt:
I have added an offset option for the AD8495 amplifier. By default the amplifier output's 0V for 0C. This is sufficient for printers, but some module manufacturers might want to be able to measure negative temperatures. For this they shift the output voltage at 0C up. In the case of the adafruit modules they have shifted 0C up by 1.25V or 250C. Thus adding this option to the config will remove the offset.

temperature_control.hotend.AD8495_offset     250              # AD8495 offset


martink1974:
hi Errolt
that works now great just one thing , I've noticed a difference in capitalisation in config it right be good to make it uniformed as to not cause confusion example of my working config 

temperature_control.hotend.AD8495_pin        0.24             # Pin for the thermistor to read
temperature_control.hotend.heater_pin        2.7              # Pin that controls the heater
temperature_control.hotend.AD8495_offset     250		  # AD8495 offset
temperature_control.hotend.sensor            ad8495        # Weird this needs to be lower case 

thanks again

Commander_K

errolt:
Glad it works for you. :)

Yeah, don't know how the case thing slipped in. Changed everything to lower case as that seems to be the norm...

https://github.com/Smoothieware/Smoothieware/pull/777

PR #782, opened 2015-11-29
Draft temperature runaway detection

arthurwolf:
Hello.

This pull request is not tested. It's meant as a way to discuss *how to* implement thermal runaway detection, rather than as a proper implementation to be merged.

It pretty much copies what Marlin does ( with one exception, marlin seems to only detect problems in one direction ).

This code is intended to detect the following cases : 
* Temperature target has been reached, and thermistor falls off the hotend ( temperature is lower than target for a given duration )
* Temperature target has been reached, and the mosfet dies and stays always on ( temperature is higher than target for a given duration )

The code triggers the HALT state ( much like the MIN_TEMP exception ), and expects that to turn the heaters and the PSU ( in case the mosfets controlling the heaters are failing HIGH ) off.

I am looking for opinions as to how to implement this properly, and any improvements that can be made.

In particular, I do not think this is able to detect problems when we *start* heating.

Cheers.


triffid:
In teacup I wrote one that uses a sliding window to detect fallen
thermistor during warmup, it never got much testing so I never found out
how well it works for people other than me. You could check out that code
for ideas if you want to detect this case

arthurwolf:
@triffid got a link ?

triffid:
https://github.com/Traumflug/Teacup_Firmware/blob/master/heater.c#L192-L246

src/modules/tools/temperaturecontrol/TemperatureControl.cpp
```diff
@@ -510,8 +510,48 @@ void TemperatureControl::pid_process(float temperature)
...
+                }
+                break;
+            case 2: // If we are in state 2 ( target temperature reached ), check for thermal runaway
+                // If the temperature is outside the acceptable range
+                if( fabs( this->get_temperature() - this->target_temperature ) > 10 ){ // 10 degrees, should be a configurable value
+                    // Increase the timer, aka « One more second with a problem maybe occuring »
+                    this->runaway_timer++;
+
+                    // If the timer has a too large value ( we have been too long outside the desired temperature range )
+                    if( this->runaway_timer > 15 ){ // 15 seconds, should be a configurable value
```
wolfmanjm: looks good except you need to make sure once triggered this does not keep happening.

src/modules/tools/temperaturecontrol/TemperatureControl.h
```diff
@@ -80,6 +80,10 @@ class TemperatureControl : public Module {
         float d_factor;
         float PIDdt;
 
+        // Temperature runaway values
+        uint16_t runaway_timer;
+        uint8_t  runaway_state;
```
wolfmanjm: state should be an enum

src/modules/tools/temperaturecontrol/TemperatureControl.cpp
wolfmanjm: add a..
```
if(THEKERNEL->is_halted()) return;
```
so it does not keep halting once halted

Veloce7:
Is there a way to set a threshold value for "temperature_control.module_name.max_temp" so that if the PID's are a little off and the bed temp overshoots by one or two degrees the print won't be halted?

wolfmanjm:
@Veloce7 No why would you need that? just set max_temp a few degrees higher if it is triggering when you don't want it to. It is an emergency/safety feature so what you are suggesting makes no sense.

Veloce7:
I used to work heavily in materials/furnace processing and setting a max temp limits the set point that a user can set. Then there is an alarm temp that is set a little higher that will kill the power in the event of an emergency.

wolfmanjm:
@Veloce7 max_temp sets the cutoff temp, if the temp goes above this then it shuts off, there is no setting to limit the temp a user can set.

dcousens:
I'll be willing to pick this up in a month or so if no progress is made on it by then :+1:

arthurwolf:
@dcousens I was planning on doing this "in a few weeks", but if you are ready to do it, I'd be extremely thankful !

dcousens:
@arthurwolf rather than being so specific, shouldn't the code just account for two cases:

* The temperature should not be changing
* The temperature should be changing

You could then just set minimum and maximum temperature deltas over given time frames depending on which case you are accounting for?

triffid:
That's basically how the one in teacup works ;)

arthurwolf:
On Tue, Mar 29, 2016 at 5:39 AM, Daniel Cousens <notifications@github.com>
wrote:

> @arthurwolf <https://github.com/arthurwolf> rather than being so
> specific, shouldn't the code just account for two cases:
>
>    - The temperature should not be changing
>    - The temperature should be changing
>
> You could then just set minimum and maximum temperature deltas over given
> time frames depending on which case you are accounting for?
>

Yes that does sound smart.


> —
> You are receiving this because you were mentioned.
> <https://github.com/Smoothieware/Smoothieware/pull/782#issuecomment-202693429>
>



-- 
Courage et bonne humeur.

dcousens:
@arthurwolf better/smarter deltas could also come from a manual calibration,  perhaps a configuration option?
Users could set the minimum/maximum delta values depending on their setup,  which they can watch/determine through testing.

arthurwolf:
@dcousens I made a small test rig to be able to test this code, and i'm going to try testing it in the days to come, and make it better if I can. However, I would very very much appreciate it if you took this up as you offered :)

dcousens:
@arthurwolf awesome,  I've almost finished my build for my new CoreXY,  at which point I'll be focused solely on software :+1:

arthurwolf:
Was finally about to test it : 

ok T:28.1 /60.0 @255 B:93.4 /0.0 @0 
ok T:27.9 /60.0 @255 B:94.1 /0.0 @0 
ok T:27.7 /60.0 @255 B:94.6 /0.0 @0 
Error: Temperature runaway ! Check whether the heater controller is failing HIGH, or the temperature sensor is detached. IMMEDIATELY\0xc2\0xa0TURN\0xc2\0xa0OFF\0xc2\0xa0POWER
HALT asserted - reset or M999 required
ok T:26.5 /0.0 @0 B:99.5 /0.0 @0 

Looks like it's working.
I need a way to detect if we take too long to heat up, but it can't take too much ram. I think I'll just implement a "halt if we take longer than x to heat up".

arthurwolf:
I do believe that this now allows to detect the 4 following cases : 
* Thermistor gets detached during ( or before ) initial heatup : heating timeout is triggered
* Mosfet fails always-high during heatup ( or before ) : out-of-range is triggered ( up )
* Thermistor gets detached while at target temp : out-of-range is triggered ( down )
* Mosfet fails always-high while at target temp : out-of-range is triggered ( up )

This still needs testing, documentation, and I'd really like it in somebody had an idea how to make it take less RAM, I think this takes 6+ bytes per temperature control module. But even if we can't fix that, I think the feature is very much worth it.

src/modules/tools/temperaturecontrol/TemperatureControl.cpp
```diff
@@ -510,8 +519,54 @@ void TemperatureControl::pid_process(float temperature)
...
+
+    // Check whether or not there is a temperature runaway issue, if so stop everything and report it
+    if(THEKERNEL->is_halted()) return;
+    
+    if( this->target_temperature <= 0 ){ // If we are not trying to heat, state is NOT_HEATING
+        this->runaway_state = NOT_HEATING;
+        this->runaway_timer = 0;
+    }else{
+        switch( this->runaway_state ){
+            case NOT_HEATING: // If we were previously not trying to heat, but we are now, change to state WAITING_FOR_TEMP_TO_BE_REACHED
```
dcousens: > If we weren't trying to heat,  but we are now, ...

Reads simpler IMHO

src/modules/tools/temperaturecontrol/TemperatureControl.cpp
```diff
@@ -510,8 +519,54 @@ void TemperatureControl::pid_process(float temperature)
...
+                    this->runaway_heating_timer = 0;
+                }
+                break;
+            case WAITING_FOR_TEMP_TO_BE_REACHED: // In we are in state 1 ( waiting for temperature to be reached ), and the temperature has been reached, change to state TARGET_TEMPERATURE_REACHED
+                if( this->get_temperature() >= this->target_temperature ){
+                    this->runaway_state = TARGET_TEMPERATURE_REACHED;
+                }
+                this->runaway_heating_timer++;
+                if( this->runaway_heating_timer > this->runaway_heating_timeout && this->runaway_heating_timeout != 0 ){
+                    this->runaway_heating_timer = 0;
```
dcousens: Could `this->runaway_heating_timer` and `this->runaway_timer` be merged to save 1 byte?
From a once-over,  it looks like they are never used at the same time.

arthurwolf:
I believe I'm going to remove "runaway_timeout", I don't think it is useful.
It'll also make things simpler/easier to understand for users.

Once that's done, configuration will be just two parameters : 
* Heating timeout cuts everything if it takes too long to heat
* Runaway range cuts everything if we go out of a given range

Posting here before doing it in case somebody has a comment. Once that's done I believe this will be ready for merging.

arthurwolf:
Alright, this is now in it's final state. It does what we want, it's been tested, and I believe it's ready to be merged.
I would appreciate it very much if others would test it. Testing is good.
I'm now going to write documentation for it, and once that's done, will wait a bit and then merge it.

arthurwolf:
New documentation is now live at : http://smoothieware.org/temperaturecontrol#toc27

src/modules/tools/temperaturecontrol/TemperatureControl.cpp
arthurwolf: runaway_timer has been removed

303055:
Hi 
the timeout tests were performed without the cartridge heater is connected
the range tests were conducted after the temperature set has been reached.

I try : 
the temperature_control.hotend.runaway_heating_timeout      120 
10 times --> 100% ok

I try  after the timeout :
temperature_control.hotend.runaway_heating_range      20
10 times --> 0% ok

I try only the runaway range
temperature_control.hotend.runaway_heating_range      20
10 times --> 0% ok

Boboss

303055:
Hi
we find the problem with arthur :)
temperature_control.hotend.runaway_heating_range     20  <-- wrong
temperature_control.hotend.runaway_range      20   <---ok
arthur corrected the wiki.

I try :
the temperature_control.hotend.runaway_heating_timeout 120
10 times --> 100% ok

I try after the timeout :
temperature_control.hotend.runaway_heating_range 20
10 times --> 100% ok

I try only the runaway range
temperature_control.hotend.runaway_heating_range 20
10 times --> 100% ok

Boboss

arthurwolf:
Alright, this has been tested by two different persons, reviewed, and the documentation has been added to the wiki. 
I'm merging it.
Thanks a lot to all those who helped.

PR #797, opened 2015-12-09
Avoiding conditional directives that split up parts of statements

RomeroMalaquias:


wolfmanjm:
I am not sure I see the advantage to this PR. You are changing UIP library code (not smoothieware code), and to be honest I cannot test this at the moment and given it changes a core library with no real advantage I am not realy comfortable accepting this. If you could explain why this is necessary, or what bug it fixes, it can be reconsidered.

RomeroMalaquias:
Hi @wolfmanjm, i see your point, this changes doesn't fix bugs, this change is about to improve code understanding and maintainability.
Sorry, i thought that i was changing smoothieware code.
Thanks for your time.

wolfmanjm:
Thank you for the help :) 
Basically I used UiP as is, and would like to keep it as little unchanged as possible. If you really want this change you could submit it to the UiP upstream maintainer (although I do not believe UiP is maintained any longer).

RomeroMalaquias:
Hello @wolfmanjm, i'm doing a research and i need to ask if you prefer this code before or after the commit, what do you think?
Thanks in advance.

RomeroMalaquias:
I mean, would you accept if it was not for the fact we submitted the pull request to a code written by third parties?

PR #798, opened 2015-12-10
Feature/PanelDue

chrisbrent:
Starting a pull request to let you know I'm working on this. Interested in thoughts on handling the various types that might be requested. It's going to be one nasty case statement right now :)
Started case statement for various response types as per:
http://reprap.org/wiki/G-code#M408:_Report_JSON-style_response
and
http://reprap.org/wiki/Proposed_RepRap_Duet_Status_Responses#Status_Response_Definitions


src/modules/tools/reporter/Reporter.cpp
```diff
@@ -0,0 +1,76 @@
...
+    Gcode *gcode = static_cast<Gcode *>(argument);
+    if (gcode->has_m) {
+        if( gcode->m == 408 ) {
+            if (gcode->has_letter('S')) {
+                switch(static_cast<int>(gcode->get_value('S'))){
+                        case 0:
+                              // Beginning
+                              gcode->stream->printf("{\"status\":\"I\",");
+                              gcode->stream->printf("\"heaters{{\":[");
+                              for(auto it = THEKERNEL->temperature_control_pool->get_controllers().begin();
```
wolfmanjm: This is not the correct way to keep a list of temperature controllers, there is a better way and it is used in several other modules. so you will need to change this to comply with the way it is currently done. It also wastes a lot of memory.

src/modules/tools/temperaturecontrol/TemperatureControlPool.cpp
```diff
@@ -29,6 +29,7 @@ void TemperatureControlPool::load_tools()
         // If module is enabled
         if( THEKERNEL->config->value(temperature_control_checksum, cs, enable_checksum )->as_bool() ) {
             TemperatureControl *controller = new TemperatureControl(cs, cnt++);
+            controllers.push_back( cs );
```
wolfmanjm: remove this and all other references to it.

src/libs/Kernel.h
```diff
@@ -56,6 +57,7 @@ class Kernel {
         Planner*          planner;
         Config*           config;
         Conveyor*         conveyor;
+        TemperatureControlPool* temperature_control_pool;
```
wolfmanjm: remove this

wolfmanjm:
Both the panel and the temperature switch  handle getting data from all temperature controls, so please use that method, and not add additional, unnecessary overhead to do the same thing. Thank you.

wolfmanjm:
Plus temperaturecontrolpool does not exist once the config part of the boot has been finished, so this would never have worked anyway you would simply crash.

src/modules/tools/temperaturecontrol/TemperatureControlPool.h
```diff
@@ -13,6 +13,10 @@
 class TemperatureControlPool {
     public:
         void load_tools();
+        const std::vector<uint16_t>& get_controllers() const { return controllers; };
```
wolfmanjm: remove this

wolfmanjm:
This is how you access all temperature controllers... (from SimpleShell)

```
            // scan all temperature controls
            std::vector<struct pad_temperature> controllers;
            bool ok = PublicData::get_value(temperature_control_checksum, poll_controls_checksum, &controllers);
            if (ok) {
                for (auto &c : controllers) {
                   stream->printf("%s (%d) temp: %f/%f @%d\r\n", c.designator.c_str(), c.id, c.current_temperature, c.target_temperature, c.pwm);
                }

            } else {
                stream->printf("no heaters found\r\n");
            }
```

chrisbrent:
Thanks, sorry for the poor code. I originally started doing this off Smothie2 as Arthur posted a link to that in a G+ post. This looks better I think? M408  the following:
heaters: current heater temperatures, numbered as per the machine (typically, heater 0 is the bed)
active:  active temperatures of the heaters
standby: standby temperatures of the heaters
hstat:   status of the heaters, 0=off, 1=standby, 2=active, 3=fault
Do you know if "active" is the target temp? 
It looks like Smoothie doesn't have the concept of a standby temperature?
The TemperatureControl doesn't have a status, is there a class to us for this?

wolfmanjm:
There is no standby temperature for smoothie so you would ignore that. 
You can determine the status of the heater from the returned data.
So if target_temperature is 0 it is off, if target_temperature is > current_temperature then it is heating if target_temperature is equal or close to the current_temperature then it has reached its temperature.

src/modules/tools/reporter/Reporter.cpp
```diff
@@ -0,0 +1,147 @@
...
+                                bool b = *static_cast<bool *>(returned_data);
+                                if(b){
+                                  gcode->stream->printf("A");
+                                }
+                              }
+                              gcode->stream->printf("\",\"heaters{{\":[");
+                              std::vector<struct pad_temperature> controllers;
+                              if (PublicData::get_value(temperature_control_checksum, poll_controls_checksum, &controllers)) {
+                                for (auto &c : controllers) {
+                                  if(&c == &controllers.back()){
```
wolfmanjm: not sure I understand what this is, the results are identical.

src/modules/tools/reporter/Reporter.cpp
```diff
@@ -0,0 +1,147 @@
...
+                              for(auto &c : controllers){
+                                // off
+                                if(c.target_temperature == 0){
+                                  gcode->stream->printf("0");
+                                }
+                                // active
+                                if(c.target_temperature>c.current_temperature){
+                                  gcode->stream->printf("2");
+                                }
+                                if(&c != &controllers.back()){
```
wolfmanjm: This is really not a good way to determine if it is the last one. maybe use an index instead

src/modules/tools/reporter/Reporter.cpp
```diff
@@ -0,0 +1,147 @@
...
+                              // Heater status
+                              // status of the heaters, 0=off, 1=standby, 2=active, 3=fault
+                              gcode->stream->printf("\"hstat\":[");
+                              for(auto &c : controllers){
+                                // off
+                                if(c.target_temperature == 0){
+                                  gcode->stream->printf("0");
+                                }
+                                // active
+                                if(c.target_temperature>c.current_temperature){
```
wolfmanjm: This is not the correct way to determine it is active... if target_temperature is > 0 then it is active it has nothing to do with the current_temperature at all.

src/modules/tools/reporter/Reporter.cpp
```diff
@@ -0,0 +1,147 @@
...
+                              if(PublicData::get_value( extruder_checksum, (void **)&rd )){
+                                gcode->stream->printf("%f", *(rd+5));
+                              } else {
+                                gcode->stream->printf("0");
+                              }
+                              gcode->stream->printf("],\"sfactor\":100.00,\"efactor\":[100.00,100.00],\"tool\":1,");
+                              gcode->stream->printf("\"probe\":\"535\",\"fanRPM\":0,\"homed\":[0,0,0],\"fraction_printed\":");
+                              // Get fraction printed
+                              void *completed_data;
+                              if (PublicData::get_value( player_checksum, get_progress_checksum, &completed_data )) {
```
wolfmanjm: This only works if printing from an sd card.

src/modules/tools/reporter/Reporter.cpp
chrisbrent: Thanks, I mixed up active and heating (which doesn't exist as a status)

src/modules/tools/reporter/Reporter.cpp
chrisbrent: It's supposed to do the , correctly for the Json array but I copy and pasted poorly. I need to rethink this anyway as the heater array for M408 says "current heater temperatures, numbered as per the machine (typically, heater 0 is the bed)".

626Pilot:
Just to be perfectly clear, this is for the PanelDue by dc42, and not some other panel that works with Due controllers, right?

arthurwolf:
Yes, it's for the paneldue

On Tue, Dec 29, 2015 at 6:51 AM, 626Pilot <notifications@github.com> wrote:

> Just to be perfectly clear, this is for the PanelDue by dc42, and not some
> other panel that works with Due controllers, right?
>
> —
> <https://github.com/Smoothieware/Smoothieware/pull/798#issuecomment-167723380>
> .
>



-- 
Courage et bonne humeur.

626Pilot:
Which wires (red, etc.) are being connected to which pins? I have a PanelDue, and I would like to bring this into my fork when it's working.

arthurwolf:
black is gnd, red is 5v, blue is rx, green is tx

On Thu, Dec 31, 2015 at 11:58 AM, 626Pilot <notifications@github.com> wrote:

> Which wires (red, etc.) are being connected to which pins? I have a
> PanelDue, and I would like to bring this into my fork when it's working.
>
> —
> <https://github.com/Smoothieware/Smoothieware/pull/798#issuecomment-168169942>
> .
>



-- 
Courage et bonne humeur.

chrisbrent:
@626Pilot This is a lot more complete now. It's merged with the current edge so if you can build this give it a try.

626Pilot:
@chrisbrent I just merged with upstream, but your code isn't there. I guess that's because it's in a feature. I'll wait until the code is checked into the main repo, and there is some documentation.

arthurwolf:
@chrisbrent Where is this at ? Anything blocking you ?

arthurwolf:
@chrisbrent The pins that are normally used for the RRD GLCD panel ( the SPI connector right next to the microcontroller ) can also be used as a UART ( UART1 ) just by creating a new SerialConsole object using those pins.
Like this : https://github.com/Smoothieware/Smoothieware/blob/edge/src/libs/Kernel.cpp#L60
Don't forget to register it as a module : https://github.com/Smoothieware/Smoothieware/blob/edge/src/libs/Kernel.cpp#L106
And then it should be able to work just like UART0 does now.

chrisbrent:
Yes, I just worked that out :) Sorry it should be obvious by now!

src/main.cpp
```diff
@@ -172,7 +172,17 @@ void init() {
     kernel->add_module( new Spindle() );
     #endif
     #ifndef NO_UTILS_PANEL
+<<<<<<< HEAD
```
wolfmanjm: can't merge something that does not compile... you need to fix these messages.

wolfmanjm:
I'm going to close this until it is fixed. RE open when it builds cleanly.

AHHams:
Is anyone still working on this? Chris?

chrisbrent:
Yes, I just need to find two hours to fix one bug and get the merge sorted.
But finding that time is tricky.

AHHams:
I fixed the issue with main.ccp. It is in a feature/due_panel branch on my fork. It works and is otherwise in sync with latest edge.  

My PanelDue seems to work fine up till now, but I did only very testing to date. Reading / setting temps works,  jogging works. Did not do any prints yet. 

I did have to update the firmware on the PanelDue to latest 1.15c. I did not get it to connect with prior version.

I am new to git/Github. Don't know proper etiquette and process for contributing to pull requests. Let me know if there's anything further I can do to help get this fix into latest edge.

src/modules/tools/reporter/Reporter.cpp
```diff
@@ -0,0 +1,181 @@
...
+                        // Find the status
+                        //I=idle, P=printing from SD card, S=stopped (i.e. needs a reset),
+                        //C=running config file, A=paused, D=pausing, R=resuming, B=busy (running a macro)
+                        gcode->stream->printf("{\"status\":\"");
+                        if(THEKERNEL->conveyor->is_queue_empty()) {
+                                gcode->stream->printf("I");
+                        }else{
+                                gcode->stream->printf("P");
+                        }
+                        if (THEKERNEL->is_halted()) {
```
AHHams: This would result in "IS" if printer is idle and halted?

src/modules/tools/reporter/Reporter.cpp
```diff
@@ -0,0 +1,181 @@
...
+                                gcode->stream->printf("S");
+                        }
+                        void *returned_data;
+                        if (PublicData::get_value( player_checksum, is_suspended_checksum, &returned_data )) {
```
AHHams: This would result in "PA" if queue not empty and paused? Suggest to make code a little safer, so only "I", "P", "S", or "A" are returned.

PR #802, opened 2015-12-12
Implement G54-G59.3 Select Coordinate System, and change G92 handling, merge nmotors

wolfmanjm:
This is still a WIP, no where near ready to merge :)

    add handling for G54-G59.3 Select Coordinate System
    add handling for G10 L2 to set them
    change the way G92 is handled, just sets a global offset which is not persistent across reset
    Save the WCS offsets in M500 and display them in M503 (if they are non zero)

    Merge branch 'feature/nmotors' of https://github.com/trayracing/Smoothieware into feature/wcs


wolfmanjm:
moving to my repo as this is a lot more complex than it originally looked

PR #803, opened 2015-12-13
Replace Hook with std::function

darthscsi:
Replace Hook with std::function.  Prototypes for callbacks are changed to void(void) since none of the uses of hook either pass arguments or return data.  Some unnecessary indirection is removed in SlowTicker.


wolfmanjm:
No I tried this, it increases ram usage a lot, so I reverted it.

Thanks for the suggestion though. It would be nice to do if we had unlimited ram available.

wolfmanjm:
Also in future please clean up your PRs so the comits are squashed. It makes it a lot easier to review.

Thanks

wolfmanjm:
I suggest you do a mem before and after and make sure how much RAM your changes use. If it uses less that is great if it uses more and there is no bug fixed, then it must be rejected. We are extremely short on RAM.

darthscsi:
Bind wasn't triggering small-function optimization in std::function.  Lambdas do, however.  See next pull request.

PR #805, opened 2015-12-14
Hook removal and conversion

darthscsi:
Take 2.  lambdas trigger small function optimization while bind doesn't.  This version Increases free memory on my config from 3376 to 3412.  I believe that is primarily in the conversion of a vector of pointers to a vector of objects in the timer. 

Base:   3376
Lambda: 3412


wolfmanjm:
I hate to be picky, but I really do not see this as a significant improvement.

lambdas use a bunch of stack to pass the context in, and the syntax is much messier than it was.
and it still uses std::function which uses more memory than the old Hook.
The memory saving in this case is nice, but maybe due to the refactor.
I'll have to think a lot about this, because it touches way too much of the system.

Have you tested every edge case this touches? and every function you have changed? like filament detector and other edge cases?

wolfmanjm:
I really have to ask why do this? Where is the advantage? what additional features/functionality does it offer? Exactly how is this any better than the original? These questions need to be considered when you touch major portions of core code which entails a huge amount of testing.

wolfmanjm:
FYI of you check the size results after the build you will see this uses 3k more flash than the previous version. Is it really worth using 3k more flash memory (which we are also running short on) for this refactor? Basically this does exactly what it did before, just in a different way, and whether that way is better is debatable. I really like lambdas and std::function and try to use it in new code, but there is a trade off, and balancing that trade off is tough. Sometimes you just have to use ugly old methods in embedded code just so it will all fit.

darthscsi:
The stack size is equivalent.  This turns into a load and an indirect call (two instructions on x86, similar on arm) and a null pointer check (~2 instructions).   No additional function calls are generated.

darthscsi:
There is also the minor detail that the existing code is illegal c++.  Casting the pointer to a member function from one class to another (namely FPointerDummy) is fine, as long as you never call it on the second class.  You must cast it back first (which it isn't, see the 6th line of call).  Although it will almost certainly work in the way used in the code given the current state of gcc and the lack of lto being applied, as a compiler writer it makes me nervous.

darthscsi:
So, to your points:
Con: more instructions.
Pro: not more stack space.
Pro: memory savings, but almost certainly from the conversion of vector of pointers to vector of objects and other removals of unnecessary heap allocations.
Pro: code is now legal C++
Con: no savings in complexity (arguable, converting to standard mechanism and removing a file seems a savings in complexity).
Con: testing.

I consider removal of code with undefined behavior a sufficient pro to be worth the testing effort.  It is true I don't have the hardware to test everything I touched.

wolfmanjm:
I am not disagreeing with you . I think the refactor is worthy. The FPointer code is a library that a lot of people use, it is somewhat obsolete now with C++11 and std::function though. We used the library as is, and it is still used throughout the mbed code, so it is not actually being removed.
The issue left now is the significant increase in flash usage (with no functionality added), and the lack of testing. Keeping smoothie stable is very important, so this needs to be very well tested, and to be honest I am somewhat busy with other stuff to test this at the moment. So you will have to figure out a way to get it tested.

wolfmanjm:
Actually mbed uses mbed/src/cpp/FunctionPointer.cpp, which looks similar. We could switch to using that :)

darthscsi:
The mbed library does it correctly.  I can update to it.

darthscsi:
The mbed library also uses more memory to do so.

wolfmanjm:
so then leave it as it is, it works.

darthscsi:
3472.  Smaller still.

wolfmanjm:
closed as it was not updated for three months

PR #807, opened 2015-12-18
New/original glyphs for WatchScreen

rparrett:
Here's an alternate set of gylphs for the Watch Screen. I thought that Smoothie could use some graphics that are unique to Smoothie. They come with a modest ~136 byte decrease in flash size.

I plan on looking into improving the look/feel further, but thought I'd test the waters here with a simple PR like this.

Let me know if I've done anything horribly wrong.

glyphs
newglyphs

PR #813, opened 2015-12-29
Fix messed up rrd glyph drawing when drawing not on a byte boundary

rparrett:
Related to https://github.com/Smoothieware/Smoothieware/pull/807


wolfmanjm:
This is already in edge

PR #814, opened 2015-12-31
Feature/responsive filescreen

rparrett:
For feedback/discussion. Proof of concept for the FileScreen improvements I was rambling about on IRC.

Included on-demand FileScreen creation, but that is not actually a requirement.

(Theoretical) one directory traversal per FileScreen menu refresh, instead of 8 (depending on panel)
(Anecdotal) FileScreen feels very responsive now. 

All for the price of 0 additional ram and 272 additional flash.

### DIFF
--- a/src/modules/utils/panel/PanelScreen.h
+++ b/src/modules/utils/panel/PanelScreen.h
@@ -28,7 +28,7 @@ public:
     void refresh_screen(bool clear);
     void refresh_menu(bool clear);
     void refresh_menu(void) { refresh_menu(true); };
-    virtual void display_menu_line(uint16_t line) = 0;
+    virtual void display_menu_line(uint16_t line) {};
     virtual void display_menu_lines(uint16_t line, uint16_t num);
     // default idle timeout for a screen, each screen can override this
     virtual int idle_timeout_secs(){ return 10; }
--- a/src/modules/utils/panel/PanelScreen.h
+++ b/src/modules/utils/panel/PanelScreen.h
@@ -29,6 +29,7 @@ public:
     void refresh_menu(bool clear);
     void refresh_menu(void) { refresh_menu(true); };
     virtual void display_menu_line(uint16_t line) = 0;
+    virtual void display_menu_lines(uint16_t line, uint16_t num);
     // default idle timeout for a screen, each screen can override this
     virtual int idle_timeout_secs(){ return 10; }
 
--- a/src/modules/utils/panel/screens/FileScreen.cpp
+++ b/src/modules/utils/panel/screens/FileScreen.cpp
@@ -71,20 +71,43 @@ void FileScreen::enter_folder(const char *folder)
     this->refresh_menu();
 }
 
-// Called by the panel when refreshing the menu, display .. then all files in the current dir
+
 void FileScreen::display_menu_line(uint16_t line)
 {
-    if ( line == 0 ) {
-        THEPANEL->lcd->printf("..");
-    } else {
-        bool isdir;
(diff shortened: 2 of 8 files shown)


src/modules/utils/panel/screens/MainMenuScreen.cpp
```diff
@@ -97,6 +96,13 @@ void MainMenuScreen::setupConfigureScreen()
     THEPANEL->enter_screen(mvs);
 }
 
+void MainMenuScreen::setupFileScreen()
```
wolfmanjm: This appears to never be called from anywhere

src/modules/utils/panel/screens/MainMenuScreen.cpp
```diff
@@ -136,7 +142,7 @@ void MainMenuScreen::clicked_menu_entry(uint16_t line)
                 send_command("M999");
                 THEPANEL->enter_screen(this->watch_screen);
             }else if(THEPANEL->is_playing()) abort_playing();
-             else THEPANEL->enter_screen(this->file_screen); break;
+             else THEPANEL->enter_screen((new FileScreen())->set_parent(this)); break;
```
wolfmanjm: maybe this is supposed to call setupFileScreen?

src/modules/utils/panel/screens/MainMenuScreen.cpp
rparrett: There's a call in Panel.cpp

src/modules/utils/panel/screens/MainMenuScreen.cpp
rparrett: but it would make sense to do the same thing here.

src/modules/utils/panel/PanelScreen.h
```diff
@@ -29,6 +29,7 @@ class PanelScreen
     void refresh_menu(bool clear);
     void refresh_menu(void) { refresh_menu(true); };
     virtual void display_menu_line(uint16_t line) = 0;
+    virtual void display_menu_lines(uint16_t line, uint16_t num);
```
wolfmanjm: why is this virtual?

src/modules/utils/panel/screens/FileScreen.cpp
```diff
@@ -69,20 +71,43 @@ void FileScreen::enter_folder(const char *folder)
     this->refresh_menu();
 }
 
-// Called by the panel when refreshing the menu, display .. then all files in the current dir
+
 void FileScreen::display_menu_line(uint16_t line)
```
wolfmanjm: should remove this if it is not used, and remove the pure virtual in panelscreen (remove the =0)

rparrett: I think to do this, I need to provide a (no-op)? implementation of display_menu_line in PanelScreen.

Is that acceptable?

rparrett:
Made the suggested changes. I can see how this dirties up the PanelScreen classes/subclasses a bit. If anyone has any suggestions for accomplishing the same optimization in a cleaner way, I'm all ears.

wolfmanjm:
It looks Ok, I just need to test it on a 4 line screen as well as the 8 line ones.

wolfmanjm:
FYI I think it does use more RAM as there is now another virtual method which increases the size of the vtable. I'll see how much extra ram this causes. I think it can be done without adding the extra virtual method if that is the case.

rparrett:
Thanks for making me aware of that possibility. I based my assertion on the output of "mem" before/after.

rparrett:
I'm actually seeing some issues with long filenames on my viki2.

PR #820, opened 2016-01-07
One line of code. Simple fix on a printf text typo.

acurana:
This fixes some text typos in Touchprobe.


wolfmanjm:
touchprobe is no longer supported or compiled. it will be removed from the code base

PR #821, opened 2016-01-11
feature: Soft Limits

qharley:
Added config controlled soft limits to the cartesian movement of any system

config example for a machine with dimentions X:380 Y220 Z200
* Cartesian axis soft limits
software_limits                              true
software_limits.x_axis_min                   0.0
software_limits.x_axis_max                   380.0
software_limits.y_axis_min                   0.0
software_limits.y_axis_max                   220.0
software_limits.z_axis_min                   -10.0
software_limits.z_axis_max                   210.0

Movement will be constrained to the set build envoronment except if
the last_milestone is outside the build environment.  In that case it
will not allow movements further from the build environment, only closer.

### DIFF
--- a/src/modules/robot/Robot.cpp
+++ b/src/modules/robot/Robot.cpp
@@ -65,6 +65,14 @@ using std::string;
 #define  kossel_checksum                     CHECKSUM("kossel")
 #define  morgan_checksum                     CHECKSUM("morgan")
 
+#define  software_limits_checksum            CHECKSUM("software_limits")
+#define  x_axis_min_checksum                 CHECKSUM("x_axis_min")
+#define  x_axis_max_checksum                 CHECKSUM("x_axis_max")
+#define  y_axis_min_checksum                 CHECKSUM("y_axis_min")
+#define  y_axis_max_checksum                 CHECKSUM("y_axis_max")
+#define  z_axis_min_checksum                 CHECKSUM("z_axis_min")
+#define  z_axis_max_checksum                 CHECKSUM("z_axis_max")
+
 // new-style actuator stuff
 #define  actuator_checksum                   CHEKCSUM("actuator")
 
@@ -183,6 +191,14 @@ void Robot::load_config()
     this->max_speeds[X_AXIS]  = THEKERNEL->config->value(x_axis_max_speed_checksum    )->by_default(60000.0F)->as_number() / 60.0F;
     this->max_speeds[Y_AXIS]  = THEKERNEL->config->value(y_axis_max_speed_checksum    )->by_default(60000.0F)->as_number() / 60.0F;
     this->max_speeds[Z_AXIS]  = THEKERNEL->config->value(z_axis_max_speed_checksum    )->by_default(  300.0F)->as_number() / 60.0F;
+
+    this->software_limits       = THEKERNEL->config->value(software_limits_checksum    )->by_default( false )->as_bool();
+    this->softlimits[X_AXIS][0] = THEKERNEL->config->value(software_limits_checksum, x_axis_min_checksum    )->by_default(  0.0F)->as_number();
+    this->softlimits[X_AXIS][1] = THEKERNEL->config->value(software_limits_checksum, x_axis_max_checksum    )->by_default(  0.0F)->as_number();
+    this->softlimits[Y_AXIS][0] = THEKERNEL->config->value(software_limits_checksum, y_axis_min_checksum    )->by_default(  0.0F)->as_number();
+    this->softlimits[Y_AXIS][1] = THEKERNEL->config->value(software_limits_checksum, y_axis_max_checksum    )->by_default(  0.0F)->as_number();
+    this->softlimits[Z_AXIS][0] = THEKERNEL->config->value(software_limits_checksum, z_axis_min_checksum    )->by_default(  0.0F)->as_number();
+    this->softlimits[Z_AXIS][1] = THEKERNEL->config->value(software_limits_checksum, z_axis_max_checksum    )->by_default(  0.0F)->as_number();
 
     // Make our 3 StepperMotors
     uint16_t const checksums[][5] = {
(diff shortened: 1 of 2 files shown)


arthurwolf:
The fact that we already have [actuator]_[min or max] config options is going to be confusing to the users, but I guess that's going to be a job for the documentation to explain ...

The code does look ok to me at first glance.

wolfmanjm:
Sorry but this is not the way to do soft limits IMO, if done at all they need to be done in Endstops the way hard limits are done.

wolfmanjm:
I'll be adding a different version of soft limits the way linuxcnc does them. This implementation however will be more of a problem than a help as it will not know if you are homed or not, thus you won't be able to move or jog on power up.

qharley:
It works for me this way.

What are the disadvantages of doing it in robot?

qharley:
Soft stops should be soft after all.  if the movement is limited in the
planned control it can do de-acceleration at the right time, whereas
end-stops will cause a hard stop.  This is not used during printing unless
you do not know what you are doing, but prevents inexperienced users from
jogging the machines into physical end-stops by mistake, causing noise and
more confusion.

No problem.  I will keep it in my local repository.
Cheers,
Q

arthurwolf:
I actually think we want both : an option to act like limit switches if soft endstops are set, and an option to simply ignore/cut movement that would go outside the boundaries. I think both are useful.

wolfmanjm:
you are using more ram that is already defined in endstops, and we cannot afford to waste ram like that. Also you need to know when you are homed or not to turn it on or off, it is NOT ok to have this on when not homed. Look at how linuxcnc handles the issue. That is the correct way to do it.

wolfmanjm:
No the machine needs to go into halt state if limits are breached, there is no worth in ignoring movers that go outside the limits it would cause absolute chaos.

qharley:
Ok.  Don't fret.  My customers wants Marlin like soft end-stops. This is
what I implemented, and until Smoothieware does it or something like it
natively, I will keep it in my local repositories.

Thanks,
Q

wolfmanjm:
This would be disastrous for anyone doing 3D printing BTW. this would change the shape of any object intersecting the boundaries, and that requires re slicing, you will have really bad prints if even part of the object intersect the limits. It will extrude incorrectly too, probably spewing plastic everywhere at the edges as yo are not recalculating the E term for a line that is being cut.

qharley:
It will not come be a problem when printing, not on my machines anyway.
They are factory configured, and customers use my slicing profiles that
define the boundaries of the machines.

The only reason I added it was because a machine used in a makerspace keeps
getting into strange positions when jogged - many inexperienced users, and
this causes belt tension issues.
The result is that people do not use the machine - not ideal.

Getting a machine in halt every time you press a wrong jog button will have
much the same effect in my opinion.
I know this is not as correct as other solutions, but limiting frustration
for users and myself is worth it.

greyltc:
What's the status on the soft limit feature? Is this something that still needs to be implemented?

arthurwolf:
It's still something that needs to be implemented. If you are interested,
we can discuss it. There has been a few discutions about it recently, I
think it'd be a good idea to write a spec for it like we did for some other
features, then discuss that spec, then find folks that are ready to
implement it.

On Fri, May 20, 2016 at 2:44 PM, Grey Christoforo <notifications@github.com>
wrote:

> What's the status on the soft limit feature? Is this something that still
> needs to be implemented?
>
> —
> <https://github.com/Smoothieware/Smoothieware/pull/821#issuecomment-220595792>
>



-- 
Courage et bonne humeur.

qharley:
I implemented soft limits for the Morgan printers, and it is working very
nicely, but the pull request was rejected.

arthurwolf:
Yes, we need to have a discution about what we want soft endstops to do,
and how to implement it, for all arm solutions and uses, and once we have
that, start working on it.
We'll get there, any help is very welcome.

On Fri, May 20, 2016 at 2:51 PM, Quentin Harley <notifications@github.com>
wrote:

> I implemented soft limits for the Morgan printers, and it is working very
> nicely, but the pull request was rejected.
>
>
> On Fri, May 20, 2016 at 2:49 PM Arthur Wolf <notifications@github.com>
> wrote:
>
> > It's still something that needs to be implemented. If you are interested,
> > we can discuss it. There has been a few discutions about it recently, I
> > think it'd be a good idea to write a spec for it like we did for some
> other
> > features, then discuss that spec, then find folks that are ready to
> > implement it.
> >
> > On Fri, May 20, 2016 at 2:44 PM, Grey Christoforo <
> > notifications@github.com>
> > wrote:
> >
> > > What's the status on the soft limit feature? Is this something that
> still
> > > needs to be implemented?
> > >
> > > —
> > > You are receiving this because you commented.
> > > Reply to this email directly or view it on GitHub
> > > <
> >
> https://github.com/Smoothieware/Smoothieware/pull/821#issuecomment-220595792
> > >
> > >
> >
> >
> >
> > --
> > Courage et bonne humeur.
> >
> > —
> > You are receiving this because you authored the thread.
> > <
> https://github.com/Smoothieware/Smoothieware/pull/821#issuecomment-220596651
> >
> >
>
> —
> <https://github.com/Smoothieware/Smoothieware/pull/821#issuecomment-220597196>
>



-- 
Courage et bonne humeur.

greyltc:
In case it helps with designing a solution to fit people's needs, here's what I'm looking for:

I've only got hardware limit switches on one side of my axes. I'd like it if I could define some maximum travel limits in my config file so that if some gcode asks my machine to travel beyond these limits, a safety stop would be triggered just like what happens with my hardware switches when they're hit when I have `alpha_limit_enable` set in my config file.

qharley:
The way it works in my machines is arm solution agnostic... Which is
probably why it was rejected to begin with.  It does a small modification
of the Robot code.
It will allow the head to move into the active region, even when coming
from a home position outside of the active region, but not out.  We needed
this for Morgans in makerspaces so that people do not jog the arms into
positions beyond the magnetic end stops that can cause some disconcerting
skipping  noise when the arms run into the mechanical ends if homed.

Details here:
https://github.com/RepRapMorgan/Smoothieware/commit/1577a885dfe04ae26d7283771ed59b80f9065934

You can always try it out by using the Morgan smoothie firmware.  It is
basically Smoothie edge with a couple of custom items added, but it will
work for all other machines as well.
https://github.com/RepRapMorgan/Smoothieware/blob/Morgan/FirmwareBin/firmware.bin?raw=true

PR #822, opened 2016-01-12
needs definition in cpp file to compile with optimization off (-O0)

acurana:
When switching off optimization in current edge branch (-O0) Smoothieware does not compile. A definition for a static member is missing. Note: it compiles with -O2 because the static is const and the compiler optimizes out the static entirely. Therefore it will not eat any memory, it is always optimized out in -O2.


wolfmanjm:
This is not the correct fix ```k_max_wcs``` is supposed to be a constant not a static. It should not use up any ram. I'll fix it correctly.

Thank you.

PR #829, opened 2016-01-24
Merge edge

chrisbrent:


chrisbrent:
Arg I keep going the wrong way sorry.

PR #831, opened 2016-01-27
Zgrid homing

qharley:
This pull request does the following:
1) repair Z offset after running routine with a well adjusted probe
2) Adds an optional manual probe attachment point
3) changes the normalisation method to zero the home position, preventing jerky movements on the first move after homing.

### DIFF
--- a/ConfigSamples/Snippets/ZprobeGrid.config
+++ b/ConfigSamples/Snippets/ZprobeGrid.config
@@ -8,7 +8,7 @@ leveling-strategy.ZGrid-leveling.bed_z           20
 # Manual probe attachment point (Defaults to the calculated values above)
 # leveling-strategy.ZGrid-leveling.bed_x          100
 # leveling-strategy.ZGrid-leveling.bed_y          100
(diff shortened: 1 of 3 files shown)


ConfigSamples/Snippets/ZprobeGrid.config
```diff
@@ -5,6 +5,11 @@ leveling-strategy.ZGrid-leveling.bed_y          200        # Bed Size
 # Machine height, used to determine probe attachment point (bed_z / 2)
 leveling-strategy.ZGrid-leveling.bed_z           20
 
+# Manual probe attachment point (Defaults to the calculated values above)
+# leveling-strategy.ZGrid-leveling.bed_x          100
+# leveling-strategy.ZGrid-leveling.bed_y          100
+# leveling-strategy.ZGrid-leveling.bed_y          50
```
wolfmanjm: should this be z?

qharley: Yes

PR #832, opened 2016-01-27
Endstops: limits condition

qharley:
Repaired homing: non-delta machines always moved out of the endstop positions, even is no limits were enabled.

### DIFF
--- a/src/modules/tools/endstops/Endstops.cpp
+++ b/src/modules/tools/endstops/Endstops.cpp
@@ -717,7 +717,8 @@ void Endstops::on_gcode_received(void *argument)
             if(!is_delta) {
                 if(this->move_to_origin_after_home) move_to_origin(axes_to_move);
                 // if limit switches are enabled we must back off endstop after setting home
-                back_off_home(axes_to_move);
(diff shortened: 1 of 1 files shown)


wolfmanjm:
This is unnecessary as the back_off_home already checks if limits are enabled for each axis.
See https://github.com/Smoothieware/Smoothieware/blob/edge/src/modules/tools/endstops/Endstops.cpp#L330

qharley:
I have no limits enabled, and it still backs off Z.
Why then?

wolfmanjm:
is is_delta or is_scara set?

qharley:
Is scara

wolfmanjm:
then remove the is_scara test here https://github.com/Smoothieware/Smoothieware/blob/edge/src/modules/tools/endstops/Endstops.cpp#L320

qharley:
Ok

wolfmanjm:
and add it here https://github.com/Smoothieware/Smoothieware/blob/edge/src/modules/tools/endstops/Endstops.cpp#L717

PR #840, opened 2016-02-05
Initalise ethernet pin connections correctly

PeeJay:
Previously the ethernet initialisation would clobber PINSEL2 + 3
registers, thereby removing functionality from any previously
configured pins. Eg. ADC4, ADC5 which use PINSEL3.

Signed-off-by: Paul Jones <paul@pauljones.id.au>

### DIFF
--- a/src/libs/Network/Drivers/LPC17XX_Ethernet.cpp
+++ b/src/libs/Network/Drivers/LPC17XX_Ethernet.cpp
@@ -200,9 +200,10 @@ LPC17XX_Ethernet::LPC17XX_Ethernet()
 
 void LPC17XX_Ethernet::on_module_loaded()
 {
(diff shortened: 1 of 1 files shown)


wolfmanjm:
@triffid please comment and approve/disapprove. Thanks

triffid:
Pinsel2 change looks sensible enough, but the pinsel3 change is completely
equivalent and thus unnecessary.

2<<0 is the same as 1<<1 but I find the former more readable since there's
two bits per pin and we're masking off the higher of the pair

PeeJay:
Notice that the original line 204 is missing the negation operator, so it is clearing all bits except 0-3. ADC4+5 use bits 28/29 + 30/31 set to 0b11, and these were being reset to 00. I found it with the debugger. My thermistors are working now too.

triffid:
You're right, I missed that.. Too busy seeing what should be there rather
than what actually is there ;)

Care to simply add appropriate negation operator rather than changing both
pinsel3 lines?

PeeJay:
I just made both lines consistent in style and function, but I’m not fussed as long as it works.

PR #848, opened 2016-02-16
Feature/laser ttl

FredMurphy:
Hi. This is my first ever contribution to an open source project and first ever pull request, so I kept it simple.I hope I've got it right.

The first commit (4287ba0) adds an optional _laser_module_ttl_pin _output. It's set when the laser is fired on G1/2/3 moves. Existing PWM functionality is not affected, whether the TTL pin is used or not.

The second commit (39b0a65) optionally allows the PWM pin to be referred to as _laser_module_pwm_pin_ so it's not confused with the TTL pin. The existing _laser_module_pin_ name is retained for backward compatibility.

Tested using a logic analyzer and also on my DC-K40 CO2 laser.

### DIFF
--- a/src/modules/tools/laser/Laser.cpp
+++ b/src/modules/tools/laser/Laser.cpp
@@ -23,6 +23,7 @@
 
 #define laser_module_enable_checksum        CHECKSUM("laser_module_enable")
 #define laser_module_pin_checksum           CHECKSUM("laser_module_pin")
+#define laser_module_pwm_pin_checksum       CHECKSUM("laser_module_pwm_pin")
 #define laser_module_ttl_pin_checksum    	CHECKSUM("laser_module_ttl_pin")
 #define laser_module_pwm_period_checksum    CHECKSUM("laser_module_pwm_period")
 #define laser_module_maximum_power_checksum CHECKSUM("laser_module_maximum_power")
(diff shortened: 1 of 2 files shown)


src/modules/tools/laser/Laser.cpp
```diff
@@ -54,13 +60,22 @@ void Laser::on_module_loaded() {
...
     dummy_pin = NULL;
 
-    this->laser_pin->period_us(THEKERNEL->config->value(laser_module_pwm_period_checksum)->by_default(20)->as_number());
-    this->laser_pin->write(this->laser_inverting ? 1 : 0);
+    // TTL settings
+    this->ttl_pin = new Pin();
+    ttl_pin->from_string( THEKERNEL->config->value(laser_module_ttl_pin_checksum)->by_default("nc" )->as_string())->as_output();
+    this->ttl_used = ttl_pin->connected();
+    this->ttl_inverting = ttl_pin->inverting;
+    if (ttl_used)
```
wolfmanjm: You need to delete the pin if it is not being used. otherwise it is using up memory for no reason.

FredMurphy: Thanks, wolfmanjm. I'll get that fixed.

FredMurphy:
Fixed the failure to delete unused pin as noted by wolfmanjm.

PR #853, opened 2016-02-21
Change the initial MR0 value which otherwise generates a large number…

treepleks:
The initial value of the Match Register 0 (MR0) for Timer 2 which is used for the SlowTicker is set to 10000. A large number of interrupts are fired during Smoothie initialization, before MR0 is correctly configured. This generates a large useless (or worse) number of calls (eg. second tick events and SlowTicker attached hooks).

This simple edit sets the initial value to 1000000 instead, removing this large number of useless spurious interrupts/timer based events.

NB: I noticed this as I was extending the SimpleShell module so that it feeds an ESP8266 module connected to the (non USB) UART every second, using the ON_SECOND_TICK event.  When Smoothieware boots, it initially emits the ON_SECOND_TICK event a large number of times, in less than one seconds. This is fixed with the commit above.

The "wifi module" code is in commit 35740cb434c7aeb1a63804a3a51c4d4c2d8053cd on my edge branch if you are interested.

### DIFF
--- a/src/libs/SlowTicker.cpp
+++ b/src/libs/SlowTicker.cpp
@@ -27,7 +27,6 @@ SlowTicker::SlowTicker(){
     max_frequency = 0;
     global_slow_ticker = this;
 
-
     // ISP button FIXME: WHy is this here?
     ispbtn.from_string("2.10")->as_input()->pull_up();
 
@@ -37,7 +36,7 @@ SlowTicker::SlowTicker(){
 
     // Configure the actual timer after setup to avoid race conditions
     LPC_SC->PCONP |= (1 << 22);     // Power Ticker ON
-    LPC_TIM2->MR0 = 10000;          // Initial dummy value for Match Register
+    LPC_TIM2->MR0 = 1000000;        // Initial dummy value for Match Register
     LPC_TIM2->MCR = 3;              // Match on MR0, reset on MR0
     LPC_TIM2->TCR = 1;              // Enable interrupt
     NVIC_EnableIRQ(TIMER2_IRQn);    // Enable interrupt handler
@@ -84,7 +83,6 @@ void SlowTicker::tick(){
     // TODO: This should have it's own module
     if (ispbtn.get() == 0)
         __debugbreak();
-
 }
 
 bool SlowTicker::flag_1s(){


arthurwolf:
Hey.

I'll let jim comment on the specifics of this, but firing on_second_tick
many times at boot does indeed sound like something that needs to be fixed.

about your wifi stuff, that sounds interresting, but why are you having the
smoothie feed things to the esp instead of having it query the smoothie ?

cheers.

On Sun, Feb 21, 2016 at 6:16 PM, T. Reepleks <notifications@github.com>
wrote:

> The initial value of the Match Register 0 (MR0) for Timer 2 which is used
> for the SlowTicker is set to 10000. A large number of interrupts are fired
> during Smoothie initialization, before MR0 is correctly configured. This
> generates a large useless (or worse) number of calls (eg. second tick
> events and SlowTicker attached hooks).
>
> This simple edit sets the initial value to 1000000 instead, removing this
> large number of useless spurious interrupts/timer based events.
>
> NB: I noticed this as I was extending the SimpleShell module so that it
> feeds an ESP8266 module connected to the (non USB) UART every second, using
> the ON_SECOND_TICK event. When Smoothieware boots, it initially emits the
> ON_SECOND_TICK event a large number of times, in less than one seconds.
> This is fixed with the commit above.
>
> The "wifi module" code is in commit 35740cb
> <https://github.com/Smoothieware/Smoothieware/commit/35740cb434c7aeb1a63804a3a51c4d4c2d8053cd>
> on my edge branch if you are interested.
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/Smoothieware/Smoothieware/pull/853
> Commit Summary
>
>    - Change the initial MR0 value which otherwise generates a large
>    number of initial counts.
>
> File Changes
>
>    - *M* src/libs/SlowTicker.cpp
>    <https://github.com/Smoothieware/Smoothieware/pull/853/files#diff-0>
>    (4)
>
> Patch Links:
>
>    - https://github.com/Smoothieware/Smoothieware/pull/853.patch
>    - https://github.com/Smoothieware/Smoothieware/pull/853.diff
>
> —
> <https://github.com/Smoothieware/Smoothieware/pull/853>.
>



-- 
Courage et bonne humeur.

treepleks:
Hello Arthur,

Note that the ESP module is not acting just as a serial bridge, it gives
a full web interface (similar to a pronterface screen) to Smoothieware.

Leaving Smoothie "as is" would be ideal. But I started this on the basis
of a Repetier oriented version (https://github.com/luc-github/ESP8266)
which does querying but also relies on a version of Repetier that sends
info to the ESP on its own
(https://github.com/luc-github/Repetier-Firmware-0.92).

This generates a lot of complexity in the ESP module sending M105, M220,
M221, M114... and then parsing the results to identify where things are.
The ESP8266 code is ok but far less mature than Smoothieware and not
event/IRQ driven. The M codes do not give all the information I want:
eg, the number of configured extruders - my "under design" printer has 2
extruders. This should allow javascript (on the web client) to generate
the graphics to represent the extra extruder and its associated
status/controls.

The ESP module is already struggling a bit with web page generation and
TCP-IP/Wifi. It just seems simpler to simply send an easy to parse
string from Smoothie, synchronously. The SimpleShell looked like a
suitable place to do this.

Your opinion is welcome.
T.

arthurwolf:
If you need info ( like number of extruders ), we/you can add Gcodes that
return it, that's no problem. But hosts should be querying Smoothie, that's
a saner and better long-term solution.

On Sun, Feb 21, 2016 at 7:46 PM, T. Reepleks <notifications@github.com>
wrote:

> Hello Arthur,
>
> Note that the ESP module is not acting just as a serial bridge, it gives
> a full web interface (similar to a pronterface screen) to Smoothieware.
>
> Leaving Smoothie "as is" would be ideal. But I started this on the basis
> of a Repetier oriented version (https://github.com/luc-github/ESP8266)
> which does querying but also relies on a version of Repetier that sends
> info to the ESP on its own
> (https://github.com/luc-github/Repetier-Firmware-0.92).
>
> This generates a lot of complexity in the ESP module sending M105, M220,
> M221, M114... and then parsing the results to identify where things are.
> The ESP8266 code is ok but far less mature than Smoothieware and not
> event/IRQ driven. The M codes do not give all the information I want:
> eg, the number of configured extruders - my "under design" printer has 2
> extruders. This should allow javascript (on the web client) to generate
> the graphics to represent the extra extruder and its associated
> status/controls.
>
> The ESP module is already struggling a bit with web page generation and
> TCP-IP/Wifi. It just seems simpler to simply send an easy to parse
> string from Smoothie, synchronously. The SimpleShell looked like a
> suitable place to do this.
>
> Your opinion is welcome.
> T.
>
> Le 21/02/2016 18:24, Arthur Wolf a écrit :
> > Hey.
> >
> > I'll let jim comment on the specifics of this, but firing on_second_tick
> > many times at boot does indeed sound like something that needs to be
> > fixed.
> >
> > about your wifi stuff, that sounds interresting, but why are you
> > having the
> > smoothie feed things to the esp instead of having it query the smoothie ?
> >
> > cheers.
>
> —
> <https://github.com/Smoothieware/Smoothieware/pull/853#issuecomment-186883870>
> .
>



-- 
Courage et bonne humeur.

wolfmanjm:
good catch, however it is not really an issue as even though it may interrupt there is nothing attached to the interrupts so nothing actually gets called. As soon as the first module registers a callback the frequency is set correctly, so this really does not fix much. The actual issue is enabling the interrupt at all in the constructor, that is wrong.
So I'd prefer to simply not enable interrupts until the setup  process is complete, like the stepticker does, so I will fix it that way instead of this way.

Thanks for the heads up.

wolfmanjm:
I fixed this in PR #854, thanks for the notice.

wolfmanjm:
regarding the push/pull issue I agree with @arthurwolf you need to pull the info not push it. but it is up to you. I think nearly all the info you need is available via either commands or m codes. look at the new grbl queries that were recently added for instance.

treepleks:
Hello Jim,

> good catch, however it is not really an issue as even though it may
> interrupt there is nothing attached to the interrupts so nothing
> actually gets called.
>
Granted, not really an issue (it never bothered any of us, the
Smoothieware users/fans). But the "on_second_tick" routine in
SimpleShell (and all other subscribers) was called something like 30-50
times in less than a second upon boot.

I didn't dig into the SlowTicker code to fully understand it, but the
flag_1s_flag looks like a counter and is probably incremented a large
number of time before MR0 is set. This later gives a large number of
"true" returns from flag_1s() and thus a lot of
call_event(ON_SECOND_TICK) in the on_idle() of SlowTicker.

> As soon as the first module registers a callback the frequency is set
> correctly, so this really does not fix much. The actual issue is
> enabling the interrupt at all in the constructor, that is wrong.
> So I'd prefer to simply not enable interrupts until the setup process
> is complete, like the stepticker does, so I will fix it that way
> instead of this way.
>
Much better solution indeed !
T.

treepleks:
Ok. I understand this from the perspective of Smoothie (answering hosts
requests only). And so I haven't and won't ask for a pull. It's probably
better to keep Smoothie clean :-)

I will come back to this later. I have shifted to the ESP8266 code (and
have time only in the week-end). Having (a dedicated M) code in
Smoothieware that directly spits JSON formatted info that the ESP can
directly forward to the web client w/o any parsing is so comfortable :-)

arthurwolf:
The PanelDue wants a Gcode like that ( M408 I think ) and some folks are
working on implementing it ( search the issues or the forum, I think ).
It'd still be better if people would use the normal gcodes, but panels and
web interfaces both insist on getting a json gcode.

On Mon, Feb 22, 2016 at 8:47 PM, T. Reepleks <notifications@github.com>
wrote:

> Ok. I understand this from the perspective of Smoothie (answering hosts
> requests only). And so I haven't and won't ask for a pull. It's probably
> better to keep Smoothie clean :-)
>
> I will come back to this later. I have shifted to the ESP8266 code (and
> have time only in the week-end). Having (a dedicated M) code in
> Smoothieware that directly spits JSON formatted info that the ESP can
> directly forward to the web client w/o any parsing is so comfortable :-)
>
> Le 21/02/2016 21:58, Jim Morris a écrit :
> >
> > regarding the push/pull issue I agree with @arthurwolf
> > <https://github.com/arthurwolf> you need to pull the info not push it.
> > but it is up to you. I think nearly all the info you need is available
> > via either commands or m codes. look at the new grbl queries that were
> > recently added for instance.
> >
> > —
> > <
> https://github.com/Smoothieware/Smoothieware/pull/853#issuecomment-186914607
> >.
>
> >
>
> —
> <https://github.com/Smoothieware/Smoothieware/pull/853#issuecomment-187347455>
> .
>



-- 
Courage et bonne humeur.

PR #877, opened 2016-03-20
more useful error message from temperature control

wolfmanjm:


src/modules/communication/GcodeDispatch.cpp
```diff
@@ -31,7 +31,7 @@
 #define panel_checksum             CHECKSUM("panel")
 
 // goes in Flash, list of Mxxx codes that are allowed when in Halted state
-static const int allowed_mcodes[]= {105,114,119,80,81,911,503,106,107}; // get temp, get pos, get endstops etc
+static const int allowed_mcodes[]= {2,5,9,30,105,114,119,80,81,911,503,106,107}; // get temp, get pos, get endstops etc
```
dcousens: Any reason for not using enums?

PR #879, opened 2016-03-20
Fix mistaken fwrite (should be fread)

hyperair:
This should fix the "Failed to read grid radius" error when using M375 with delta-grid.


wolfmanjm:
good catch, thank you

PR #887, opened 2016-03-28
modify how homing order is set.

wolfmanjm:
Make sure that endstops set as nc are not used.


src/modules/tools/endstops/Endstops.cpp
```diff
@@ -670,13 +670,22 @@ void Endstops::process_home_command(Gcode* gcode)
 
     // Do we move select axes or all of them
     char axes_to_move = 0;
-    // only enable homing if the endstop is defined, deltas, scaras always home all axis
-    bool home_all = this->is_delta || this->is_rdelta || this->is_scara || !( gcode->has_letter('X') || gcode->has_letter('Y') || gcode->has_letter('Z') );
 
-    for ( int c = X_AXIS; c <= Z_AXIS; c++ ) {
-        if ( (home_all || gcode->has_letter(c + 'X')) && this->pins[c + (this->home_direction[c] ? 0 : 3)].connected() ) {
-            axes_to_move += ( 1 << c );
+    // deltas, scaras always home all axis
```
dcousens: `scaras`?

dcousens:
>set as nc

Do you mean normally closed?
Why would they not be used in the homing order?

PR #891, opened 2016-03-31
Max31855: stop using SPI from ISR

greyltc:
This addresses https://github.com/Smoothieware/Smoothieware/issues/711 by doing SPI comms to the Max31855 thermocouple ADC in `on_idle()` instead of in code called from an ISR.


greyltc:
I've tested this PR in my setup (with Max31855 temperature sensing).

It should definitely not be merged until someone has tested it with thermistor-based temperature sensing. Could someone else help with that please?

hakalan:
Hi!
Isn't there a synchronization issue between ringbuffer accesses from isr and on_idle?
The timer interrupt can occur while on_idle is modifying the ring buffer, leading to incorrect buffer access.

triffid:
The advantage of ringbuffers is that they're immune to concurrent access
problems if used correctly, as long as writing the index pointers is atomic

greyltc:
About the ring buffer: How important is it to have the moving average that the ring buffer gives us?
I was thinking about removing averaging here completely (in another PR).
I suspect it might not matter anyway since I think @wolfmanjm plans to redo all that using the adc like for the thermistor.

hakalan:
Ok, I accept that the ringbuffer access is safe if you say so. :)

The reason I added a moving average in the first place was to try to limit the effect of the 0.25 degree quantization of the Max31855 readouts. Since the PID regulator is updated often but does not filter the signal (at least not when I wrote the thermocouple code in the first place, the situation might have changed?) the derivative term became too unstable and made the autotuning fail unless some smoothing was applied.
So this should probably be solved in PID algorithm instead.

I don't use my Smoothie anymore, otherwise I'd be glad to help...

src/modules/tools/temperaturecontrol/max31855.h
```diff
@@ -21,9 +21,10 @@ class Max31855 : public TempSensor
     ~Max31855();
     void UpdateConfig(uint16_t module_checksum, uint16_t name_checksum);
     float get_temperature();
+    void on_idle();
+    struct { bool read_flag:1; } ; //when true, the next call to on_idle will read a new temperature value
```
wolfmanjm: this needs to be moved to private.

arthurwolf:
Anybody want to test this ?

zerox613:
Hi, I don't have any coding experience, but I have ordered 2 Max31855 (from adafruit) for the smoothieboard to find out they don't work together at the same time. I'd be happy to help with testing, just let me know what needs to be testet.

JeffreyRodriguez:
The conflict here is a very minor syntax issue (extra curly braces). I've merged greyltc's branch with edge and sent him a pull request for this branch.

To replicate the comments on that PR:

I've verified this works with a max31855 and bed thermistor simultaneously.

There is a quirk to the max31855 and max6675 chips: They take 100 and 220 milliseconds, respectively, to perform the temp conversion. Sampling them faster than this will interrupt the conversion process, causing the chip to return the same value every for every subsequent sampling. To resolve that issue, readings_per_second should be 9 for the max31855 and 4 for the max6675.

The lower-than-default (20Hz) sampling rate makes the ring buffer a bit long at 16 samples, and it should *probably* be reduced, but IMO that's not a show-stopper for this very old and blocker PR.

Issue #890, opened 2016-03-30
the current sense resistor value in motor_driver_control for DRV8711 seems to be hard coded


greyltc:
Is there anything I can put into my config file to tell my smoothieboard what value to use for the current sense resistor when using the motor_driver_control module with a DRV8711?

Is it hard coded to be 50mOhm [here] (https://github.com/Smoothieware/Smoothieware/blob/467c3642c15958c96aa96062feaa7172a079c507/src/modules/utils/motordrivercontrol/drivers/DRV8711/drv8711.h#L142)?

The sense resistor I have is 33mOhm. So maybe as a temporary workaround to this I could scale the `current` setting in my config file by 33/50 = 0.66?

wolfmanjm:
it is hard coded. That should be fixed.

wolfmanjm:
fixed in #897

greyltc:
Thanks!

PR #892, opened 2016-03-31
motor_driver_control: user configurable current sense resistor

greyltc:
This adds a "sense_resistor" configuration key (set in mΩ) to the motor_driver_control module.
This impacts/works with both the TMC26X and the DRV8711 drivers.

The default value I chose here (50mΩ) agrees with what was previously set by the TMC26X and the DRV8711 drivers, so this change should give nobody any unpleasant surprises.

### DIFF
--- a/src/modules/utils/motordrivercontrol/MotorDriverControl.cpp
+++ b/src/modules/utils/motordrivercontrol/MotorDriverControl.cpp
@@ -29,6 +29,7 @@
 
 #define current_checksum               CHECKSUM("current")
 #define max_current_checksum           CHECKSUM("max_current")
+#define sense_resistor_checksum        CHECKSUM("sense_resistor")
 
 #define microsteps_checksum            CHECKSUM("microsteps")
 #define decay_mode_checksum            CHECKSUM("decay_mode")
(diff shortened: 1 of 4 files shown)


greyltc:
I haven't tested this in my setup yet.
It's meant to address https://github.com/Smoothieware/Smoothieware/issues/890.

wolfmanjm:
Sorry this is not the place to put it, the sense resistor setting needs to go in the driver itself not in the motorcontrol, as some may not have a sense resistor.
Also it needs to be a float not an int.

greyltc:
I put it in the MotorDriverControl class because 100% (both) of the drivers I can see in the the repo have the hard coding issue so this gets two birds with one stone.

I decided to store the value as an unsigned int because [the TMC26x already does that](https://github.com/Smoothieware/Smoothieware/blob/edge/src/modules/utils/motordrivercontrol/drivers/TMC26X/TMC26X.h#L427) and it makes more sense than a float to me because I can't think of a time when I'd need to specify a negative current sense resistor value or with more precision than a whole number of mOhm.

wolfmanjm:
Well the thing is you need to consider future additions, not just what is there now.
If you fix this I'll reopen, but otherwise It'll remain closed and I'll add it myself at some point.

greyltc:
What's wrong with a future motor control driver just not using the sense_resistor variable if it doesn't need it?

I'm not understanding the drive behind your desire to keep sense_resistor out of MotorDriverControl. Is this coming from a ram savings point of view? Three new motor drivers will have to be added to the project that don't use external current sense resistors before the necessity of that option becomes a minority, right?

Heck, decay_mode is a config param currently in MotorDriverControl and it's not currently passed to either of the existing drivers!

I find it much more likely that a potential future addition won't have a configurable decay_mode (or one so wildly different that a one size-fits-all config parameter can't be shared) than for it not to have an external current sense resistor.

greyltc:
Assuming the root of the problem you have here is the storage of sense_resistor in MotorDriverControl, I've just made a commit to this branch that undoes that. See my branch for this PR: https://github.com/greyltc/Smoothieware/tree/configurable-sense-resistor

Does that address your concern here?

wolfmanjm:
this is not a storage issue (in this case) it is an architectural issue. C++ is all about encapsulation and data hiding. As such IMO hardware details like sense resistor is also an encapsulation issue and sense resistor is specific to H/W and therefore is encapsulated in the driver.
One advantage of this is you only need to change the driver you can test. You do not have the other driver so you can;t test it, forcing me to have to setup a test environment for that driver which I really do not have the time to do. If it remains encapsulated as I suggest it be, then you only change the driver you can test and leave the other driver alone.
Also the interpretation of the sense resistor value can be different for different drivers, in this case TMC uses Mohms and DRV uses float ohms.
I am afraid I will not accept a patch like this unless it conforms to my view of the architecture.
It really is a matter of perspective, your perspective is not necessarily incorrect, it just differs from mine.

wolfmanjm:
fixed in #897

Issue #894, opened 2016-03-31
max31855: more robust detection of errors


greyltc:
Maybe the issue title is an over reaction, maybe not, but [these lines](https://github.com/Smoothieware/Smoothieware/blob/edge/src/modules/tools/temperaturecontrol/max31855.cpp#L69-L73) especially when taken in conjunction with [these other lines](https://github.com/Smoothieware/Smoothieware/blob/edge/src/modules/tools/temperaturecontrol/max31855.cpp#L99-L104) seem to me to mean that thermocouple temperature measurement errors get silently ignored.

In real life, this means if your thermocouple becomes disconnected/open circuit (a very real possibility when you're heating things that are also moving) when you're by chance in a heating cycle, then your heater will run continuously. That's a fire hazard in my book.

We should do our very best to consider all the potential (temperature measurement) failure modes the firmware can possibly detect and make sure they result in a safe situation if we can help it (like the heaters being shut off).

greyltc:
@hakalan, do you agree with my analysis here? or am I missing something?
When I unplug my thermocouple while using the latest released firmware.bin my temperature readings are frozen to the last value measured, so something clearly needs to change. I hope https://github.com/Smoothieware/Smoothieware/pull/895 fixes it. It will mean the ring buffer will get "poisoned" by an inf value whenever there is a read error. Why and how often were you seeing read errors?

hakalan:
I agree that the error handling seems to be broken and that this is bad. Your test confirms this. I distantly remember doing more work on this which unfortunately never got mature enough for a pull request. 

In my setup there were occasional errors which I blamed on switching of the heater (due to cables running alongside). So some robustness is needed, but several consecutive errors are not to be expected.

Grogyan:
Recently we have fixed a similar bug in Marlin firmware, my diagnostic code was to ignore n over temps to find a solution to the random over temps.
This was fixed by reducing the SPI baud rate only for the MAX31855 from 1/16 to 1/64th of the 16MHz Oscillator, i'd expect a similar approach could be done here. Changing SPI bit SPR0 to SPR1.

In almost all printing areas, a thermocouple is more reliable and consistent over a thermistor, and safer.

At 1/16 Osc this is the limit of the clock for the MAX31855 where the max transition time for the CS pin seems to be around 200ns

hakalan:
For me, the issue was never spi comms, but rather that the max31855 signalled various measurement errors via its status bits. Of course it makes sense to  robust against both.

[...]

PR #895, opened 2016-03-31
stop ignoring failed temperature readings

greyltc:
For https://github.com/Smoothieware/Smoothieware/issues/894 .
Maybe this breaks max31855 in practice, but I don't really care. It makes it safer.

### DIFF
--- a/src/modules/tools/temperaturecontrol/max31855.cpp
+++ b/src/modules/tools/temperaturecontrol/max31855.cpp
@@ -66,12 +66,6 @@ float Max31855::get_temperature()
 
 	float temp = read_temp();
 
-	// Discard occasional errors...
-	if(!isinf(temp))
-	{
-		readings.push_back(temp);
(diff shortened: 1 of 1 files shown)


wolfmanjm:
this is not the way to do it. I suggest you allow a bad reading and if you get 10 bad readings in a row, then return inf which will trigger the failsafe.

greyltc:
Yeah, I'd say this logic should be one layer higher though, say in the TemperatureControl class with the individual drivers being kept simple and just reporting reading failures (say by sending inf).

dcousens:
>  the individual drivers being kept simple and just reporting reading failures

Agreed,  this logic is higher than the driver.
The driver should be as stateless as possible IMHO.

wolfmanjm:
no this test is done here, it just needs to allow for a few failed readings and only barf when there are like a buffer full of bad readings.

PR #897, opened 2016-03-31
refactor to allow for driver specific config settings

wolfmanjm:
drv8711 has gain and sense_resistor
 TMC has sense_resistor


src/modules/utils/motordrivercontrol/MotorDriverControl.h
```diff
@@ -28,7 +28,7 @@ class MotorDriverControl : public Module {
 
     private:
         bool config_module(uint16_t cs);
-        void initialize_chip();
+        void initialize_chip(uint16_t cs);
```
dcousens: what is `cs`?

PR #900, opened 2016-04-01
BuildShell usable from Eclipse

MoonCactus:
Running on Linux, I found no way to use BuildShell for building from Eclipse with an external builder.
This modification is compatible with existing behavior: When no arguments is given to BuildShell, the behavior is left unchanged. Else, additional arguments are executed directly from BuildShell instead of the existing "exec bash".
This way Eclipse can run "./BuildShell make" or "./BuildShell make clean".

Admittedly, the name "BuildShell" becomes misleading. And I did not check Mac & Windows platforms.

### DIFF
--- a/linux_install
+++ b/linux_install
@@ -84,7 +84,11 @@ echo "SCRIPT_PATH=\${SCRIPT_PATH%/*}">>$BUILDSHELL_CMD
 echo "cd \$SCRIPT_PATH">>$BUILDSHELL_CMD
 echo "SCRIPT_PATH=\$PWD">>$BUILDSHELL_CMD
 echo "export PATH=\$SCRIPT_PATH/gcc-arm-none-eabi/bin:\$PATH">>$BUILDSHELL_CMD
-echo "exec bash">>$BUILDSHELL_CMD
+echo "if (( \$# )); then">>$BUILDSHELL_CMD
+echo "  \$@ # so it supports ./BuildShell make, e.g. for calling from an Eclipse builder">>$BUILDSHELL_CMD
+echo "else">>$BUILDSHELL_CMD
+echo "  exec bash">>$BUILDSHELL_CMD
+echo "fi">>$BUILDSHELL_CMD
 chmod +x $BUILDSHELL_CMD
 
 echo Cleaning up intermediate files...


wolfmanjm:
Eclipse is not supported. So feel free to keep this patch in your fork. Thank you.

wolfmanjm:
I would prefer to see a separate script for use with Eclipse. Not changing the way the current BuildShell stuff work saves testing.

arthurwolf:
I'd love to see eclipse officially supported by Smoothie, as it's something
a lot of folks ask for.
But it'd need to be done in a way that's simple ( this :
http://smoothieware.org/eclipse never works for anyone ).
And it'd need to stay in it's own branch for a while for testing, and it'd
need to be done in a way that doesn't impact the "normal"/"traditional" way
Smoothie works.
A separate script sounds like the right way to go.

Thanks for looking into this Jeremie :)

On Fri, Apr 1, 2016 at 8:10 PM, Jim Morris <notifications@github.com> wrote:

> I would prefer to see a separate script for use with Eclipse. Not changing
> the way the current BuildShell stuff work saves testing.
>
> —
> You are receiving this because you are subscribed to this thread.
> <https://github.com/Smoothieware/Smoothieware/pull/900#issuecomment-204498728>
>



-- 
Courage et bonne humeur.

MoonCactus:
Thanks for updating your comment wolfman ... man, you are quite harsh (I sincerely hope you are not telling me you allow me to keep the patch in my own branch?). Sorry for the "testing" branch though (I am just a total noob with github, not with programming though). I like it how you take care about keeping a stable build and continuous integration.
Now as I said, I did take care not to have any single side effect beyond allowing more IDE (hence more programmers) to contribute transparently, without even an additional file in the project. Looks like there are some according to Arthur.
Twice in a row and no ego: the "risk" is just that I may simply not bother to try and contribute at all from my fork. Especially if it is less work for everyone ;)
BTW: yes, I am currently working & compiling with Eclipse with your own makefile. I will have a look to the wiki if it fails for other people (or even post in my own blog, as I should with GRBL and Marlin under Eclipse). It is really nailed in three simple steps, not sure an additional script file is worth the burden.

MoonCactus:
(update: not debbugging yet, though!)

wolfmanjm:
BuildShell is only supposed to be used with command line makefile usage. 
If Eclipse were setup properly it does not need to use BuildShell at all.
BuildShell is supposed to create a shell for commandline usage NOT eclipse.
Also BuildShell comes from gcc4mbed so any changes to it should be submitted upstream.

MoonCactus:
OK np. Sure, purists want it all 100% configured via and per the deepest guts of eclipse (so they get screwed with updates...). As for me I like it better when it works with existing portable setups like good old command line, makefiles and gdb server. I cannot + will not help there, so it's closed also for me :)

Issue #886, opened 2016-03-27
ZGrid for PCBs


FreezeS:
Hi guys,

I have been trying to use the inbuilt zgrid probing for PCB milling but in it's current status I haven't been able to.

I have done the following:
- placed the etching bit 2 mm above the left bottom corner of the pcb and faked a homing (G28.3 X0 Y0 Z0) since my shapeoko doesn't have endstops.
- probed until it hit the pcb and set the offset (G38.3 Z-10; G92 Z0)
- lifted the bit 1 mm above the pcb and ran the scanning procedure (G32)

The result is this:
```
;Bed Level settings:
X0 A0.00 B-1.20 C-1.22 D-1.25 E-1.28
X1 A-1.05 B-1.11 C-1.14 D-1.17 E-1.20
X2 A-0.99 B-1.06 C-1.10 D-1.13 E-1.16
X3 A-0.94 B-1.01 C-1.05 D-1.09 E-1.12
X4 A-0.93 B-1.00 C-1.05 D-1.08 E-1.11
```

Good news is that it keeps the home position but the zgrid offsets are all bad.
Shouldn't the values be around zero? Instead they are around 1 (except for X0 A)

My config is:
```
zprobe.enable                true               # set to true to enable a zprobe
zprobe.probe_pin             1.28!^             # pin probe is attached to if NC remove the !, Azteeg X5 this is 1.29
zprobe.slow_feedrate         5                  # mm/sec probe feed rate
#zprobe.debounce_count       100                # set if noisy
zprobe.fast_feedrate         50                # move feedrate
#zprobe.probe_height          2                  # how much above bed to start probe NB only needed for G32 on delta


leveling-strategy.ZGrid-leveling.enable         true
leveling-strategy.ZGrid-leveling.bed_x           50
leveling-strategy.ZGrid-leveling.bed_y           50
leveling-strategy.ZGrid-leveling.bed_z           10
leveling-strategy.ZGrid-leveling.rows           5          # X divisions (Default 5)
leveling-strategy.ZGrid-leveling.cols           5         # Y divisions (Default 5)
leveling-strategy.ZGrid-leveling.probe_offsets  0,0,0
leveling-strategy.ZGrid-leveling.slow_feedrate  50
leveling-strategy.ZGrid-leveling.wait_for_probe  true
leveling-strategy.ZGrid-leveling.home_before_probe  nohome;    #  nohome homexy homexyz (default)
```

[...]

PR #910, opened 2016-04-16
Z-offset repairs

qharley:
Added M306.1 to be able to jog Z calibration during printing
Repaired reset_axis_position in robot to take compensation transform
into account.

This fixes M306 without homing: ZGrid for PCBs (#886)

### DIFF
--- a/src/modules/robot/Robot.cpp
+++ b/src/modules/robot/Robot.cpp
@@ -763,10 +763,33 @@ void Robot::distance_in_gcode_is_known(Gcode * gcode)
 // so in those cases the final position is compensated.
 void Robot::reset_axis_position(float x, float y, float z)
 {
+    //float target[3];
+
+
+
     // these are set to the same as compensation was not used to get to the current position
     last_machine_position[X_AXIS]= last_milestone[X_AXIS] = x;
     last_machine_position[Y_AXIS]= last_milestone[Y_AXIS] = y;
     last_machine_position[Z_AXIS]= last_milestone[Z_AXIS] = z;
+
+    // last machine position should actually include compensation if applicable to prevent transform jumping...
+    // check function pointer and call if set to transform the target to compensate for bed
+    if(compensationTransform) {
+        // some compensation strategies can transform XYZ, some just change Z
+        compensationTransform(last_machine_position);
+    }
+/*
+    // unity transform by default
+    memcpy(transformed_target, target, sizeof(transformed_target));
+
+    // check function pointer and call if set to transform the target to compensate for bed
+    if(compensationTransform) {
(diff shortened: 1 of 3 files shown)


wolfmanjm:
The compensation fixes you have made are incorrect. Homing does not apply compensation so there is nothing to compensate for.
I am rejecting this. You may need to do a PR with only the zgrid stuff.

qharley:
If you Do the following:

G28
G0 X0 Y0 Z10 F1000
M306 Z10
G0 X1

The Z axis moves by the compesation offset on the last command. This is not
correct

qharley:
If you have no compensation there is no adverse effect. Without this, you
can close the pcb one as well. Ther is no way to fix it then

wolfmanjm:
In my opinion that is correct. This has been discussed in other issues ad nauseam :)

Basically homing sets the actuator position it does no coordinated moves via robot, so the actuator position is being set to the endstop position for the actuator. This is especially true of the rotatary delta where actuator position is degrees and FK is used to find the cartesian coordinates.

In your example G28 homes the actuators and sets the XYZ based on the actuator positions, no compensation has been applied, so last machine position is the same as last milestone.

When you set the current homing offset via M306/M206 it sets the homing offset which only takes effect on the next home.. (it is a homing offset and not expected to take immediate effect) so it is basically a NOP in your example.
then when you do G0 X1 it will apply the compensation for the position at at X1 Y0 as expected.

if you did 

````
G28
G0 X0 Y0 Z10 F1000
M306 Z10
G28
G0 X1 Y0 Z10
```

you would get exactly what you expect.

FYI to set dynamic z offsets you would use G92 or G54 (G10 L2 P1 xxxx) you would not use homing offsets.

wolfmanjm:
FWIW we need an inverse compensation transform for other situations, but not this one.

qharley:
I agree that homing would fix the issue, but in the case of the pcb
engraver there is no home. In that instance there is no way to correct the
position.

Homing after doing the m306 as I modified it would have the exact same
effect as if I did not, except that it is usable without homing as well.

The 306.1 function is very useful, allowing you to adjust the z during
printing of the first layer, but unfortunately is would not work either if
the function was not modified.

No problem. At least my customers will have the advantage of the function.

Peace,
Q

wolfmanjm:
In the case there is no physical homing I added M28.3 to allow manual homing, and adding in homing offsets here makes no sense.

There is a reason linuxcnc and grbl do not allow you to do anything until the axis are homed. In the case of grbl not being able to home is an issue, whereas linuxcnc allows you to manually home, as does smoothie now.

FreezeS:
So, does the z-grid work properly now without homing switches in the main firmware? If not, is there a build of @qharley's branch?

PR #912, opened 2016-04-17
set "Access-Control-Allow-Origin" in all headers

lautr3k:


lautr3k:
Included in new PR #913

PR #913, opened 2016-04-17
Allow remote file upload (CORS)

lautr3k:
Implement preflighted requests (OPTIONS) to allow remote files upload (CORS).

- https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Preflighted_requests
- https://www.w3.org/TR/cors/


wolfmanjm:
Can you please explain what this is used for, how it is used, and what happens without it?

The code looks fine, however I am not understanding what it is used for, and I am concerned about the significant addition in Flash storage.

Thank you.

lautr3k:
When a remote POST request is made to send a file to the server, the JavaScript XMLHttpRequest() object change the request type to 'OPTIONS' to ensure the request is secure and wait for some headers to validate the request. This PR handle the OPTIONS request and send apropriate header to allow remot file upload via XMLHttpRequest(). Otherwise a client that does not reside on the board can not upload file.

dcousens:
CORS should only be necessary if your origin is different?  Was this for API access?

lautr3k:
@dcousens Yes for an API access i am working on : http://lautr3k.github.io/Smoothie-Happy/
If the client is not on the board (hosted on sd card) it is not on the same ip/url.

PR #919, opened 2016-04-27
Add option to restrict Play file selector to a subdir.

pelrun:
In bulletproofing and streamlining the workflow of my hackerspace's laser cutter that is using a smoothieboard, I want to hide the details of the filesystem on the sd card and only show users the directory that our gcode generator uploads to.

This patch adds a "panel.root_path" configuration option that is used as a virtual root by the Play menu. It opens the configured path (or THEKERNEL->current_dir if it's underneath that path) when the Play menu is selected, and exits when you try to go to it's parent, as if the directory was actually the root of the filesystem. Otherwise you can still navigate into subdirectories and back out again as normal, and if the option is unspecified it defaults to the normal root dir "/".

### DIFF
--- a/src/modules/utils/panel/Panel.cpp
+++ b/src/modules/utils/panel/Panel.cpp
@@ -65,6 +65,8 @@
 #define bed_temp_checksum    CHECKSUM("bed_temperature")
 #define panel_display_message_checksum CHECKSUM("display_message")
 
+#define root_path_checksum         CHECKSUM("root_path")
+
 Panel* Panel::instance= nullptr;
 
 #define MENU_MODE                  0
@@ -179,6 +181,7 @@ void Panel::on_module_loaded()
 //    this->back_button.set_longpress_delay(longpress_delay);
 //    this->pause_button.set_longpress_delay(longpress_delay);
 
+    this->root_path = THEKERNEL->config->value( panel_checksum, root_path_checksum )->by_default("/")->as_string();
 
     THEKERNEL->slow_ticker->attach( 50,  this, &Panel::button_tick );
     if(lcd->encoderReturnsDelta()) {
--- a/src/modules/utils/panel/Panel.h
+++ b/src/modules/utils/panel/Panel.h
@@ -88,6 +88,9 @@ class Panel : public Module {
         LcdBase* lcd;
         PanelScreen* custom_screen;
 
+        // path prefix to restrict Play panel to
+        std::string root_path;
+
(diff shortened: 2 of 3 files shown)


pelrun:
This explicitly only affects the Play menu UI; the full sd card is still accessible through every other method.

arthurwolf:
I think this is very useful.

Jim : if ram use is a big concern here, maybe we can reduce this to just being a boolean option, and force the folder to be something specific ( like /sd/gcode/ ) ?

JesperKrogPoulsen:
I agree that this is useful, but restricting to something on /sd will prevent people using external sd cards from using it and the other way around.

wolfmanjm:
I do not like this PR. For a very limited usefulness it uses too much memory. the string needs to be replaced somehow so this uses virtually no memory when not in use (which is 99.9% of the time).

ConfigSamples/Smoothieboard/config
```diff
@@ -298,6 +298,8 @@ panel.gamma_jog_feedrate                     200               # z jogging feedr
 panel.hotend_temperature                     185               # temp to set hotend when preheat is selected
 panel.bed_temperature                        60                # temp to set bed when preheat is selected
 
+#panel.root_path                             /sd/gcode         # Only show this path and below in the Play menu
```
wolfmanjm: do not put this in the config it should sijmply be documented in the wiki or maybe in the snippets.

pelrun:
It's really easy as a dev to ignore or criticise usability improvements because you're familiar with the system and can dodge all the sharp edges. It's still an issue that should be addressed.

I explicitly did the simplest possible implementation in part to trigger some discussion. Replacing it with an enum would be equally satisfactory, and would avoid the issue @JesperKrogPoulsen raised about users of external sd cards.

You say you're happy with "virtually" no RAM impact. What is the size of a string set to "/" these days? Obviously it's a bit heavier than a cstr, but I haven't actually done the profiling yet.

arthurwolf:
@wolfmanjm we might be getting a ton of flash/ram saving soon on the panel side with the file-based panel/menu structure thingie I mentionned and somebody is looking into. So if this can be reduced to using only a few bytes, I really think we can get it in. Especially since it'll just use no ram at all later on.

arthurwolf:
@pelrun Why close ? This was marked as "needs testing". 
It's possible this might not be something we can get into v1, but I'd very like to have this in v2, so we don't want to forget about this.

PR #921, opened 2016-05-04
Feature/cnc pendant panel

wolfmanjm:
New CNC based Panel screens, Only compiled with ```make CNC=1```
(Uses no extra resources if not selected).
Introduces a watch screen with WCS and MCS and a CNC like pendant view.
Adds MPG (Manual Pulse Generation) as a jog option. (requires a High PPS encoder).

Added test console commands, to test basic movements, useful for tuning speed and acceleration.
Try to make kill maintain a correct position (still needs work).
Refactored makefiles to allow a ```src/default_excludes.mk``` which is not in git, but makes the default build set make options and select/deselect modules.
eg for a CNC build....
```
export CNC=1
export NONETWORK=1
export EXCLUDE_MODULES = tools/laser tools/filamentdetector tools/scaracal tools/temperaturecontrol tools/extruder
```


dcousens:
You could add a `src/default_excludes.mk.example`

PR #929, opened 2016-05-23
Merging current edge into runaway to resume work on runaway

arthurwolf:


dcousens:
:+1:

wolfmanjm:
Needs to reset the timer and state when a new temperature is set,otherwise it could trigger the halt if it was running then the temp was increased beyond the threshold.

Optionally make it more noise tolerant by making sure that the test fails two consecutive times before triggering halt.

arthurwolf:
Also, the timeout for heating up is too slow as a uint8_t apparently, some users have beds that heat in more time than that.

PR #932, opened 2016-05-28
Added no caching metatag, to compat chromes agressive caching

dumle29:


wolfmanjm:
We deliberatey force caching, we do not want the smoothie to have to refresh the page every time something changes as the smoothie virtually has to halt to send a page. Just force refresh on your browser.

dumle29:
That's not really an option, as I can't even get to the page. It's a bit odd really. Say I'm on http://google.com. I enter the local IP in the address bar, and hit enter. Then the page just refreshes, and stays on google. I have to open a fresh incognito window in order to access the smoothieboard between power cycles.

Because of that, I can't force refresh the page, as I can't even get to it :/

wolfmanjm:
well not allowing the page to cache is not an option for everyone else. So not sure what to suggest. BTW I don;t have that issue with the chromium browser

PR #941, opened 2016-06-09
how to file issues document

wolfmanjm:


dcousens:
My only suggestion might be to also link to this somewhere in the README :+1: , but otherwise thanks for adding this.

PR #946, opened 2016-06-17
Feature/acceleration per tick

wolfmanjm:
Complete rewrite of the tick generation code, the Conveyor, and some of the planner.
_Currently only works for a CNC build. (A lot more to come)._

Currently to build the following modules need to be disabled...

- tools/laser
- tools/filamentdetector
- tools/scaracal
- tools/temperaturecontrol
- tools/extruder
- tools/zprobe

**What changed**
1. Adds acceleration per tick code from Arthur Wolf
2. Conveyor handles queue differently
3. removed on_gcode_execute and storage of gcode in the block
4. Removed Stepper.cpp
5. added solo_move (TBD) to replace old direct stepper motor commands (may be removed)
6. Updated Endstops to do homing using planned moves, and new way to detect endstops

This fixes several issues with step generation, and is a lot smoother.


arthurwolf:
I can't test this but just wanted to say it's really great, and I strongly encourage anybody with time to do so, to test this, we all really really want this to be merged into smoothie, this is a great step forward and will allow for many great things to come.

wolfmanjm:
this is being replaced by the feature/5-axis branch which is based on this but adds 5 axis (XYZ + 2E) and has a new extruder module.

PR #948, opened 2016-06-22
Feature/5 axis

wolfmanjm:
New motion control and stepticker as before.
adds 5 axis (XYZ + 2E) and has a new extruder module.
laser module must be disabled to compile
add
```
export EXCLUDE_MODULES = tools/laser
```

to the src/default_excludes.mk file


wolfmanjm:
this is now in edge-unstable branch

PR #951, opened 2016-06-22
Implementation of software movement limits/endstops.

sareph:
Feature that should be present from the beginning, IMO. If enabled, will limit movements within specified XYZ positions (after homing is done). Just to prevent breaking something.
I'm not entirely sure, if it's done right - after all, first time when I saw smoothie code was few hours ago, so my knowledge about it's internal magic is limited.

By the way, did You know, that config params checksums can collide? And that can lead to very strange debug problems. Two hours of fighting with enabled, but magically self unloading endstops module.

### DIFF
--- a/src/modules/robot/Robot.cpp
+++ b/src/modules/robot/Robot.cpp
@@ -82,6 +82,15 @@ using std::string;
 #define  beta_checksum                       CHECKSUM("beta")
 #define  gamma_checksum                      CHECKSUM("gamma")
 
+#define  soft_limits_enable_checksum         CHECKSUM("soft_limits_enable")
+
+#define  soft_limits_x_min_checksum          CHECKSUM("soft_limits_x_min")
+#define  soft_limits_y_min_checksum          CHECKSUM("soft_limits_y_min")
+#define  soft_limits_z_min_checksum          CHECKSUM("soft_limits_z_min")
+#define  soft_limits_x_max_checksum          CHECKSUM("soft_limits_x_max")
+#define  soft_limits_y_max_checksum          CHECKSUM("soft_limits_y_max")
+#define  soft_limits_z_max_checksum          CHECKSUM("soft_limits_z_max")
+
 #define NEXT_ACTION_DEFAULT 0
 #define NEXT_ACTION_DWELL 1
 #define NEXT_ACTION_GO_HOME 2
@@ -124,6 +133,9 @@ Robot::Robot()
     this->g92_offset = wcs_t(0.0F, 0.0F, 0.0F);
     this->next_command_is_MCS = false;
     this->disable_segmentation= false;
+    this->homing_done[X_AXIS]= false;
+    this->homing_done[Y_AXIS]= false;
+    this->homing_done[Z_AXIS]= false;
 }
 
 //Called when the module has just been loaded
@@ -192,6 +204,16 @@ void Robot::load_config()
     this->segment_z_moves     = THEKERNEL->config->value(segment_z_moves_checksum     )->by_default(true)->as_bool();
     this->save_g92            = THEKERNEL->config->value(save_g92_checksum            )->by_default(false)->as_bool();
 
+    this->soft_limits_enable       = THEKERNEL->config->value(soft_limits_enable_checksum )->by_default(false)->as_bool();
+
+    this->soft_limits_min[X_AXIS]  = THEKERNEL->config->value(soft_limits_x_min_checksum    )->by_default(  0.0F)->as_number();
+    this->soft_limits_min[Y_AXIS]  = THEKERNEL->config->value(soft_limits_y_min_checksum    )->by_default(  0.0F)->as_number();
+    this->soft_limits_min[Z_AXIS]  = THEKERNEL->config->value(soft_limits_z_min_checksum    )->by_default(  0.0F)->as_number();
+
+    this->soft_limits_max[X_AXIS]  = THEKERNEL->config->value(soft_limits_x_max_checksum    )->by_default(200.0F)->as_number();
+    this->soft_limits_max[Y_AXIS]  = THEKERNEL->config->value(soft_limits_y_max_checksum    )->by_default(200.0F)->as_number();
+    this->soft_limits_max[Z_AXIS]  = THEKERNEL->config->value(soft_limits_z_max_checksum    )->by_default(200.0F)->as_number();
+            
     // Make our 3 StepperMotors
     uint16_t const checksums[][5] = {
         ACTUATOR_CHECKSUMS("alpha"),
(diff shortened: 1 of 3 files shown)


wolfmanjm:
Sorry this will not work in the new code. 

Also there is a document in process to handle soft endstops. IMO this is a very naive method and not appropriate and wil break moist peoples systems.

If this is what yo uwant I suggest you keep it in your own fork. But it is not appropriate for main line.

Thank you

wolfmanjm:
Here is the document if you wish to contribute... 

https://docs.google.com/document/d/1U6nzx1boqF-J2GGPWF4yIaaVib0JNodVWSKBfwiyp_M/edit

it explains why your approach is not acceptable. Thank you.

arthurwolf:
Hello !

Thanks for trying to implement this !

We are currently ( well, wolfmanjm is ) working on a refactor of the Smoothie code, which means this kind of feature won't work with the new codebase, so we need to wait until that refactor is done before implementing anything like this.

I wrote a document about implementing soft endstops here : https://docs.google.com/document/d/1U6nzx1boqF-J2GGPWF4yIaaVib0JNodVWSKBfwiyp_M/edit?usp=sharing
I recommend you read it, it should give you a better idea of what needs to be done, what the community wants, etc.

Tell us what you think.

Cheers.

qharley:
The soft endstops I impemented for Morgan's firmware lives in Robot.  It
works well for us, but does not have the flexibility as described in
Arthur's document yet.

I am willing to spend some time to implement it on the refactor.  Our users
are used to the machine protecting itself against any SU Errors, and we
will implement it on the new codebase if only for our systems.

arthurwolf:
Well I'd be glad if somebody implemented it yes. Thanks for offering. But
it's going to have to wait for Jim to be done with the refactor.

On Wed, Jun 22, 2016 at 10:53 PM, Quentin Harley <notifications@github.com>
wrote:

> The soft endstops I impemented for Morgan's firmware lives in Robot. It
> works well for us, but does not have the flexibility as described in
> Arthur's document yet.
>
> I am willing to spend some time to implement it on the refactor. Our users
> are used to the machine protecting itself against any SU Errors, and we
> will implement it on the new codebase if only for our systems.
>
>
>
> On Wed, Jun 22, 2016 at 10:44 PM Arthur Wolf <notifications@github.com>
> wrote:
>
> > Hello !
> >
> > Thanks for trying to implement this !
> >
> > We are currently ( well, wolfmanjm is ) working on a refactor of the
> > Smoothie code, which means this kind of feature won't work with the new
> > codebase, so we need to wait until that refactor is done before
> > implementing anything like this.
> >
> > I wrote a document about implementing soft endstops here :
> >
> https://docs.google.com/document/d/1U6nzx1boqF-J2GGPWF4yIaaVib0JNodVWSKBfwiyp_M/edit?usp=sharing
> > I recommend you read it, it should give you a better idea of what needs
> to
> > be done, what the community wants, etc.
> >
> > Tell us what you think.
> >
> > Cheers.
> >
> > —
> > You are receiving this because you are subscribed to this thread.
> > Reply to this email directly, view it on GitHub
> > <
> https://github.com/Smoothieware/Smoothieware/pull/951#issuecomment-227870820
> >,
> > or mute the thread
> > <
> https://github.com/notifications/unsubscribe/AA0PWCww2LiHJikNcFaQaASZUWnIWIy9ks5qOZ7CgaJpZM4I8MaJ
> >
> > .
> >
>
> —
> Reply to this email directly, view it on GitHub
> <https://github.com/Smoothieware/Smoothieware/pull/951#issuecomment-227873050>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAGpFTHIrHw84eV9lEuZ9aRRpZ56hrH_ks5qOaDSgaJpZM4I8MaJ>
> .
>



-- 
Courage et bonne humeur.

qharley:
Sure.  We are patient... ish...
;-)

sareph:
> The question is : in which of those case do we trigger soft endstops,

IMO if it's enabled, all movements, except for homing. For reason I can't understand, Smoothie ignores even hardware endstops for movements.  
Software or hardware - threat them the same, limits are configured for something, and it's not that they are hardcoded, config can be changed if it's not correct. If user configured safe work area, then don't leave it. Well, maybe with exception for input from panel. But even then, warning should exist like "you are about to exit safe operating area. Continue?" which will disable software endstops until next homing.

> [enabling endstops only for selected axes]

Oh, this one I like.

> What do we do when soft endstops are triggered ?

Well, You are currently doing nothing, even for hardware ones (insert lot's of complaining here). 
The same what other firmwares do. Stop movement toward that endstop for sure, and/or if configured - enter halt state.

arthurwolf:
« Software or hardware - threat them the same »

It's great that that's what you want, but having done Smoothie for years,
and having had many folks asking for soft endstops ( and hard before that
), people want *very* varied things, not just what you want.
What you want would just sound horrible to some other person. That's why it
absolutely needs to be configurable and offer as large a variety of options
as possible ( thus what is explained in the specification, and the proposed
implementation ).

I'm not sure what you mean about hardware endstops "doing nothing" ?

On Wed, Jun 22, 2016 at 11:32 PM, Tomek Nagisa <notifications@github.com>
wrote:

> The question is : in which of those case do we trigger soft endstops,
>
> IMO if it's enabled, all movements, except for homing. For reason I can't
> understand, Smoothie ignores even hardware endstops for movements.
>
> Software or hardware - threat them the same, limits are configured for
> something, and it's not that they are hardcoded, config can be changed if
> it's not correct. If user configured safe work area, then don't leave it.
> Well, maybe with exception for input from panel. But even then, warning
> should exist like "you are about to exit safe operating area. Continue?"
> which will disable software endstops until next homing.
>
> [enabling endstops only for selected axes]
>
> Oh, this one I like.
>
> What do we do when soft endstops are triggered ?
>
> Well, You are currently doing nothing, even for hardware ones (insert
> lot's of complaining here).
> The same what other firmwares do. Stop movement toward that endstop for
> sure, and/or if configured - enter halt state.
>
> —
> Reply to this email directly, view it on GitHub
> <https://github.com/Smoothieware/Smoothieware/pull/951#issuecomment-227883547>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAGpFXNRRM22sXsVO13bebfEaSlJ5Qolks5qOaoIgaJpZM4I8MaJ>
> .
>



-- 
Courage et bonne humeur.

wolfmanjm:
You are just plain wrong about hardware endstops. If you were to actually enable them (see the wiki) then the system will halt if the endstop is triggered during any move.
I suggest you look at the endstop.cpp code and read the wiki on endstops :) before you complain too much.

Additionally you really need to think and understand the implications of truncating a move on a 3D printer. Consider a long move that is truncated early, however it will still extrude the same amount as the long move, this would make a nasty mess at best, and at worst totally destroy the hotend. Plus of course the print will be totally destroyed. Which i why IMO the only thing to do on a limit is to halt safely.

sareph:
> people want *very* varied things, not just what you want.

Oh I know. But, what I want can be found in... 99%? Of other firmware implementations, and I didn't see much complaining about how it works there. So, implement it as found there, and wait for user feedback? And then do whatever they want, if it's worth the time.

> I'm not sure what you mean about hardware endstops "doing nothing" ?
> You are just plain wrong about hardware endstops.  If you were to actually enable them

No, not really, or maybe partially. If I enable *_limit_enable - yes, machine will enter halt state. But I don't want it to enter halt state, it's irritating overkill, I just want it to stop moving in the endstop direction. if *_limit_enable is not configured then it will just smash into endstop.
Tell me, how could I guess, that if this is disabled (but endstops are enabled), smoothie will simply ignore and break them? I not sure if anyone else had similar idea, to ignore installed endstop.
IMO - if endstop is configured, smoothie should **always** stop movement in that direction if it was activated. And, but only if configured to do so, enter complete halt state.

> Additionally you really need to think and understand the implications of truncating a move...

I have other machine with repetier installed there, and it just do that (although I'm not sure if it halts extruder). Implications in this case are - for very large object brim gets truncated, but who cares. 

It's unlikely that slicer will generate such a long line, so it's most likely that will be transmission error, good we have checksums. In all other cases, hotend will start to move in any other direction soon, just to continue print. And besides, user should be always around to supervise printer.

arthurwolf:
Hey.

In Smoothie, "endstops" are the sensors at the end of each axis that can be used for homing, or optionally for the limit switches functionality. The limit switches functionality, when enabled, makes Smoothie enter HALT state whenever these switches are hit.

I think the documentation is pretty clear on this, so when you ask "how could I guess", I really want to point you at the documentation.

Thanks for your attempt at implementing soft endstops. As we mentionned, this can not happen before the current refactors are done, and then, we would like it to happen in a specific manner ( which is quite different from what you implemented ).

If you want to help with implementing this, when the time comes, you are welcome and I will try to help you as best I can.

Cheers.

Issue #952, opened 2016-06-24
G20 causes final move in home to be in inches not mm.


Zaaphod:
I have smoothie configured to home the way I want it to, with the retract set to 5mm. When it's done homing it moves all the axis 5mm off the switches.   I then switch to inches with G20.  If I home it while I'm still in G20 It does all the movements correctly, except the final move where all axis move off the switches.  It's moving them all 5 inches not 5mm.   The first retract between fast home and slow home is correct.. 5mm, it's the final move that moves all axis together that it's using the wrong value.  The home procedure should be the same weather you are in G20 or G21.  The retract distance is specified as mm in the config file, not what ever units you are set for... it's the retract distance in mm, always regardless of what units you are working in.  It is doing this in both the current edge version and new test firmware 3

PR #953, opened 2016-06-24
Fix Homing when in G20

Zaaphod:
This is a bug fix when homing when in G20  mode.  It was making the
final move in inches instead of mm.  Homing should always bring you to
the same exact point whether you are in G20 or G21.  See
https://github.com/Smoothieware/Smoothieware/issues/952

### DIFF
--- a/src/modules/tools/endstops/Endstops.cpp
+++ b/src/modules/tools/endstops/Endstops.cpp
@@ -345,6 +345,7 @@ void Endstops::back_off_home(char axes_to_move)
         append_parameters(gcode_buf, params, sizeof(gcode_buf));
         Gcode gc(gcode_buf, &(StreamOutput::NullStream));
         THEKERNEL->robot->push_state();
+		THEKERNEL->robot->inch_mode = false;     // needs to be in mm
         THEKERNEL->robot->absolute_mode = false; // needs to be relative mode
         THEKERNEL->robot->on_gcode_received(&gc); // send to robot directly
         // Wait for above to finish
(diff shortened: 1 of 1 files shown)


wolfmanjm:
Please follow the contribution guidelines... you have whitespace issues in this PR the indentations needs to be spaces not tabs.

src/modules/tools/endstops/Endstops.cpp
```diff
@@ -367,8 +368,8 @@ void Endstops::move_to_origin(char axes_to_move)
     // Move to center using a regular move, use slower of X and Y fast rate
     float rate = std::min(this->fast_rates[0], this->fast_rates[1]) * 60.0F;
     char buf[32];
-    snprintf(buf, sizeof(buf), "G53 G0 X0 Y0 F%1.4f", rate); // must use machine coordinates in case G92 or WCS is in effect
     THEKERNEL->robot->push_state();
+	snprintf(buf, sizeof(buf), "G21 G53 G0 X0 Y0 F%1.4f", rate); // must use mm units & machine coordinates in case G92 or WCS is in effect
```
wolfmanjm: I believe G53 mist be the first thing on the line, so this will break the G53 stuff AFAIK, and I don't have time to check it at the moment. as state is pushed, yo can just set inch_mode as in the other routine.

wolfmanjm: Does this really need to be in inch mode? isn't X0Y0 the same regardless of mm or inch? 0 is zero right?

wolfmanjm:
fixes issue #952

src/modules/tools/endstops/Endstops.cpp
Zaaphod: The feedrates when in G20 mode are inches per minute, so this move was happening at a different speed when you were in G20 mode than when in G21.  This move to origin is happening at what appears to be the fast home speed, but I think it should be at the default_seek_rate.  However this is a different issue and I'm not sure it should be changed, as many people may be depending on it moving at the speed it currently is.

PR #954, opened 2016-06-27
Home sequence adjustemnt

Zaaphod:
Retract fast only if not delta
Backoff Use the slowest of slow rates for delta
Backoff use the slowest of fast rates if not delta
Slow seek uses slowest of slow rates
move to origin at whatever G0 feedrate is set at as this should always
be acceptable.  If G0 was never set with an F parameter then
default_seek_rate is used
back off endstops first when moving to origin in case all axis do not
need to move, it's never a good idea to have any axis barely on the edge
of a limit trigger as machine vibrations could false trip it.

### DIFF
--- a/src/modules/tools/endstops/Endstops.cpp
+++ b/src/modules/tools/endstops/Endstops.cpp
@@ -350,8 +350,14 @@ void Endstops::back_off_home(std::bitset<3> axis)
     if(!params.empty()) {
         // Move off of the endstop using a regular relative move
         params.insert(params.begin(), {'G', 0});
-        // use X slow rate to move, Z should have a max speed set anyway
-        params.push_back({'F', this->slow_rates[X_AXIS] * 60.0F});
+        if(is_delta) {
+            // use slow rate on delta, fast is too violent
+            // use X slow rate to move, Z should have a max speed set anyway
+            params.push_back({'F', this->slow_rates[X_AXIS] * 60.0F});
+        } else {
+            // use fast rate of slowest axis for backoff
+            params.push_back({'F', std::min(fast_rates[Z_AXIS],std::min(fast_rates[X_AXIS],fast_rates[Y_AXIS])) * 60.0F});
+        }
         char gcode_buf[64];
         append_parameters(gcode_buf, params, sizeof(gcode_buf));
         Gcode gc(gcode_buf, &(StreamOutput::NullStream));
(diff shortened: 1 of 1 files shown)


Zaaphod:
I tested this thoroughly on my router and my milling machine, but it needs to be tested on a delta

wolfmanjm:
sorry but i simply think this is inappropriate for general use. I suggest you keep this in your fork.

src/modules/tools/endstops/Endstops.cpp
```diff
@@ -376,12 +382,11 @@ void Endstops::move_to_origin(std::bitset<3> axis)
...
 
     this->status = MOVE_TO_ORIGIN;
-    // Move to center using a regular move, use slower of X and Y fast rate
-    float rate = std::min(this->fast_rates[0], this->fast_rates[1]) * 60.0F;
+    // Move to center using a regular G0 move, use whatever feed rate was previously defined for G0
     char buf[32];
     THEROBOT->push_state();
     THEROBOT->inch_mode = false;     // needs to be in mm
-    snprintf(buf, sizeof(buf), "G53 G0 X0 Y0 F%1.4f", rate); // must use machine coordinates in case G92 or WCS is in effect
+    snprintf(buf, sizeof(buf), "G53 G0 X0 Y0"); // must use machine coordinates in case G92 or WCS is in effect
```
wolfmanjm: FWIW I think this is totally inappropriate as the G0 modal speed could be anything, like after a retract it could be extremely slow.

src/modules/tools/endstops/Endstops.cpp
```diff
@@ -705,10 +720,12 @@ void Endstops::process_home_command(Gcode* gcode)
     // on some systems where 0,0 is bed center it is nice to have home goto 0,0 after homing
     // default is off for cartesian on for deltas
     if(!is_delta) {
-        // NOTE a rotary delta usually has optical or hall-effect endstops so it is safe to go past them a little bit
-        if(this->move_to_origin_after_home) move_to_origin(haxis);
         // if limit switches are enabled we must back off endstop after setting home
```
wolfmanjm: This removed something that is extremely important for rotary deltas. In general do not remove something if it is not understood why it is there :)

src/modules/tools/endstops/Endstops.cpp
```diff
@@ -705,10 +720,12 @@ void Endstops::process_home_command(Gcode* gcode)
...
+        // backoff endstops first so machine vibration cannot trip them during mot to origin
```
wolfmanjm: This is not a problem as endstops are disabled during the backoff (or should be)

src/modules/tools/endstops/Endstops.cpp
Zaaphod: I see what you mean and agree with you.  

My understanding of G0 is that it is always a rapid rate and not something that changes.  If you want to move something at a speed other than rapid G1 should be used.   In fact on all the old CNC controllers I have ever used, G0 had no option for an F parameter

Would it be more appropriate to just use default_seek_rate for this?  it would get put back to whatever was in there modal after the pop_state

src/modules/tools/endstops/Endstops.cpp
Zaaphod: If the endstops were disabled for the backoff then I agree it would be no problem, I didn't even think of that because I've never been able to enable or disable endstops before.  I suppose that the endstops must be disabled for the backoff because otherwise you wouldn't be able to get off the endstops.. I forgot it doesn't matter which one you hit.

src/modules/tools/endstops/Endstops.cpp
Zaaphod: Are the endstops disabled during the move_to_origin?  If I home to Y_Max what would happen if while moving to Y=0 it went too far perhaps because someone set the config incorrectly... would the Y-Min limit switch stop it or has it not been re-enabled yet?

wolfmanjm: limit switches are enabled during move_to_origin.

wolfmanjm: limit switches are disabled during move_to_origin. otherwise they could trigger initially.

Zaaphod: I think the intent with move_to_origin to be moving to the center of the table, this would be fine.  If I want the limit switches active, I can just do G28 then G28.1 to position to park, which could be 0,0 or anywhere else.  I'll put this back the way it was to prevent any other issues

PR #956, opened 2016-06-30
G28.1 Fix

Zaaphod:
Allow G28.1 to also be set by specifying positions.  This is necessary for G28.1 settings saved in config-override to be resorted.  Also added a line to indicate what the new G28.1 positions are, based off M206

I have tested this thoroughly on a CNC Router and CNC Mill, Please note the exception for rotary delta.  There is a similar exception for M206 & M306 so I felt it would be best to exclude rotary delta for this as well.   The rotary delta module may need to be looked at to see if it handles config-override setting for G28.1


wolfmanjm:
already done.. thanks

PR #958, opened 2016-06-30
G28.1 recovery

Zaaphod:
Recover G28.1 Positions from config-overide file

Requesting these be included in the current edge version for two reasons:

1. I am using these predefined positions on my machines in my homing routine, having them in the official edge version as well as edge-unstable will allow me to switch between the two versions and home correctly for testing purposes.

2. I wanted to contribute to the documentation here: http://smoothieware.org/supported-g-codes to include G28.1, G28.2, G28.3, and G28.4 however if the current edge version does not properly restore G28.1 it could cause issues for people using the current edge version.

### DIFF
--- a/src/modules/tools/endstops/Endstops.cpp
+++ b/src/modules/tools/endstops/Endstops.cpp
@@ -637,7 +637,11 @@ void Endstops::process_home_command(Gcode* gcode)
 
     } else if(gcode->subcode == 1) { // G28.1 set pre defined position
         // saves current position in absolute machine coordinates
-        THEKERNEL->robot->get_axis_position(saved_position);
+        THEROBOT->get_axis_position(saved_position); // Only XY are used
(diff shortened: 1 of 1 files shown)


wolfmanjm:
No sorry I cannot allow edge and edge-unstable diverge like this.

wolfmanjm:
However I can cherry pick it which will do the same thing.

Zaaphod:
I actually went back to edge and just made the changes there..  but whatever works would be appreciated

wolfmanjm:
actually I can't even cherry pick. I cannot allow them to diverge as it would make it impossible to maintain them, and no one uses G28.1 anyway. so is not critical. IMO it was supposed to only be used in grbl mode for CNC

triffid:
Can't put it in edge then merge it into edge-unstable, and keep them in
sync that way?

PR #959, opened 2016-07-02
Home Speed Adjustments

Zaaphod:
This is to adjust the feedrates during home procedures so they are the same whether you home one, two, or all three axis at the same time.

The most important section of this is the slow move back toward the limits.  If this is done at different speeds, the home position will be different.. that is why we do dual speed homing in the first place.

When doing a diagonal move at a feedrate, each axis needs to move slower so that the combined movement is at the specified feedrate.  This is working and is correct, however it causes the slow limit seek to be at a different speed if one axis is homed compared to if all three are homed.    Another issue was that if one axis was specified to be a different speed than the others, when homed together they would all go the same speed, however when homing individually, they would use the speed assigned to them, this was causing home speeds to be different during a multi-axis home compared to a single axis home.

There are two methods used to correct this, when homing there are two types of movements used, and each requires a different correction method. 
The first type of movement is moving toward the limit switches to find them. With this movement we must move at least far enough to hit them, but we could move further without hurting anything, because motion is stopped by the limit switch.  
This is the procedure used to adjust the feedrate for this method:
1. Find the minimum distance required to achieve the longest travel
2. Find the slowest designated feedrate
3. scale the distances moved so the faster axis needs to move further, This will slow down the others
4. a new feedrate is then calculated for the combined movement taking into account the speed designated speed of each axis
5. when this combined movement is made at the new feedrate, each individual axis will move  at precisely the specified feedrate.  This will be the same feedrate used if only one axis is homed. Also each axis will always have the required minimum distance required to make it to the switch.

The second type of movement is when we are backing off away from the switches.  For this move, we cannot change the lengths traveled. The reason for adjusting this is so if you have one axis that is drastically slower than the others, you can back it off at the same time while staying within the designated feedrates.
This is the procedure used for this method:
1. find the axis that needs the most amount of time to make it's movement, note that it is not necessarily the axis with the shortest distance or the slowest feedrate
2. calculate the combined length of all axis to be moved
3. calculate a new feedrate which will make the combined move in the time needed to accommodate the axis needing the most time to complete

when this method is used all axis will make the move needed in the amount of time required by the axis needing the longest time to complete.  note that during some movements different axis will move at different speeds if they are homed together as opposed to homing separately, but the exact speed is not important here as we are just getting off the switch. what is important is that no matter which axis are included in the home procedure, they will all take the time needed by the slowest requirement and therefore no axis can move faster than specified.

This has been tested on my CNC router and CNC mill with a wide rage of homing distance differences, speed differences. All axis moved at the expected rates.

### DIFF
--- a/src/modules/tools/endstops/Endstops.cpp
+++ b/src/modules/tools/endstops/Endstops.cpp
@@ -350,8 +350,8 @@ void Endstops::back_off_home(std::bitset<3> axis)
     if(!params.empty()) {
         // Move off of the endstop using a regular relative move
         params.insert(params.begin(), {'G', 0});
-        // use X slow rate to move, Z should have a max speed set anyway
-        params.push_back({'F', this->slow_rates[X_AXIS] * 60.0F});
+        // use the retract rate already calculated. It's the speed at which the slowest axis will complete at the speed specified
+        params.push_back({'F', retract_rate * 60.0F});
         char gcode_buf[64];
         append_parameters(gcode_buf, params, sizeof(gcode_buf));
         Gcode gc(gcode_buf, &(StreamOutput::NullStream));
@@ -447,10 +447,15 @@ void Endstops::home_xy()
 {
     if(axis_to_home[X_AXIS] && axis_to_home[Y_AXIS]) {
         // Home XY first so as not to slow them down by homing Z at the same time
-        float delta[3] {alpha_max*2, beta_max*2, 0};
+        float min_distance= std::max(alpha_max*2, beta_max*2); //mimimum distance that will accomodate either axis
+        float feed_rate = std::min(fast_rates[X_AXIS], fast_rates[Y_AXIS]); //slowest designated feedrate
+        //if X&Y have differnt home rates, this will make them each move at their designated rate during the combined move
+        //calulate new distance so they are relative to each other, this will cause each axis to move at the designated rate
+        float delta[3] {min_distance*fast_rates[X_AXIS]/feed_rate, min_distance*fast_rates[Y_AXIS]/feed_rate, 0}; 
+        feed_rate= sqrtf(pow(fast_rates[X_AXIS],2)+pow(fast_rates[Y_AXIS],2)); //calculate new feedrate for combined movement
         if(this->home_direction[X_AXIS]) delta[X_AXIS]= -delta[X_AXIS];
         if(this->home_direction[Y_AXIS]) delta[Y_AXIS]= -delta[Y_AXIS];
-        float feed_rate = std::min(fast_rates[X_AXIS], fast_rates[Y_AXIS]);
+
         THEROBOT->delta_move(delta, feed_rate, 3);
 
     } else if(axis_to_home[X_AXIS]) {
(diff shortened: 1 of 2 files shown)


wolfmanjm:
I basically do not agree with this at all. Sorry. But thanks for submitting it. I'll look at some of the code and consider if I want to use some of the ideas.
Basically slow rates should be used not some hack of it.
Moving off the endstop could be faster and I will try to see if what you did was valid or not.
There are some coding errors here too which would need to be fixed (you need to use the ..f() versions of math functions. I'll try to mark up where you need to fix those when I get some time.

Zaaphod:
I agree slow speeds should be used, but when you do a 3 axis move, only one of the axis is going at the speed you specified in the config.  if you have very different slow speeds for Z  and you use the slowest speed, now when you home xy and z together x and y are going way slower than if you home x or y by themselves.  this is to correct that so that when you do the 3 axis move all 3 axis are moving at precisely the slow speed specified

Thank you for having a look at it and I appreciate your comments.   sorry about the errors, I would really appreciate you pointing out how I should have done it, so I can learn to do it correctly.  Even if you don't use any of this, I would still like your opinion on the details in case I use something like it in my fork

Zaaphod:
correction.... when you do a 3axis move.. none of the 3 axis are ever going the speed you specified. 
If you have all equal distances and move F100, all 3 axis only move F58 each.. this creates a combined move of f100 but each axis is only hitting the switch at F58..  but when you home one axis by itself, it would go 100.

if you have one axis specified much slower, the problem is worse

I noticed this when my router was off by 1/8" after I homed y alone..   it's the difference between hitting the switches at 58 during a 3axis home compared to hitting them at 100.     I put a dial indicator on my machine to see the effect of hitting the switch at different rates.  While testing this, home was much much more reliable no matter what combination of axis I homed at a time

wolfmanjm:
Ok I think that your calculation of distance to determine speed is wrong. This is a homing sequence which means there is no way to know where the axis are so no way to know how far any axis will move.
This is why I choose the slowest. axis, however that is really not needed either as it will adhere to the max speed settings anyway, so simply selecting fast move would be sufficient as if it exceeds the speed it would be changed. However again given there is no way to know how far any axis will travel even that calculation would be incorrect. So I stick by what I did originally and move at the slowest of the given axis.

src/modules/tools/endstops/Endstops.cpp
```diff
@@ -447,10 +447,15 @@ void Endstops::home_xy()
 {
     if(axis_to_home[X_AXIS] && axis_to_home[Y_AXIS]) {
         // Home XY first so as not to slow them down by homing Z at the same time
-        float delta[3] {alpha_max*2, beta_max*2, 0};
+        float min_distance= std::max(alpha_max*2, beta_max*2); //mimimum distance that will accomodate either axis
```
wolfmanjm: says min distance but uses std::max

src/modules/tools/endstops/Endstops.cpp
```diff
@@ -502,31 +507,59 @@ void Endstops::home(std::bitset<3> a)
     // Move back a small distance for all homing axis
     this->status = MOVING_BACK;
     float delta[3]{0,0,0};
-    // use minimum feed rate of all three axes that are being homed (sub optimal, but necessary)
-    float feed_rate= slow_rates[X_AXIS];
+    float feed_time= 0;
+    // find slowest time for move taking into accound distance to move of all three axes and use that amount of time for movement
+    for ( int c = X_AXIS; c <= Z_AXIS; c++ ) {
+        if(axis_to_home[c]) {
+            feed_time= max(retract_mm[c]*(1/slow_rates[c]),feed_time);
```
wolfmanjm: need to use std::max

src/modules/tools/endstops/Endstops.cpp
```diff
@@ -502,31 +507,59 @@ void Endstops::home(std::bitset<3> a)
...
+        }
+    }
+    float combined_retract=0;
     for ( int c = X_AXIS; c <= Z_AXIS; c++ ) {
         if(axis_to_home[c]) {
             delta[c]= this->retract_mm[c];
             if(!this->home_direction[c]) delta[c]= -delta[c];
-            feed_rate= std::min(slow_rates[c], feed_rate);
+            combined_retract = sqrtf(pow(retract_mm[c],2)+pow(combined_retract,2));
```
wolfmanjm: needs to use powf

src/modules/tools/endstops/Endstops.cpp
```diff
@@ -502,31 +507,59 @@ void Endstops::home(std::bitset<3> a)
...
 
     // Start moving the axes towards the endstops slowly
     this->status = MOVING_TO_ENDSTOP_SLOW;
+    // the following finds the minimum distance to move that will accomodate all axis used
+    // starting off with a value all axis can use, then scale up from there to adjust 
+    // there is no problem with telling it to move further than nessecary to adjust relative speed of each axis
+    float min_retract=0;
+    for ( int c = X_AXIS; c <= Z_AXIS; c++ ) {
+        if(axis_to_home[c]) {
+            min_retract= max(min_retract,(this->retract_mm[c]*2));
```
wolfmanjm: use std::max

src/modules/tools/endstops/Endstops.cpp
```diff
@@ -502,31 +507,59 @@ void Endstops::home(std::bitset<3> a)
...
+    }
+    float fix_rate=0;
     for ( int c = X_AXIS; c <= Z_AXIS; c++ ) {
         if(axis_to_home[c]) {
-            delta[c]= this->retract_mm[c]*2; // move further than we moved off to make sure we hit it cleanly
+            // adjust the lengths of each axis so their speed relative to each other is what is specified
+            delta[c]= min_retract * slow_rates[c]/feed_rate;
+            //adjust feedrate so each axis is traveling at the speed specified by it's slow_rates
+            //this is so each axis will be travelling at the same speed whether homed individually or together
+            fix_rate= sqrtf(pow(slow_rates[c],2)+pow(fix_rate,2));
```
wolfmanjm: use powf.. this goes for a number of lines.

src/modules/tools/endstops/Endstops.cpp
Zaaphod: well it's the minimum allowable distance, but it's the largest distance to go.   It's the minimum because I use it as a base guaranteeing that the axis with the furthest distance to go will have that much room.. it's a minimum because I modify the distances by possibly making them longer, not shorter.  I could probably come up with a better variable name

src/modules/tools/endstops/Endstops.cpp
Zaaphod: noted - thank you

src/modules/tools/endstops/Endstops.cpp
Zaaphod: thanks again, will use powf

Zaaphod:
I see what you mean.  I thought it was a relative move of twice the axis length.  a relative move would make it acurate

wolfmanjm:
no it is the maximum it should move before giving up.

Zaaphod:
If I changed it to a relative move, then my formula would be correct

wolfmanjm:
no it would not be because there is no way to know how far it will actually move before it hits the endstop.

Zaaphod:
THEROBOT->delta_move(delta, feed_rate, 3);         is this not a relative move?

It doesn't matter when you hit the endstops, the speeds are set and correct because of how far it thinks it needs to move at the beginning of the move.. just because the move is cut short by the endstops makes no difference, they were all going at the correct speeds while moving

PR #974, opened 2016-07-09
Reject extremely short GCodes

Zaaphod:
Reject Gcodes which will not result in one step.  There is no need of filling the queue with useless commands that will not generate any motion.  These microscopic segments are typical if you are generating GCode base off true type fonts, and it's too much work to clean them up manually.

there was already a rejection in place for extremely small movements, I just took 1/steps_per_mm to figure out how far one step was and use that instead of an arbitrary number

Also removed check for mm_max_arc_error.  It's best to always use mm_max_arc_error.  you get just a few more segments at very small radii, but you are better off with them them if you are doing very fine work.

### DIFF
--- a/src/modules/robot/Robot.cpp
+++ b/src/modules/robot/Robot.cpp
@@ -1191,13 +1191,13 @@ bool Robot::append_arc(Gcode * gcode, const float target[], const float offset[]
 
     // discard segments so small they cause no movement
     if ((gcode->has_letter('X')) && (millimeters_of_travel < (1/(actuators[0]->get_steps_per_mm())))) {
-        return this->append_milestone(target, rate_mm_s);
+        return false;
     }
     if ((gcode->has_letter('Y')) && (millimeters_of_travel < (1/(actuators[1]->get_steps_per_mm())))) {
-        return this->append_milestone(target, rate_mm_s);
+        return false;
     }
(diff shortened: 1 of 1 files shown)


wolfmanjm:
this is already done in several paces no need to do it again.

src/modules/robot/Robot.cpp
```diff
@@ -1190,16 +1186,12 @@ bool Robot::append_arc(Gcode * gcode, const float target[], const float offset[]
     float millimeters_of_travel = hypotf(angular_travel * radius, fabsf(linear_travel));
 
     // discard segments so small they cause no movement
-    if ((gcode->has_letter('X')) && (millimeters_of_travel < (1/(actuators[0]->get_steps_per_mm())))) {
+    if (((gcode->has_letter('X')) && (millimeters_of_travel < (1/(actuators[0]->get_steps_per_mm())))) && 
+        ((gcode->has_letter('Y')) && (millimeters_of_travel < (1/(actuators[1]->get_steps_per_mm())))) &&
+        ((gcode->has_letter('Z')) && (millimeters_of_travel < (1/(actuators[2]->get_steps_per_mm()))))) {
```
Zaaphod: @wolfmanjm This formula still isn't right, and I know we don't need this.. but while I was putting it together I had the question of formatting for very long formulas like this.  

What is the preference for the smoothie project?  leave it one big long line, or break it up as I have here?  

Thanks for the advice

PR #976, opened 2016-07-09
Fix/jogging

DavidMFlynn:
After seeking home, panel controlled jogging is kept inside the travel
limits. Prevents having to re-seek home after turning the knob too far.
Before seeking home it function as before allowing the used move at
will even into a hard stop.  Tested on my 3D printer.


wolfmanjm:
I don't think this is generally useful. I think it also introduces potential issues with other types of machine.
It also breaks backward compatibility as it cannot be disabled and is on by default.
Lastly the check for being homed or not is inadequate as homing can be lost for various reasons.

I suggest this be kept as a separate fork in case anyone else would find it useful, and add it to thje thrid part forks list on the wiki.

Thank you for contributing, sorry this was not accepted.

FWIW I think enabling limits would be adequate for limiting the movements while jogging via panel.

pelrun:
What's the point of even trying to contribute to this project when you block and close anything you didn't write? Especially the jogging limits thing, which you seem to have a fixation on keeping out of the tree at all costs even though it's continually being requested by people.

Might as well go closed source.

arthurwolf:
Hey James.

We don't reject anything related to jog limits or soft endstops.
There has actually been quite a lot of work and discution about how to
implement this. This just happens not to match the way it has been figured
out this should be implemented.

For example you can read this document to get kind of an idea of what is
involved in this issue :
https://docs.google.com/document/d/1U6nzx1boqF-J2GGPWF4yIaaVib0JNodVWSKBfwiyp_M/edit#heading=h.79oxultkpcgc

You'll note that we accept new features all the time. There is a good
reason why that is not the case for soft endstops-related features, this is
not random, and it is not a block.
I expect we'll be getting a *good* implementation of this at some point,
but in the meantime, "partial"/"incorrect" implementations are likely to be
rejected.

Cheers.

On Sun, Jul 10, 2016 at 3:28 PM, James Churchill <notifications@github.com>
wrote:

> What's the point of even trying to contribute to this project when you
> block and close anything you didn't write? Especially the jogging limits
> thing, which you seem to have a fixation on keeping out of the tree at all
> costs even though it's continually being requested by people.
>
> Might as well go closed source.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/Smoothieware/Smoothieware/pull/976#issuecomment-231589066>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAGpFcxl22siOwBAVJ7EVYdMh22YAZMVks5qUPOUgaJpZM4JIrj2>
> .
>



-- 
Courage et bonne humeur.

pelrun:
Certainly, this PR isn't in a fit state to be committed as is, regardless of how closely it hews to your plan. But the way this was closed is very much "we don't commit things we won't personally use, go away and don't bother us again."

arthurwolf:
I think to the contrary, I don't think the way it was closed is how you
preceptive it at all. None of what you seem to think was meant was actually
meant. I think in this sort of context it's best to stick to what was
*actually* said ...

This doesn't fit the current plan for that feature, doesn't implement
things correctly even if it did. Thus closing it is perfectly reasonable.

The proper place for something like that, is in a separate fork ( which is
what was suggested ). That way if somebody wants to test/use that feature,
they can. And that way we can make progress, and things can eventually
evolve into something that can be merged ( or it stays in a fork forever ).

On Sun, Jul 10, 2016 at 3:53 PM, James Churchill <notifications@github.com>
wrote:

> Certainly, this PR isn't in a fit state to be committed as is, regardless
> of how closely it hews to your plan. But the way this was closed is very
> much "we don't commit things we won't personally use, go away and don't
> bother us again."
>
> —
> Reply to this email directly, view it on GitHub
> <https://github.com/Smoothieware/Smoothieware/pull/976#issuecomment-231590186>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAGpFXj_wpQwEU1VCqpxwjcg3bTEMpSuks5qUPlwgaJpZM4JIrj2>
> .
>



-- 
Courage et bonne humeur.

PR #983, opened 2016-07-16
Fixed StepperMotor direction being undefined

Viproz:
With dirpin inverted the value of direction was wrong until manually set.


wolfmanjm:
why is this needed? it does not seem relevant to anything?

wolfmanjm:
Ok I see the issue, it is manual step (which is actually rarely used) .. thank you.

PR #993, opened 2016-08-06
cmath include for ARM GCC 5.4 Q2-2016

milkowski:
The last release of ARM GCC toolchain (Q2-2016) needs additional <cmath> include for proper float calculations in DeltaCalibrationStrategy.cpp file.

### DIFF
--- a/src/modules/tools/zprobe/DeltaCalibrationStrategy.cpp
+++ b/src/modules/tools/zprobe/DeltaCalibrationStrategy.cpp
@@ -16,6 +16,7 @@
 
 #include <tuple>
 #include <algorithm>
+#include <cmath>
 
 #define radius_checksum         CHECKSUM("radius")
 #define initial_height_checksum CHECKSUM("initial_height")


wolfmanjm:
seems to works fine without. Can you explain why this is needed. It compiles fine for me and works fine for me.,

wolfmanjm:
FYI we don't use that toolchain and it is not supported at the moment. Please use the toolchain we supply...  gcc version 4.8.3 20140228 or keep your changes in your own fork. Thank you

milkowski:
FYI I use latest ARM GCC tools w/o problems for quite some time. The C++11 (gnu11 to be exact), that is default in 5.x line, has moved float math routines to separate "cmath" header for std namespace. So forward compatibility needs this change anyway.
http://stackoverflow.com/questions/14664325/c11-math-library-has-changed

wolfmanjm:
Until there is some overwhelming reason to upgrade the toolchain it will remain at the version it is currently at. Changing the tool chain is a major headache as it needs to be tested for all and every configuration and type of machine.

Issue #996, opened 2016-08-08
Problem with '?' key functionality -- output not flushed properly


johnkingsley:
#### What were you doing?

  I use the '?' command to get the current position.
  Like this:
```
?<Idle,MPos:0.0000,0.0000,0.0000,WPos:0.0000,0.0000,0.0000>
?<Idle,MPos:0.0000,0.0000,0.0000,WPos:0.0000,0.0000,0.0000>
?<Idle,MPos:0.0000,0.0000,0.0000,WPos:0.0000,0.0000,0.0000>
?<Idle,MPos:0.0000,0.0000,0.0000,WPos:0.0000,0.0000,0.0000>
```

That works fine. As soon as the ? key is pressed, the position is printed.

However when I move to X 100, like this:
```
G0X100
```
there is a problem.  The position isn't immediately printed when I press ?.
It requires me press the ? key a second time.

Like this:
```
??<Idle,MPos:100.0000,0.0000,0.0000,WPos:100.0000,0.0000,0.0000>
<Idle,MPos:100.0000,0.0000,0.0000,WPos:100.0000,0.0000,0.0000>
??<Idle,MPos:100.0000,0.0000,0.0000,WPos:100.0000,0.0000,0.0000>
<Idle,MPos:100.0000,0.0000,0.0000,WPos:100.0000,0.0000,0.0000>
??<Idle,MPos:100.0000,0.0000,0.0000,WPos:100.0000,0.0000,0.0000>
<Idle,MPos:100.0000,0.0000,0.0000,WPos:100.0000,0.0000,0.0000>
??<Idle,MPos:100.0000,0.0000,0.0000,WPos:100.0000,0.0000,0.0000>
<Idle,MPos:100.0000,0.0000,0.0000,WPos:100.0000,0.0000,0.0000>
??<Idle,MPos:100.0000,0.0000,0.0000,WPos:100.0000,0.0000,0.0000>
<Idle,MPos:100.0000,0.0000,0.0000,WPos:100.0000,0.0000,0.0000>
??<Idle,MPos:100.0000,0.0000,0.0000,WPos:100.0000,0.0000,0.0000>
<Idle,MPos:100.0000,0.0000,0.0000,WPos:100.0000,0.0000,0.0000>
```

#### What did you expect to happen?

The ? functionality should work everywhere the same and not be dependant on
the current location.

#### What happened instead?

See above.

#### Branch & Commit or Version of Smoothieware

[...]

PR #997, opened 2016-08-08
"Fix" problem with flushing 64 byte packets

johnkingsley:
Regarding issue  #996: 

When exactly 64 bytes are sent, it isn't flushed immediately.
(it is sent later, whenever we have more bytes to send)
If the data is 63 or less bytes, then it is flushed immediately.

This is a problem if the user is waiting for an immediate response from Smoothie.
For example, when sending the '?' character to get the location.

This "fix" is a work-around: we split the data into packets
of 63 bytes or less.  The actual bug should probably be hunted down
and properly fixed.

Other people have hit this problem.  E.g.
   https://community.nxp.com/message/360747

### DIFF
--- a/src/libs/USBDevice/USBSerial/USBSerial.cpp
+++ b/src/libs/USBDevice/USBSerial/USBSerial.cpp
@@ -137,8 +137,11 @@ bool USBSerial::USBEvent_EPIn(uint8_t bEP, uint8_t bEPStatus)
 
     int l = txbuf.available();
     if (l > 0) {
-        if (l > MAX_PACKET_SIZE_EPBULK)
-            l = MAX_PACKET_SIZE_EPBULK;
+        // Use MAX_PACKET_SIZE_EPBULK-1 below instead of MAX_PACKET_SIZE_EPBULK
+        // to work around a problem sending packets that are exactly MAX_PACKET_SIZE_EPBULK
+        // bytes in length. The problem is that these packets don't flush properly.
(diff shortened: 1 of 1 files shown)


wolfmanjm:
Thank you . I have been searching for this bug for a long time now. It is somewhere deep in the USB stack and really hard to find. 
This work around will do for now :)

johnkingsley:
Ya, this one was tricky to track down.
Thanks for merging.

triffid:
Nice find :)

wolfmanjm:
@triffid this is not a fix, it is a work around. The 64 byte bug is somewhere deeper in the USB stack. I have not been able to find that. Setting the max buffer size to 63 avoids that bug, that is still there just not triggered now.

triffid:
Hm the linked literature suggests that sending a 0-byte packet after a
final 64-byte packet should flush things, however I thought we were already
doing that..

Perhaps
https://github.com/johnkingsley/Smoothieware/blob/a06441bf86abd997ba7499fb1b202e4a3606ca08/src/libs/USBDevice/USBSerial/USBSerial.cpp#L150-L151
should be

if (txbuf.available() == 0) {
    if (l == MAX_PACKET_SIZE_EPBULK)
        send(b, 0); // send 0-byte packet after 64-byte queue flush to
indicate end of data
    r = false;
}

johnkingsley:
I started to dig into the USB internals when trying to figure this out. However it's not so simple. :-)

@triffid -- maybe your fix will work.  Give it a try! The bug is easy to reproduce: see #996

PR #1003, opened 2016-08-21
New config example

arthurwolf:
This is not for merging yet, just for review/discution.

I'm trying to re-do part of the example config file.

Things this is trying to achieve : 
* Remove options we don't want people to modify anyway ( making sure they are still in the documentation )
* Have less config options in the file overall ( using defaults in code )
* Have better comments/info for each option ( based on commonly misunderstood things from IRC/forum )
* Have more links to the documentation, and more warnings

### DIFF
--- a/ConfigSamples/Smoothieboard/config
+++ b/ConfigSamples/Smoothieboard/config
@@ -1,325 +1,236 @@
-# NOTE Lines must not exceed 132 characters
-## Robot module configurations : general handling of movement G-codes and slicing into moves
-default_feed_rate                            4000             # Default rate ( mm/minute ) for G1/G2/G3 moves
-default_seek_rate                            4000             # Default rate ( mm/minute ) for G0 moves
-mm_per_arc_segment                           0.0              # Fixed length for line segments that divide arcs 0 to disable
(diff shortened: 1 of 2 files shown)


wolfmanjm:
i'd move laser into  snippets and use include.
i'd move panel into snippets and use include. (I do this on mine).
I'd mention at the start that many settings can be overridden with M500 and checked with M503

wolfmanjm:
Maybe temperature control should also be in an included file.

arthurwolf:
I'm not sure about using includes for the default example config file. 
While it's neater, it's going to confuse a lot of users ...
I would need to first go over all of the documentation and change a lot of things to explain/use includes before that happened I think, so it'd be for a next revision of the config example.

Completely agree on M500

wolfmanjm:
how about having laser in a separate examples? strip out all the temperature and  extruder stuff. it will look a lot simpler for laser users.
Also for a future version I'd split all sections out like motor pins temp, extruder probe, endsrops, and just include them in the main file. this will make having support for azteeg etc much simpler. 
maybe a section just for pin definitions, which is the main difference between azteeg and sb. then they can share the includes for the other sections?

arthurwolf:
Yeah that does sound good that's a direction we should be going in

PR #1004, opened 2016-08-21
Repaired Morgan scara homing on new motor control

qharley:
Added a boolean that can inform arm solution that homing is active.  Scara homing relies on cartesian homing movement rather than inverse kinematics.

### DIFF
--- a/src/modules/robot/arm_solutions/BaseSolution.h
+++ b/src/modules/robot/arm_solutions/BaseSolution.h
@@ -17,6 +17,7 @@ class BaseSolution {
         typedef std::map<char, float> arm_options_t;
         virtual bool set_optional(const arm_options_t& options) { return false; };
         virtual bool get_optional(arm_options_t& options, bool force_all= false) { return false; };
+        bool homing_active;
 };
 
 #endif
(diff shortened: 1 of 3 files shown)


qharley:
I prepared this pull request on the latest edge-unstable from Smoothieware.

wolfmanjm:
this really does not fit in with the way I would do things. Sorry 
Also it has merge conflicts.

wolfmanjm:
I would allow the fixes to the scara homing, but not the nasty hack to detect homing. I'll try to figure out an acceptable way to do that.

qharley:
Ok. I could think of no other way to do it, unless there is a way to see
the homing status of endstops as I requested before.. In the mean time I
will do it this way in my repo.

Thanks.

qharley:
PS. Why the merge conflicts? I did the modification on
Upstream/edge-unstable

wolfmanjm:
there is no edge-unstable branch, no idea where you got that from.

qharley:
Git says that upstream/edge is ambiguous...

wolfmanjm:
Then you have a misconfigured git I suggest you start from a clean clone

qharley:
K

PR #1005, opened 2016-08-22
Scara homing repair: New motor control

qharley:
Based Morgan scara homing fix on actual Smoothieware/edge
This still has the homing notifier hack though

### DIFF
--- a/src/modules/robot/Robot.cpp
+++ b/src/modules/robot/Robot.cpp
@@ -865,7 +865,16 @@ void Robot::reset_axis_position(float x, float y, float z)
 
     // now set the actuator positions to match
     ActuatorCoordinates actuator_pos;
-    arm_solution->cartesian_to_actuator(this->last_machine_position, actuator_pos);
+    if(!disable_arm_solution) {
+        arm_solution->cartesian_to_actuator( last_machine_position, actuator_pos );
+
+    }else{
+        // DEBUG THEKERNEL->streams->printf("c2a robot reset_axis_position\n");
+        // basically the same as cartesian, would be used for special homing situations like for scara
+        for (size_t i = X_AXIS; i <= Z_AXIS; i++) {
+            actuator_pos[i] = last_machine_position[i];
+        }
+    }
     for (size_t i = X_AXIS; i <= Z_AXIS; i++)
         actuators[i]->change_last_milestone(actuator_pos[i]);
 }
@@ -905,14 +914,33 @@ void Robot::reset_position_from_current_actuator_position()
     }
 
     // discover machine position from where actuators actually are
-    arm_solution->actuator_to_cartesian(actuator_pos, last_machine_position);
+    if(!disable_arm_solution) {
+        arm_solution->actuator_to_cartesian(actuator_pos, last_machine_position);
+
+    }else{
+        // DEBUG THEKERNEL->streams->printf("a2c robot reset_position_from_current_actuator_position\n");
+       // basically the same as cartesian, would be used for special homing situations like for scara
+        for (size_t i = X_AXIS; i <= Z_AXIS; i++) {
+            last_machine_position[i] = actuator_pos[i];
+        }
+    }
+
     // FIXME problem is this includes any compensation transform, and without an inverse compensation we cannot get a correct last_milestone
     memcpy(last_milestone, last_machine_position, sizeof last_milestone);
 
     // now reset actuator::last_milestone, NOTE this may lose a little precision as FK is not always entirely accurate.
     // NOTE This is required to sync the machine position with the actuator position, we do a somewhat redundant cartesian_to_actuator() call
     // to get everything in perfect sync.
-    arm_solution->cartesian_to_actuator(last_machine_position, actuator_pos);
+    if(!disable_arm_solution) {
+        arm_solution->cartesian_to_actuator( last_machine_position, actuator_pos );
+
+    }else{
+        // basically the same as cartesian, would be used for special homing situations like for scara
+        // DEBUG THEKERNEL->streams->printf("c2a robot reset_position_from_current_actuator_position\n");
+        for (size_t i = X_AXIS; i <= Z_AXIS; i++) {
+            actuator_pos[i] = last_machine_position[i];
+        }
+    }
     for (size_t i = X_AXIS; i <= Z_AXIS; i++)
         actuators[i]->change_last_milestone(actuator_pos[i]);
 }
--- a/src/modules/tools/endstops/Endstops.cpp
+++ b/src/modules/tools/endstops/Endstops.cpp
@@ -128,6 +128,7 @@ enum {
 Endstops::Endstops()
 {
     this->status = NOT_HOMING;
+    THEROBOT->disable_arm_solution = false;
     home_offset[0] = home_offset[1] = home_offset[2] = 0.0F;
     debounce.fill(0);
 }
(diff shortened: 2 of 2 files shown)


wolfmanjm:
Try this again using PR #1007 
Basically remove your hack and instead just do ```THEROBOT->disable_arm_solution= true;``` before you home and reset after you home. Must be left as false no matter what, so make sure you cover all entries and exits.

src/modules/robot/arm_solutions/MorganSCARASolution.cpp
```diff
@@ -58,7 +60,7 @@ float MorganSCARASolution::to_degrees(float radians) const
 
 void MorganSCARASolution::cartesian_to_actuator(const float cartesian_mm[], ActuatorCoordinates &actuator_mm ) const
 {
-
+  if (!THEROBOT->disable_arm_solution){
```
wolfmanjm: This is wrong. there are no changes to the arm solution. when disable-arm-solution is set the arm solution is never called by robot.

qharley: I have proven it to be otherwise.  I put a stream print function in the
arm_solution cartesian_to_actuator, and actuator_to_cartesian and during
homing these commands gets called a number of times.

wolfmanjm: then there is something wrong.either with your merge or with the homing.

qharley:
Ok.  I will do it again tomorrow on a fresh upstream/edge, and submit the proof, with my faultfinding code in tact in a separate branch for you to look at.

wolfmanjm:
really don't bother to do that. Instead find out where it is getting called from  Workarounds to bugs are not really appropriate. If for some reason arm solution is being called during homing after you set that flag there is a bug, find that bug.

qharley:
Ok.

qharley:
arm solution is called indirectly by Endstops.cpp in lines 629 and 637 (THEROBOT->reset_axis_position)
and 509 and 544 (THEROBOT->reset_position_from_current_actuator_position) even when the flag is set

wolfmanjm:
Yes but that is not during a homing move that is after it is complete.

qharley:
No.  Those actually get called just before the movement, and is somehow
responsible for the robotic dance that ensues.  If I replace those calls
with cartesian results (as per my hack) it homes as normal.

I will post a flow diagram of what happens here later when I have access to
a non printing machine.


[image: photo]
*Quentin Harley*
Founder, Morgan 3D Printers
p:+27120041883 | e:quentin@morgan3dp.com | w:www.morgan3dp.com | a:4 Burger
Ave, Lyttelton Manor, Centurion

wolfmanjm:
Those are used to set the position to a known good position before homing. The other arm solutions need that. 
If scara does not then don't call them. However I don't see why scara would get upset by setting the reset position to 0,0,0 which it does anyway when it boots.
When it starts to actually home the flag I gave you will work.
If you don't want to do it that way that is fine, you can do whatever works for you that you are comfortable with :)

qharley:
It is strange that it has an effect, I agree.  Let me dig a bit more.
All I want is to keep SCARA in smoothieware.  If homing is broken, so is SCARA and then we have to remove it.

qharley:
I added some comments to show where calls to arm_solutions originate from while disable_arm_solution = true.

endstops process_home_command
disable_arm_solution = true
c2a robot reset_axis_position
c2a robot reset_axis_position
c2a robot reset_axis_position
endstops home
c2a robot append_milestone
endstops home_xy
c2a robot append_milestone
a2c robot reset_position_from_current_actuator_position
c2a robot reset_position_from_current_actuator_position
c2a robot append_milestone
c2a robot append_milestone
a2c robot reset_position_from_current_actuator_position
c2a robot reset_position_from_current_actuator_position
disable_arm_solution = false

If any of those calls actually get to arm_solutions:
1) the direction of homing turns around.  SCARA is polar, and the angles are inverse to the cartesian coordinates.
2) after the initial homing, the arms flail around aimlessly as the motors accelerates too fast to get to position very far away from where it is.

qharley:
This latest attempt keeps everything in Robot and Endstops.  I tried doing less, but this is the minimum change that will make it home successfully.

wolfmanjm:
well you are doing something wrong somewhere. Simply don't call the reset_axis_position at the start, and the calls after the homing are required otherwise you are not setting homing position.
You may need to do something different after homing to set the home position.
the calls from append_milestone are blocked if you actually set the flag and those are the only ones that will move the arms when it is homing.

wolfmanjm:
Why are you changing anything in robot? that would not be acceptable.

qharley:
0 in angle space is negative angles.  We need the polar angles to be zero for this to work, and to do that we need to bypass the arm solution

qharley:
Just look at my pull request...

wolfmanjm:
Sorry none of your changes to robot will be acceptable. There is no reason that test needs to be done in reset axis position. the only change is the one I made.

wolfmanjm:
Just don't call reset_axis position if it doesn't work for you. all your changes need to be isolated in endstops only.

qharley:
Lets see...

wolfmanjm:
look I know it is hard, but the only place you need to ignore the arm solution is when the thing is moving. the reset axis stuff does not move the arm. So rather than changing stuff until it works, you need to figure out what the hell is going on and fix the homing sequence for scara. I can guarantee you that the changes to reset axis are not what you need to do. and I don't want hacks in the robot code that no one understands why they are there :)

wolfmanjm:
also you did not need those changes in reset axis before so you don't need them now as that part of the code did not change. (between master and edge)

wolfmanjm:
The only call that makes an arm move is one to append_milestone().

wolfmanjm:
maybe you need to use reset_actuator_position instead of reset_axis_position. That is what I needed to do for the rotary delta.

qharley:
Disabling the calls did not work.  I will try replacing them with reset_actuator_position as you suggested.

wolfmanjm:
No you need to understand what is going on and what you are doing. this is a fundamental issue.

When you home you are setting the actuators to a known position, in your case (and rotary delta) this is an angle not a cartesian coordinate. So calling reset_axis is not the call you want to make as that sets the cartesian axis after homing.
In your case you need to call reset actuator as you are setting the actuaot angles, it will then use the FK to figure out what the cartesian axis are based on the actuator positions. This means that your homing offsets and alpha_max etc are no longer cartesian coordinates but actuator angles. so you will need to change your config as well accordingly.
Does that make sense? If not I recommend you stick with the master branch where it worked for you.
Sorry to be a pain, but I cannot allow Marlin-like hacks-until-it-works type of stuff into smoothie.

qharley:
Ok.  Seems like you will have to remove SCARA for the time being.  It will remain broken for some time to come.

wolfmanjm:
So that did not make sense? I'd do this for you but I don't have a scara and I don't understand how scara works :)
But I'm pretty sure it is not unlike the rotary delta stuff. except Z is a separate axis for you.

qharley:
Don't worry, I will try again.  I just need to work out a strategy to get it to do what I need without the need to hack anything...

wolfmanjm:
I think the problem is the existing is_scara stuff is totally wrong. as it no longer applies. I should not have left it in. I'll remove that so you can have a clean slate.
Now as I understand it you have two actuators that control X and Y and an actuator that controls Z which is a normal linear actuator? In this case it is similar to a corexy.
So what you need to do is home Z separately like corexy does. The Z homing will be normal and the same as corexy.

For XY homing you need to start the XY actuators moving in whatever direction they need to go, when they need to stop depends on how they work. either stop each individually, or both at the same time. then move one until it hits. At that point you know the actuator angles, so you then use reset_actuator_position() but you pass in the angles for X and Y (I may have to allow it to take just XY and not Z) then that sets the actuator angle directly, then uses the FK to figure out what the equivalent cartesian XY is. (this is similar to rotary delta but using only 2 actuators).

I still think that writing a scara homing module using manual stepping would be easier for you to do this. then you will find merging with edge in future much easier. and people maintaining endstops don't need to worry about how a scara works.

qharley:
Thanks.  I think it is time for me to refactor the whole scara codebase.
Will open another pull request at that time

PR #1010, opened 2016-08-26
Fixed rotary encoder implementation instead of menu offset workaround that did not work (was not implemented for eg jog)

OpenEHS:
Old work around for different types of rotary encoders (X2/X4) was implemented with a menu offset, this causes problems eg. in jog screen instead of 1 step there were 2 steps. Instead of using a menu offset and complicate all the menu/panel related code we implement the encoder readout different so we can support X2/X4 encoding

* Panel.cpp/Panel.h:
  - removed menu_offset_checksum (no longer necessary)
  - removed all menu_offsets in the code
* ReprapDiscountGLCD/ST7565:
  - add menu_offset, if == 0 (X4 mode, same as before), else X2 mode

### DIFF
--- a/src/modules/utils/panel/Panel.cpp
+++ b/src/modules/utils/panel/Panel.cpp
@@ -49,7 +49,6 @@
 #define mini_viki2_checksum        CHECKSUM("mini_viki2")
 #define universal_adapter_checksum CHECKSUM("universal_adapter")
 
-#define menu_offset_checksum        CHECKSUM("menu_offset")
 #define encoder_resolution_checksum CHECKSUM("encoder_resolution")
 #define jog_x_feedrate_checksum     CHECKSUM("alpha_jog_feedrate")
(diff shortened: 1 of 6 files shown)


wolfmanjm:
I do not think this is compatible with what we already do. Menu offset is not supposed to affect anything except menu stepping here it changes the actual resolution. whereas the resolution is what is supposed to do that. that is why menu_offset is in panel and not in the lower driver level.
This will break existing configs causing confusion and annoyance.
I'll test it to confirm later.

wolfmanjm:
This also only allows a menu offset on or off whereas before it could be set to different values.
Overall I think this is a regression.

wolfmanjm:
FYI the encoder_resolution setting changes the overall number of encoder transitions that change things, and menu_offset adds an additional setting to only affect menu selection.
This PR breaks this making menu offset set encoder click globally affecting value changes as well as menu selection.

PR #1012, opened 2016-08-29
Fix preset position

Zaaphod:
This is to fix a bug when moving to preset position with G28.2 in normal mode or G28 in grbl mode.  The problem was that if you were in G20 (inch mode) and set a predefined position with G28.1, it saved it in mm, which in itself was no problem, however when moving to the predefined position with G28.2/G28 it was not forcing G21, so it was moving the saved distance in inches instead of mm, so it was going 25.4 times too far.  To make sure the same position is achieved whether in G20 or G21, the predefined position is still saved in mm as it was, however now it forces it out of inch mode if it was active so the move correctly positions. the G20/G21 mode is then restored to what it was before the command was issued.

I also implemented two options, both of which I need on my cnc machines, so I thought I would share them.   

First is move_to_park_after_home which is the same as move_to_origin_after_home but it moves to the predefined position instead of 0,0.   This is very useful for any machine homing off either a Xmax or Ymax switch because it is unlikely anyone homing off a positive limit switch would want to move all the way to 0,0 where the machine is now in the way of loading, while it would be very useful to automatically go to the park position.  

The second option is back_off_first.  This simply allows control over whether to do the backoff before or after move_to_origin_after_home or move_to_park_after_home.    This is only in effect if the machine is not delta and it defaults to false, so leaving it out makes it exactly the same as it was.


wolfmanjm:
I am rejecting this as it violates the submission rules.
only one feature per pull request, this has at lest two.
You need to separate these in two PR. As I may not accept the homing one for instance.

PR #1015, opened 2016-08-30
Features/add laser features

wolfmanjm:
Add M221 Sxx to scale the laser power by xx%
Add fire xx to override the automatic fire control and fire the laser at xx% power, fire off will return it to auto mode.
Fix the proportional power adjustment on acceleration and deceleration.

### DIFF
--- a/src/modules/tools/laser/Laser.cpp
+++ b/src/modules/tools/laser/Laser.cpp
@@ -41,6 +41,7 @@ Laser::Laser()
 {
     laser_on = false;
     scale= 1;
+    manual_fire= false;
 }
 
 void Laser::on_module_loaded()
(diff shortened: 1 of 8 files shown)


src/modules/tools/laser/Laser.cpp
```diff
@@ -103,11 +106,77 @@ void Laser::on_module_loaded()
...
 
     // no point in updating the power more than the PWM frequency, but no more than 1KHz
     THEKERNEL->slow_ticker->attach(std::min(1000UL, 1000000/period), this, &Laser::set_proportional_power);
 }
 
+void Laser::on_console_line_received( void *argument )
+{
+    if(THEKERNEL->is_halted()) return; // if in halted state ignore any commands
+
+    SerialMessage new_message = *static_cast<SerialMessage *>(argument);
```
tritao: This way involves a struct copy (SerialMessage has std::string, so its likely copy will allocate more memory) .

Why not keep this a pointer?

src/modules/tools/laser/Laser.cpp
```diff
@@ -103,11 +106,77 @@ void Laser::on_module_loaded()
...
+            new_message.stream->printf("turning laser off and returning to auto mode\n");
+
+        }else{
+            p= strtof(power.c_str(), NULL);
+            p= confine(p, 0.0F, 100.0F);
+            new_message.stream->printf("WARNING: Firing laser at %1.2f%% power, entering manual mode use fire off to return to auto mode\n", p);
+        }
+
+        p= p/100.0F;
+        manual_fire= true;
```
tritao: This code looks almost the same as `set_proportional_power`.

Why not refactor common code into new `set_laser_power` and call it from here and `set_proportional_power`.

That way there is less duplicated code for manipulating laser pins.

tritao:
Nice to see new laser features, Jim.
Will test this later when I have access to the machine.

I added a few suggestions that came to mind while reviewing the code.

tritao:
Just saw you merged without reply to any of my comments. If they didn't make sense it'd be nice if you could explain why so I know more about Smoothie code style in case I contribute anything in future, thanks.

tritao:
Trying to understand what these new changes do, but I do not know some terminology. What is the difference between nominal rate and plateau rate? Anywhere where this is documented / explained? Thanks.

wolfmanjm:
@tritao It has not been merged, I'm still working on it. Your comments were optimizations and coding style related, and I may refactor the code once it is tested and before it is merged.

tritao:
Yeah my bad. Have you tested how big the changes affect laser engraving quality?

PR #1022, opened 2016-09-13
Features/laser panel

wolfmanjm:
Add laser menu
  sets power scaler
  does test fire
Add Dynamic Menu class
Add Goto park position and set park position for CNC mode
Add current laser power value (S value) and actual current power to watch screen
Fix potential jogging issue for 3dprinter which was fixed in CNC mode (absolute vs relative)


arthurwolf:
Wow ...

PR #1024, opened 2016-09-14
add inverse compensation transform

wolfmanjm:
Add an inverse compensation transform for all ABL types.
Use inverse transform for resetting axis from actuator position.
All probes leave compensation transform on now by default.
Homing will use inverse compensation transform to get actual Z position at the given XY position.
? and M114.1 and M114.2 will return corrected position based on compensation transform.

> The reasoning of having reset_axis setting compensated_machine_position and applying inverse transform to get machine_position is that most Z endstops are fixed on the Z actuator and trigger at the same place regardless of where the XY is. So consider a sloped bed when you home in Z you will always be at the same height, but as the bed is sloped wrt to the head you want to calculate the WCS position (returned with M114) to be the actual height above the bed at the current XY. Which will differ depending on where the XY is at the time Z is homed.

These changes only affect systems where ABL is active and a zprobe or home is performed while active.
To probe raw bed heights turn off the ABL first.

Other changes...
Fix M306 so it does not reset axis position and acts more like M206 which needs a home cycle to take effect.
Removed ZGrid from build as it has been replaced by delta grid set for square beds.
Changed G30 Z0 to use G92 to set the global offset.
Refactor naming of last_milestone in Robot to machine_position.


wolfmanjm:
Replaced by PR #1037

Issue #10, opened 2012-04-14
Fixed point counters are not big enough


arthurwolf:
Originally in : 
https://github.com/arthurwolf/Smoothie/issues/9

The fixed point counters in Stepper are 32bits, which is not good enough for long linear moves, and can cause bugs.
A temporary solution is to reduce mm_per_line_segment. 

But the bitmath of this should be changed.

PR #1035, opened 2016-10-06
Feature/file sorter

michael-moore0:
I have made several modifications and customizations to the Smoothieware codebase to fit my requirements and preferences.  Some of these might be useful to others, some are to use-case specific.  I consulted with Arthur Wolf via email, and he suggested that I start with a pull request for one of the features which may be useful to others.

The feature in this pull request is an alphabetic file list sorter.  I found it difficult to navigate the file list as the order was not easily discernible (system default).  I thought it would be helpful if the files were sorted in alphabetic order.  To that end, I created a class 'FileSorter' to accomplish that task.  The class takes a directory in its constructor and sorts the file names, which are then available through several methods.  An instance may be re-used for a different directory as well, through the open method.  Currently, the sort order is alphabetically by file name, but that may easily be expanded to additional sort orders if needed.  The class uses a callback function to filter unwanted file names from the list.  When the callback is NULL, all files in the folder are included.

I used an object of this class in the 'ls' command handler to sort the results of the command.  I also used an instance in the FileScreen class (as a member variable this time).  The FileScreen re-uses a single instance of FileSorter from 'enter_folder' till 'on_exit'.  Since the FileSorter keeps a cache of the sorted file names, the FileScreen 'file_at' method is greatly optimized.   Rather than read through the directory contents from the beginning for each call, the file name is retrieved by array index.  This of course uses more memory then the list-every-time approach.  I've tested it on my system (Azteeg X5 mini) with folders containing up to several hundred files.  In order to make it fail, I added 1500 files to a test folder.  When it fails due to memory constraints, an error message is displayed.  The essence of engineering is balancing tradeoffs.  I realize that the tradeoff here is that folders with an exceptionally large number of files cannot be displayed.  However, I think the speed gain in looking through the files is well worth it.  Scrolling to the end of a folder list with several hundred items took a very long time under the previous approach.

Please let me know what you think.  Thanks!

### DIFF
--- a/src/libs/FileSorter.cpp
+++ b/src/libs/FileSorter.cpp
@@ -12,7 +12,7 @@
 #include "PublicData.h"
 #include "modules/utils/player/PlayerPublicAccess.h"
 #include "SimpleShell.h"
-#include "libs/Kernel.h"
+#include "Kernel.h"
 
 #include <new>
 #include <malloc.h>
--- a/src/libs/FileSorter.cpp
+++ b/src/libs/FileSorter.cpp
@@ -77,12 +77,12 @@ void FileSorter::delete_file_info_array(void)
 void FileSorter::open_directory(string dir_path)
 {
     // reset all member state variables.
+    delete_file_info_array();
     this->directory_path = dir_path;
     this->error = false;
     this->file_count = 0;
     this->total_file_count = 0;
     this->file_index = -1;
-    delete_file_info_array();
 
     DIR *d;
     d = opendir(this->directory_path.c_str());
@@ -100,6 +100,7 @@ void FileSorter::open_directory(string dir_path)
             this->file_count = 0;
             array_entry = this->file_info_array;
             while ( this->file_count < this->total_file_count && (file_info = readdir(d)) != NULL ) {
+
                 // if the filename passes the filter, add it to the sort array.
                 if ( (this->filter_callback == NULL) || this->filter_callback(file_info) ) {
                     array_entry->is_dir = file_info->d_isdir;
--- a/src/libs/FileSorter.cpp
+++ b/src/libs/FileSorter.cpp
@@ -12,7 +12,7 @@
 #include "PublicData.h"
 #include "modules/utils/player/PlayerPublicAccess.h"
 #include "SimpleShell.h"
-#include "kernel.h"
+#include "libs/Kernel.h"
 
 #include <new>
 #include <malloc.h>
--- a/src/libs/FileSorter.cpp
+++ b/src/libs/FileSorter.cpp
@@ -10,9 +10,10 @@
 #include "FileStream.h"
 #include "checksumm.h"
 #include "PublicData.h"
-#include "modules/utils/player/PlayerPublicAccess.h"
+#include "PlayerPublicAccess.h"
 #include "SimpleShell.h"
 #include "Kernel.h"
+#include "StreamOutput.h"
 
 #include <new>
 #include <malloc.h>
@@ -23,13 +24,8 @@
 
 FileSorter::FileSorter(string dir_path, __file_sort_filter_fn_t callback)
 {
-    this->directory_path = dir_path;
-    this->error = false;
-    this->file_info_array = NULL;
-    this->file_count = 0;
-    this->total_file_count = 0;
-    this->file_index = -1;
     this->filter_callback = callback;
+    this->file_info_array = NULL;
 
     open_directory(dir_path);
 }
@@ -71,7 +67,6 @@ void FileSorter::sort_directory(void)
 {
     // sort the array.
     qsort(this->file_info_array, this->file_count, sizeof(file_info_t), (__compar_fn_t)compare_dir);
-    this->file_index = -1;
 }
 
 void FileSorter::delete_file_info_array(void)
@@ -84,6 +79,7 @@ void FileSorter::delete_file_info_array(void)
 
         // free the array.
         delete[] this->file_info_array;
+        this->file_info_array = NULL;
     }
 }
 
@@ -91,11 +87,12 @@ void FileSorter::open_directory(string dir_path)
 {
     // reset all member state variables.
     delete_file_info_array();
+    //
     this->directory_path = dir_path;
     this->error = false;
+    this->sorted = false;
     this->file_count = 0;
     this->total_file_count = 0;
-    this->file_index = -1;
 
     DIR *d;
     d = opendir(this->directory_path.c_str());
@@ -108,14 +105,14 @@ void FileSorter::open_directory(string dir_path)
         // or when there's not enough heap memory available for the sort array.
         uint32_t free_bytes = SimpleShell::heapWalk(NULL, false);
         uint32_t bytes_required = (this->bytes_required_for_names + (this->total_file_count * sizeof(file_info_t)) + FILE_NAME_ARRAY_MIN_PADDING);
-        this->error = (!FileSorter::can_do_sort() || free_bytes < bytes_required);
-        if ( !this->error ) {
+        this->sorted = (FileSorter::can_do_sort() && free_bytes > bytes_required);
+        if ( this->sorted ) {
 
             // allocate the array used for sorting.
             this->file_info_array = new (std::nothrow) file_info_t[this->total_file_count];
 
             // read the file info into the array.
-            if ( !(this->error = (this->file_info_array == NULL)) ) {
+            if ( (this->sorted = (this->file_info_array != NULL)) ) {
                 int itr = 0;
                 struct dirent* file_info;
                 file_info_t* array_entry;
@@ -130,7 +127,7 @@ void FileSorter::open_directory(string dir_path)
                         array_entry->file_name = strdup(file_info->d_name);
 
                         // ensure that the filename was copied successfully.
-                        if ( (this->error = (array_entry->file_name == NULL)) ) {
+                        if ( !(this->sorted = (array_entry->file_name != NULL)) ) {
                             // failed to copy the filename, clean up and get out.
                             delete_file_info_array();
                             break;
@@ -150,8 +147,56 @@ void FileSorter::open_directory(string dir_path)
         // close the directory handle and sort the file array.
         closedir(d);
         d = NULL;
-        if ( !this->error ) {
+        if ( this->sorted ) {
             sort_directory();
+        }
+    }
+}
+
+void FileSorter::print_file_list(StreamOutput* stream, bool verbose)
+{
+    // attempt to use the file sorter to sort alphabetically.
+    if ( this->sorted ) {
+        file_info_t* file_info;
+
+        for ( size_t i=0; i<this->file_count; i++ ) {
+            if ( (file_info = &this->file_info_array[i] ) != NULL ) {
+                //stream->printf("%s", lc(string(files->get_file_name())).c_str());
+                stream->printf("%s", file_info->file_name);
+                if ( file_info->is_dir ) {
+                    stream->printf("/");
+                } else if ( verbose ) {
+                    stream->printf(" %d", file_info->file_size);
+                }
+                stream->printf("\r\n");
+            }
+        }
+
+    // if the file sorter fails (likely due to memory constraints),
+    // just list the files the old fashioned way.
+    } else {
+        DIR *d;
+        struct dirent *p;
+        d = opendir(this->directory_path.c_str());
+        if (d != NULL) {
+            int itr = 0;
+            while ((p = readdir(d)) != NULL) {
+                stream->printf("%s", p->d_name);
+                if(p->d_isdir) {
+                    stream->printf("/");
+                } else if( verbose ) {
+                    stream->printf(" %d", p->d_fsize);
+                }
+                stream->printf("\r\n");
+
+                // process ON_IDLE every 20 iterations.
+                if ( itr++ % 20 ) {
+                    THEKERNEL->call_event(ON_IDLE);
+                }
+           }
+            closedir(d);
+        } else {
+            stream->printf("Could not open directory %s\r\n", this->directory_path.c_str());
         }
     }
 }
@@ -166,57 +211,42 @@ size_t FileSorter::get_total_file_count(void)
     return this->total_file_count;
 }
 
-const char* FileSorter::get_file_name(void)
+const char* FileSorter::file_at(int index, bool& isdir)
 {
-    if ( this->file_index >= 0 && (size_t) this->file_index < this->file_count ) {
-        return this->file_info_array[this->file_index].file_name;
-    } else {
-        return NULL;
-    }
-}
+    // attempt to sort the files alphabetically.
+    if ( this->sorted ) {
+        if ( index >= 0 && (size_t) index < this->file_count ) {
+            isdir = this->file_info_array[index].is_dir;
+            return this->file_info_array[index].file_name;
+        } else {
+            return NULL;
+        }
 
-size_t FileSorter::get_file_size(void)
-{
-    if ( this->file_index >= 0 && (size_t) this->file_index < this->file_count ) {
-        return this->file_info_array[this->file_index].file_size;
+    // if the file sorter failed (likely due to memory constraints), do
+    // the filename lookup the old fashioned way.
     } else {
-        return 0;
-    }
-}
+        DIR *d;
+        struct dirent *p;
+        uint16_t count = 0;
+        d = opendir(THEKERNEL->current_path.c_str());
+        if (d != NULL) {
+            int itr = 0;
+            while ((p = readdir(d)) != NULL) {
+                if ( ((this->filter_callback == NULL) || this->filter_callback(p)) && index == count++ ) {
+                    closedir(d);
+                    isdir = p->d_isdir;
+                    return p->d_name;
+                }
 
-bool FileSorter::get_file_is_dir(void)
-{
-    if ( this->file_index >= 0 && (size_t) this->file_index < this->file_count ) {
-        return this->file_info_array[this->file_index].is_dir;
-    } else {
-        return false;
-    }
-}
+                // process ON_IDLE every 20 iterations.
+                if ( itr++ % 20 ) {
+                    THEKERNEL->call_event(ON_IDLE);
+                }
+            }
 
-const char* FileSorter::next_file(void)
-{
-    if ( this->file_index == -1 || (this->file_index >= 0 && (size_t) this->file_index < (this->file_count - 1)) ) {
-        return this->file_info_array[++this->file_index].file_name;
-    } else {
-        return NULL;
-    }
-}
+            closedir(d);
+        }
 
-const char* FileSorter::first_file(void)
-{
-    if ( this->file_count > 0 ) {
-        this->file_index = 0;
-        return this->file_info_array[this->file_index].file_name;
-    } else {
-        return NULL;
-    }
-}
-
-file_info_t* FileSorter::file_at(int index)
-{
-    if ( index >= 0 && (size_t) index < this->file_count ) {
-        return &this->file_info_array[index];
-    } else {
         return NULL;
     }
 }
@@ -224,6 +254,11 @@ file_info_t* FileSorter::file_at(int index)
 bool FileSorter::has_error(void)
 {
     return this->error;
+}
+
+bool FileSorter::is_sorted(void)
+{
+    return this->sorted;
 }
 
 void FileSorter::set_filter_callback(__file_sort_filter_fn_t callback)
--- a/src/libs/FileSorter.cpp
+++ b/src/libs/FileSorter.cpp
@@ -10,6 +10,8 @@
 #include "FileStream.h"
 
 #include <new>
+#include <malloc.h>
+#include <string.h>
 
 
 FileSorter::FileSorter(string dir_path, __file_sort_filter_fn_t callback)
@@ -18,6 +20,7 @@ FileSorter::FileSorter(string dir_path, __file_sort_filter_fn_t callback)
     this->error = false;
     this->file_info_array = NULL;
     this->file_count = 0;
+    this->total_file_count = 0;
     this->file_index = -1;
     this->filter_callback = callback;
 
@@ -26,9 +29,7 @@ FileSorter::FileSorter(string dir_path, __file_sort_filter_fn_t callback)
 
 FileSorter::~FileSorter()
 {
-    if ( this->file_info_array ) {
-        delete[] this->file_info_array;
-    }
+    delete_file_info_array();
 }
 
 int FileSorter::compare_dir(file_info_t* file_a, file_info_t* file_b)
@@ -60,26 +61,37 @@ void FileSorter::sort_directory(void)
     this->file_index = -1;
 }
 
+void FileSorter::delete_file_info_array(void)
+{
+    if ( this->file_info_array != NULL) {
+        // file names were allocated previously with strdup - free them now.
+        for ( size_t i=0; i<this->file_count; i++ ) {
+            free(this->file_info_array[i].file_name);
+        }
+
+        // free the array.
+        delete[] this->file_info_array;
+    }
+}
+
 void FileSorter::open_directory(string dir_path)
 {
     // reset all member state variables.
     this->directory_path = dir_path;
     this->error = false;
     this->file_count = 0;
+    this->total_file_count = 0;
     this->file_index = -1;
-    if ( this->file_info_array ) {
-        delete[] this->file_info_array;
-        this->file_info_array = NULL;
-    }
+    delete_file_info_array();
 
     DIR *d;
     d = opendir(this->directory_path.c_str());
     if ( !(this->error = (d == NULL)) ) {
         // get a count of items in the folder.
-        size_t files_in_folder = count_files_in_folder(this->directory_path);
+        this->total_file_count = count_files_in_folder(this->directory_path);
 
         // allocate the array used for sorting.
-        this->file_info_array = new (std::nothrow) file_info_t[files_in_folder];
+        this->file_info_array = new (std::nothrow) file_info_t[this->total_file_count];
 
         // read the file info into the array.
         if ( !(this->error = (this->file_info_array == NULL)) ) {
@@ -87,11 +99,11 @@ void FileSorter::open_directory(string dir_path)
             file_info_t* array_entry;
             this->file_count = 0;
             array_entry = this->file_info_array;
-            while ( this->file_count < files_in_folder && (file_info = readdir(d)) != NULL ) {
+            while ( this->file_count < this->total_file_count && (file_info = readdir(d)) != NULL ) {
                 if ( (this->filter_callback == NULL) || this->filter_callback(file_info) ) {
                     array_entry->is_dir = file_info->d_isdir;
                     array_entry->file_size = file_info->d_fsize;
-                    std::strncpy(array_entry->file_name, file_info->d_name, FILE_SORTER_MAX_NAME_SIZE);
+                    array_entry->file_name = strdup(file_info->d_name);
                     this->file_count++;
                     array_entry++;
                 }
@@ -105,9 +117,14 @@ void FileSorter::open_directory(string dir_path)
     }
 }
 
-int FileSorter::get_file_count(void)
+size_t FileSorter::get_file_count(void)
 {
     return this->file_count;
+}
+
+size_t FileSorter::get_total_file_count(void)
+{
+    return this->total_file_count;
 }
 
 const char* FileSorter::get_file_name(void)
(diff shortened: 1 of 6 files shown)


wolfmanjm:
This is very nice, however it uses way too much memory and will crash and most systems.
The reason we do the slow method of opendir and never store the files is that we simply do not have enough memory to store an array of file names.
If this works for you you must be using very small amounts of memory. This would kill any one of the systems I use. I rarely have 2-4k memory available, and this could use much more than that depending on how many files are on the sdcard, and it is not acceptable to simply crash based on the number files on an sdcard.

wolfmanjm (changes requested):
If there were some way to limit how many files were loaded and sorted so as to not run out of memory it may be a nice feature.

michael-moore0:
Hey Jim.  I added checks to the allocation so that it does not crash the system on large directories. It will just report a read error.  I'm using an Azteeg X5 mini.  Not sure the specs, or how much free RAM.  I was able to read folders with several hundred files without issue.

wolfmanjm:
An overriding design goal of smoothie is to be stable, this uses a huge amount of memory and could crash mid print when opening the file menu or doing a ls /sd.
Caching the files is also a total waste of memory IMO. We are so short of memory during printing that even wasting 4bytes is not acceptable, and I have avoided doing a lot of great things as we are so low on memory.
IMO this is a very low priority and not worth the use of that memory. Others may disagree.
If we had so much memory available I'd prefer to increase the size of the block queue as that would allow more planning and better performance on deltas that have to segment gcode.

wolfmanjm:
Is there some way to make this optional, and not cache the files? and if it is active delete the cache unless the filescreen is open?
We have less than 4k of memory available while printing maybe a lot less. Out of memory is non deterministic in this system unfortunately and cannot be relied on, as fragmentation is a factor.
This copies upto a 32 byte filename from the Mbed library directory into memory 100 files could be 3.2k bytes,, that would not be a good use of the valuable memory.
The block queue is currently 32 which is really way too low, I'd prefer to use the memory (if it was available) to increase that to 64. It is much more important to have a big planner queue than it is to sort filenames... IMO

michael-moore0:
Hey Jim.  I completely understand the goal of stability.  I've been using the feature for several weeks and probably over 50 hours of printing without any issues.  This feature does not require any additional static RAM.  It allocates the RAM when needed (on 'ls' command or when viewing the file screen), and releases it when it is done.  If there's not enough memory available at the time, it just reports a read error, it doesn't crash the system.

wolfmanjm:
I have a suggestion as I really do like the potential of file sorting...

If it gets close to running out of memory can it revert to the old method of one file at a time unsorted which uses very little to no extra memory and will always display files, just maybe not sorted?

arthurwolf:
I wanted to point out that if this gets rejected due to lack of RAM ( which is common nowadays, can't cheat with hardware limitations ... ), it's much more likely to be accepted into Smoothie2, which will have a lot more RAM.

michael-moore0:
That should not be a problem.  I determine the file count for the directory, and then allocate the array.  If the allocation fails due to memory constraints, I set a flag:
`        // allocate the array used for sorting.
        this->file_info_array = new (std::nothrow) file_info_t[files_in_folder];

        // read the file info into the array.
        if ( !(this->error = (this->file_info_array == NULL)) ) {
       ...
`
I could add code to check this flag and revert to the non-sort approach when it's set.

wolfmanjm:
Ok if you can do that (and of course free up any memory used so far) I can accept it.
Although I am not 100% certain that the new() will return nullptr if the malloc fails, you may want to test that. We do not have exceptions enabled so I am not sure if std::nothrow works or not in the version of nanolib we use.

michael-moore0:
I have tested that the new( ) will return a nullptr when malloc fails.  I added about 1500 files to a directory to force an out of memory condition.  It fails safe, error message but no crash.

wolfmanjm:
did you try that when it was actually printing :) It can use variable amounts of memory while printing during processing of the gcode, but the block queue is mostly allocated by the time it starts a print. 
If we are really low on memory it can crash in the middle of a print. I like to keep at least 2k of headroom to ensure that does not happen. My system has 4k free after a print. that leaves about 2k of available memory not accounting for fragmentation of the free blocks.
I got a significant improvement in RAM management  in the recent rewrite by making sure the block queue is a fixed size once it is allocated. But gcode processing is still pretty variable.

src/libs/FileSorter.h
```diff
@@ -0,0 +1,55 @@
...
+#include <string>
+using std::string;
+
+
+#define FILE_SORTER_MAX_NAME_SIZE 32
+
+typedef bool (*__file_sort_filter_fn_t)(struct dirent*);
+
+typedef struct _file_info {
+    char file_name[FILE_SORTER_MAX_NAME_SIZE];
```
wolfmanjm: This is a little wastefull to allocate the entire possible file size, I'd prefer to see a pointer here and then strdup() as needed. it causes smaller blocks of memory to me malloced, and overall uses far less memory while active. Just remember to clean them up on release.

src/libs/FileSorter.cpp
```diff
@@ -0,0 +1,176 @@
...
+        if ( !(this->error = (this->file_info_array == NULL)) ) {
+            struct dirent* file_info;
+            file_info_t* array_entry;
+            this->file_count = 0;
+            array_entry = this->file_info_array;
+            while ( this->file_count < files_in_folder && (file_info = readdir(d)) != NULL ) {
+                if ( (this->filter_callback == NULL) || this->filter_callback(file_info) ) {
+                    array_entry->is_dir = file_info->d_isdir;
+                    array_entry->file_size = file_info->d_fsize;
+                    std::strncpy(array_entry->file_name, file_info->d_name, FILE_SORTER_MAX_NAME_SIZE);
```
wolfmanjm: This is problematic as if the source file name is > 32 then it will not be nul terminated making future use of this a potential problem if it is expected to be nul terminates (which it is I think).

src/libs/FileSorter.h
michael-moore0: Good point.  I'll change it accordingly.

src/libs/FileSorter.cpp
michael-moore0: That's definitely a bug.  Should have been:
`std::strncpy(array_entry->file_name, file_info->d_name, FILE_SORTER_MAX_NAME_SIZE - 1);`

I'm going to go with your previous suggestion and use `strdup` though.

wolfmanjm: FYI strncpy does not nul fill if the n limit is hit you have to do that yourself.

michael-moore0: Point taken, thanks!

michael-moore0:
I modified the code as you requested.  I added provision to revert to unsorted file list (both on FileScreen and 'ls' command) when memory is not available.  I revised the file_info_t struct to use a pointer rather than a char array.  I added a function to clean up the array when finished.  I tested both during print and idle (the FileScreen is unavailable during print though, so only 'ls' in print).  I was not able to break it.  I created a test folder with 1500 files with .gc extension to test.  It reverted these to unsorted both on FileScreen and 'ls'.   The only issue I encountered was that the print hung while the 'ls' was completing the command (on the 1500 file directory), but it would have done this with the trunk code as well.  Once the 'ls' was done, print resumed.

wolfmanjm:
ON_IDLE should be called during long operations, so I think you need to add a 

`THEKERNEL->call_event(ON_IDLE);`

at each iteration of the sorting loop if it is likely to take longer than a few milliseconds. Failure to do this will cause the watchdog to fire and/or the USB to disconnect.

wolfmanjm (approved):

michael-moore0:
The loop that caused the delay is not new code, but the original file read loop:
`
                
              while ((p = readdir(d)) != NULL) {
                stream->printf("%s", lc(string(p->d_name)).c_str());
                if(p->d_isdir) {
                    stream->printf("/");
                } else if(opts.find("-s", 0, 2) != string::npos) {
                    stream->printf(" %d", p->d_fsize);
                }
                stream->printf("\r\n");
            }
`
The only reason it was causing any delay is that I was reading a directory with 1500 files in it.  It caused printing to hang about five seconds.  I don't mind adding an ON_IDLE call to each loop iteration, but just so you know this is the trunk code, not mine.

wolfmanjm:
Yes I know it will pause that is because it  is hogging main loop which processes gcode so it stops reading gcode. But ON_IDLE cannot be paused that long it keeps things ticking like network and USB and the watchdog. SO the fact it is taking 5 seconds to porcess means ON_IDLE must be called in the inner loop, the rule is no loop can take more than about a few milliseconds without calling ON_IDLE.

src/libs/FileSorter.cpp
```diff
@@ -0,0 +1,193 @@
...
+        if ( !(this->error = (this->file_info_array == NULL)) ) {
+            struct dirent* file_info;
+            file_info_t* array_entry;
+            this->file_count = 0;
+            array_entry = this->file_info_array;
+            while ( this->file_count < this->total_file_count && (file_info = readdir(d)) != NULL ) {
+                if ( (this->filter_callback == NULL) || this->filter_callback(file_info) ) {
+                    array_entry->is_dir = file_info->d_isdir;
+                    array_entry->file_size = file_info->d_fsize;
+                    array_entry->file_name = strdup(file_info->d_name);
```
wolfmanjm: strdup can fail, so you need to test for nullptr here, and then clean up if it fails.

wolfmanjm (changes requested):

src/libs/FileSorter.cpp
michael-moore0: Sorry about that.  I added a NULL check.

wolfmanjm (approved):

wolfmanjm:
There appears to be a serious memory leak.

I added about 88 files with fairly long filenames to a subdirectoy called gcode.
Then I went to the file LCD menu  and scrolled up and down.
Then exited the menu back to the watch screen.
then checked memory with the mem command and it had lost about 5k of memory.
I repeated this and lost another 5k.
then I repeated it and it crashed hard.

wolfmanjm (changes requested):
Severe memory leak and hard crash when out of memory

michael-moore0:
Hey Jim.  That is completely unacceptable!  I'll see if I can reproduce it and get it resolved.

Based on the stack trace, it appears that the crash is due to the file name filter creating (or attempting to create) a new string instance:
`#9  0x00037f4c in FileScreen::filter_file (file_info=0x10006238) at modules/utils/panel/screens/FileScreen.cpp:202`
`#10 0x0000892a in FileSorter::open_directory (this=this@entry=0x10002d98, dir_path=...) at libs/FileSorter.cpp:104`

That would be this line here:
`string fn = lc(file_info->d_name);`

My theory is that it didn't crash before because the memory was tested up front.  That is, the `file_name` member of the `file_info_t` struct was a char array.  Now that it's a pointer, the `file_info_array` itself is relatively small.  This would mean that NULL checking after the array allocation here:
`this->file_info_array = new (std::nothrow) file_info_t[this->total_file_count];`
would not be as likely to fail.  The failure would occur later in the `strdup` (which is now NULL checked), or as the crash indicates, somewhere downstream.

One way to fix this might be to revert to the char array, thus checking memory up front and causing a safe failure, rather than a crash.  What are your thoughts on that?

As I see it, there are two problems:  the crash bug, and the memory leak.  I want to resolve the crash bug first, then address the memory leak.  I have a couple of questions on each issue.

For the crash bug:  Is there any means of determining the current heap availability before attempting to allocate?

For the memory leak bug:  how does the garbage collector work in nanolib?  Could it be that it had not run between each time you navigated to the gcode folder?  Is there a way to force garbage collection?  The destructor calls `delete_file_info_array` which frees all memory that was allocated:


```
void FileSorter::delete_file_info_array(void)
{
    if ( this->file_info_array != NULL) {
        // file names were allocated previously with strdup - free them now.
        for ( size_t i=0; i<this->file_count; i++ ) {
            free(this->file_info_array[i].file_name);
        }

        // free the array.
        delete[] this->file_info_array;
    }
}
```

Is `free` the correct means to release memory allocated by `strdup` under nanolib?  The other possibility is that the destructor was never called.  When does `FileScreen::on_exit()` get called?  This is where the `FileSorter` is destroyed.

wolfmanjm:
reverting to a fixed size string is not acceptable IMO.
There is no  garbage collecting in C++ (this isn't Java ;) Memory is freed as soon as free is called.
free() is the correct way to free strdup().
on_exit is called when the menu is left and a new menu is called up.
It is indeed crashing when the string is being allocated as there is no testing for out of memory there. (This was my worry at first about the inability to safely check for memory exhaustion).
Ideally this needs to check there is at least 1k of memory available at all times. Unfortunately there is no easy way to do that. (look at the mem command in SimpleShell.cpp) you have to traverse the malloc heap and that is really not a good thing to have to do.

michael-moore0:
Sorry, I work primarily in vanilla C, so I'm not as knowledgable on the ins-and-outs of C++.  I thought I recalled something about GC (maybe in C++2011?).  I was just trying to figure why it was not releasing memory.  If there's no GC, seems it would be pretty much deterministic.  You exit the FileScreen, `on_exit` is called, the `FileSorter` is destroyed and the memory is freed.  It seems there are only a few possibilities, either the destructor is not being called, or `free` is not freeing.  That's why I was asking about `strdup` and `free`.

It seems a bit of a hack, but what about allocating a dummy block of 1K (in addition to what is actually needed)?  Test if it fails then error out.  Either way, release it immediately after testing whether it succeeded.  That would ensure that there is a 1K buffer and should fix the crash bug.  The concern I have would be fragmentation.  Perhaps a block that large would not have much effect on fragmentation?  I'll consider the problem in depth when I get a chance.  In the mean time, what are your thoughts on that?

michael-moore0:
On another note,  would you please give me some debugging advice?  I know of the `mem` command for checking memory, but I need a few more tools to track this down.  I'd like to do a trace.  Perhaps just dump a log at several points to see what's hit and what's not.  How would I go about that?  If I connect to the shell, what's the best way to dump a log there?  Also, how did you get the stack trace of the crash you posted earlier?

wolfmanjm:
http://smoothieware.org/mri-debugging

triffid:
If smoothie runs out of memory, the downward-growing stack collides with
the upward-growing heap and data in both can get corrupted.

We do have canaries to detect this but they're not checked all the time;
it's entirely possible for free to fail if malloc's data structures are
damaged by heap/stack collision, and also very likely for extremely strange
stuff to occur if you write into an area that's now occupied by stack.

There's no OS kernel to separate the memory regions and fail
semi-gracefully in out of memory condition. On microcontrollers, out of
memory is often spectacularly fatal.

Even checking the stack pointer vs heap head doesn't help, imagine the case
where you malloc and almost but not quite touch it, then call a method on
the object you just created..

That's why Jim is strongly resistant to anything that uses lots of memory,
we're already borderline on usage in several usage scenarios and little
things can push smoothie over the edge - and the best case outcome there is
it crashes hard.. Worst case it destroys someone's hardware due to silent
corruption of internal state.

wolfmanjm:
using all the memory is simply not going to work... there are many calls made while the filescreen is open that will require memory, and most of those cannot deal with an out-of-memory issue (which is exactly what this crash is). Every time a std::string is used it allocates memory and has no mechanism to fail. Displaying the line on the lcd scree will need to use memory.
Allocating an extra 1k then releasing it once you have allocated all the file names may work but is a very ugly hack, and who is to say 1k is enough?
I cannot see where the memory leak is either, but it is repeatable, just click to go into gcode folder, then scroll up and down then exit back up two folders to exit the files screen.
I used about 80 files but each file had very long file names probably an average file length of 32 characters, but many are  40-50 character long filenames.

wolfmanjm:
One reason I am against a fixed size filename and truncation is that `ls /sd` should be able to display the full filename and not a truncated one.
If we did go the route of fixed name length then 20 and not 32 would be the case as the LCD can only display 20 characters. However that would be a retrograde step for the ls /sd command. And even then it would still run into the issue of using nearly all the memory available and having something downstream crash because its memory allocation failed.
I don't have any good workarounds for these issues I am afraid.
It all seems a lot of work simply to sort the file names :)

michael-moore0:
Hey Jim and Michael.  As you (Jim) said, it does seem a lot of work simply to sort the file names.  I'd let it go but for the fact that I have a really hard time resisting a challenge.  I completely understand all the arguments and reasons that you have both presented.  I typically work with MCUs that have resource constraints far beyond the hardware targeted by Smoothieware.  I know the pitfalls of stack overflows and memory overruns on embedded devices.  In fact, most of the devices I typically work with wont even allow for a heap at all.  Memory management is done by careful static allocation.  One of the comments you (Michael) made gives me concern:

>If smoothie runs out of memory, the downward-growing stack collides with
the upward-growing heap and data in both can get corrupted.

All of the MCUs that I've worked with have toolchain support for specifying and enforcing a heap size.  Is this not the case with the Smoothieware project?  Also, most of the toolchains/debuggers for the MCUs I have worked with provide some means of stack analysis.  These allow you to determine how deep the stack goes in any number of scenarios.  If the heap size is specified and enforced, and you have an idea how deep the stack goes in the worst case scenario, you should never encounter a heap-stack collision.  Worst case, you should have an allocation failure (which I'm not underestimating, by the way - I know that can be just as bad as memory corruption, but there should be ways to mitigate).

src/libs/FileSorter.cpp
```diff
@@ -0,0 +1,204 @@
...
+void FileSorter::open_directory(string dir_path)
+{
+    // reset all member state variables.
+    this->directory_path = dir_path;
+    this->error = false;
+    this->file_count = 0;
+    this->total_file_count = 0;
+    this->file_index = -1;
+    delete_file_info_array();
+
```
johnkingsley: delete_file_info_array() is called after setting file_count to 0.
This is a memory leak, I think.

johnkingsley: Hi Michael,

Here are some general comments/suggestions about your changes.

1. Avoid putting FileSorter on the heap.  (i.e. "new"ing it).  In an embedded environment I like to use the heap when only necessary.  In this case I think you can avoid it. I can give more details if you want.

2. Put the error handling (when malloc fails) inside FileSorter, not outside.
(i.e. "if the file sorter failed (likely due to memory constraints), do the filename lookup the old fashioned way")
This avoids code duplication (in FileScreen and SimpleShell) and will also make the code cleaner and easier to understand.

3. In FileSorter, you have the member variable file_index.  This adds state to the class. As far as I can tell, this is only used in SimpleShell.  I would remove this state since it complicates things without any advantage. Instead, get the total number of directory entries in SimpleShell and iterate through them using file_at().

I hope this helps!

John

src/modules/utils/simpleshell/SimpleShell.cpp
```diff
@@ -298,23 +299,42 @@ void SimpleShell::ls_command( string parameters, StreamOutput *stream )
...
-    DIR *d;
-    struct dirent *p;
-    d = opendir(path.c_str());
-    if (d != NULL) {
-        while ((p = readdir(d)) != NULL) {
-            stream->printf("%s", lc(string(p->d_name)).c_str());
-            if(p->d_isdir) {
+    // attempt to use the file sorter to sort alphabetically.
+    FileSorter* files = new FileSorter(path, NULL);
+    if ( !files->has_error() ) {
```
johnkingsley: Need to check if `files` is NULL (heap full)

src/modules/utils/panel/screens/FileScreen.cpp
```diff
@@ -185,3 +187,22 @@ void FileScreen::play(const char *path)
     message.stream = &(StreamOutput::NullStream);
     THEKERNEL->call_event(ON_CONSOLE_LINE_RECEIVED, &message );
 }
+
+void FileScreen::delete_file_sorter(void)
+{
+    if ( this->file_sorter ) {
```
johnkingsley: There is no need to check if `file_sorter` is non NULL, since C++ guarantees that `delete` of a NULL is ok.

src/libs/FileSorter.cpp
michael-moore0: Hey John,

I pushed changes to fix the memory leak and address your review notes below.  I've not yet addressed items 1-3 in this comment, but I'm happy to do it when I have opportunity.  I didn't originally consider items 2 and 3 because I really had not considered submitting these changes at all.  I made the modification for my own use, and later thought others might find it helpful as well.  Items 2 and 3 definitely make sense in the context of a wider audience.  I'll address them when I have some time and push them up.  As for item 1, my reasons for putting it on the heap was that it is an 'if-available' feature.  If there's enough memory to do the sort, then do it.  Otherwise, no sorting, so I wanted to completely clean up resources when done.  I suppose I can create it as a member variable and add a method to the `FileSorter` to clean up resources on exit.  I'm not certain that would get it off the heap though.  Wouldn't the fact that the fact that the `FileScreen` itself is created on the heap cause all its members to be on the heap as well:
```
this->file_screen    = (new FileScreen()            )->set_parent(this);
```

wolfmanjm:
Ok that will probably fix the memory leak (well caught @johnkingsley ).
However this does nothing to solve the out-of-memory crash that can happen downstream when this sucks every last byte of memory from the system.
Any ideas on solving that? Without a solution to that I think this will be a liability.

src/libs/FileSorter.cpp
wolfmanjm: I disagree with stuff being on the stack, The stack is an extremely limited resource and does not grow (it is fixed size), so it can fail with no errors at all. Whereas when allocated on the heap there is at least a chance of detecting it.

michael-moore0:
Hey Jim, I know that it doesn't resolve the main problem.  I just wanted to get those items addressed as John had pointed them out and it was a fairly quick fix.  I'm not ready to give up just yet, but I've not had opportunity to consider the matter in depth yet.  If you have opportunity, would you please read and comment on my comment from this morning.  It would have a bearing on any potential solution.  Thanks!

triffid:
On 10 October 2016 at 22:30, michael-moore0 <notifications@github.com>
wrote:

> One of the comments you (Michael) made gives me concern:
>
> If smoothie runs out of memory, the downward-growing stack collides with
> the upward-growing heap and data in both can get corrupted.
>
> All of the MCUs that I've worked with have toolchain support for
> specifying and enforcing a heap size. Is this not the case with the
> Smoothieware project?
>
There's a heap top in the linker script, presumably _sbrk will check it..
I'm not sure if we ever check for stack overflow however, and even less
sure what we'd do with a stack overflow condition.

> Also, most of the toolchains/debuggers for the MCUs I have worked with
> provide some means of stack analysis. These allow you to determine how deep
> the stack goes in any number of scenarios. If the heap size is specified
> and enforced, and you have an idea how deep the stack goes in the worst
> case scenario, you should never encounter a heap-stack collision.
>
There's various memory-related commands in .gdbinit which you can use
during debugbreak (firmware crash or press ISP button) and connect gdb to
the debug port. I believe the interstitial space between heap and stack is
filled with 0xdeadbeef canary so you can check if they've ever collided in
the past by dumping memory.

I don't think we have any analysis tool that can predict worst-case stack
usage, keep in mind that on the NXP1768, high priority interrupts can
pre-empt lower priority ones which makes stack usage projections rather
tricky.. you'd have to find the deepest stack from user code, then add the
stack usage of all interrupts of different priority levels, using the
highest usage interrupt when several have the same priority

If you feel like doing such an analysis, we'd love to see the results :)

johnkingsley:
Here's some ideas on how to avoid using tons of heap:

1. Keep track of how much heap you're using (essentially this is proportional to #files + each file name size) and if you hit a certain limit (say, 1k) then bail and use the old way (no sorting).

2. Once you figure out the sorting, delete all the allocated objects and instead alloc a sorted list of indexes.  So `file_at(i)` becomes `file_at(sorted_idx(i))`   So the heap required is proportional to #files and not proportional to the file name lengths.

3. If you limit the #files to <=256 then you can use one byte in 2 above (per file) for the index. So this limits the total heap usage to <=256 bytes.  If there are more than 256 files, use the old way (no sorting).  If you think 256 files is too small, then use 16 bit integer and that will allow 65536 files and still keep your heap usage to 512 bytes or less.

John

triffid:
Smoothie configures lots of stuff at runtime from the config file,
new/malloc is rather hard to avoid with runtime configuration.

Indexes are tricky because to get each actual filename you have to read the
sdcard if the fat entry is in another sector, so sorted ls could be ultra
slow if the directory entries are fragmented

Perhaps you'd be better off sorting the entries in one or two fat blocks
then writing them back on each ls , so the disk order would eventually
become sorted? That then opens up problems if you ls while it's mounted on
the host however... Tricksy indeed, MTP might be better than USB disk if
someone wants a crack at writing it
On 11 Oct 2016 22:55, "John Kingsley" <notifications@github.com> wrote:

Here's some ideas on how to avoid using tons of heap:

   1.

   Keep track of how much heap you're using (essentially this is
   proportional to #files + each file name size) and if you hit a certain
   limit (say, 1k) then bail and use the old way (no sorting).
   2.

   Once you figure out the sorting, delete all the allocated objects and
   instead alloc a sorted list of indexes. So file_at(i) becomes
   file_at(sorted_idx(i)) So the heap required is proportional to #files
   and not proportional to the file name lengths.
   3.

   If you limit the #files to <=256 then you can use one byte in 2 above
   (per file) for the index. So this limits the total heap usage to <=256
   bytes. If there are more than 256 files, use the old way (no sorting). If
   you think 256 files is too small, then use 16 bit integer and that will
   allow 65536 files and still keep your heap usage to 512 bytes or less.

John

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/Smoothieware/Smoothieware/pull/1035#issuecomment-252941684>,
or mute the thread

michael-moore0:
Hey Jim.  I updated to include the file_info_t size in the required memory calculation.  I had previously defined a padding of 1K:

```#define FILE_NAME_ARRAY_MIN_PADDING   1024```

I can modify the size if that is insufficient.

**Comment by michael-moore0** [comment]
Sorry, It appeared to be failing the CI build because of it.  It built fine on my local system, but CI threw an error complaining it couldn't find the file.  Do you want me to revert it?

**Comment by wolfmanjm** [comment]
FYI you need to merge with current edge as I fixed a potential issue with ON_IDLE being called during a panel refresh.

**Comment by michael-moore0** [comment]
Has that been pushed?  I pulled the latest before I pushed, and again just now, and the latest commit is October 4:

>commit aea3d2cba38fe5a51f4aa2ba8184c6891e21fb93
Author: Jim Morris <morris@wolfman.com>
Date:   Tue Oct 4 21:37:42 2016 -0700

**Comment by wolfmanjm** [comment]
Yes https://github.com/Smoothieware/Smoothieware/commit/bce057ef0617fbec14cb1aa4b3d0cb842163d4ba

File: src/libs/FileSorter.cpp (hunk 4 above)
Comment (michael-moore0) [inline src/libs/FileSorter.cpp]: Hey John.  I'm going to have to go with @wolfmanjm regarding item 1.  I understand the arguments against dynamic allocation in embedded systems.  In my estimation, it is dependent on the nature of the system and the available resources.  Memory management in most of the systems that I work on is done using static allocation - the memory map is defined at compile time.  However, the Smoothieware project from what I know if of it appears to be architected to rely heavily on dynamic allocation.  The second section of the Coding Standards (http://smoothieware.org/coding-standards) specifies pointers for class references (implying heap allocation):

> Use class pointers in classes and use Class xxx in header to forward define it

I did address items 2 and 3 and pushed the changes.  Please let me know if I have sufficiently dealt with your concerns.  Thanks!

**Comment by triffid** [comment]
Nothing stopping you taking a pointer to a statically allocated class,
that's what operator& is for ;)

**Comment by michael-moore0** [comment]
I suppose that's my vanilla C bias showing :).  Thanks for the reminder.   Still I have seen heavy reliance on dynamic allocations.  The FileScreen itself is allocated that way (unless that's another C++ subtlety that I'm missing).  Regardless, I'm quite willing to change the implementation if required.  Thanks!

**Comment by wolfmanjm** [comment]
No. the final word is it remains dynamically allocated. It has to be. You don't know the size before hand so can't be a member array, and it cannot go on the stack.
DO NOT put things on the stack that are large it WILL crash the system.

File: src/libs/FileSorter.h [hunk 7]
```diff
@@ -47,11 +45,11 @@ class FileSorter {
 
     string directory_path;
     bool error;
+    bool sorted;
```
wolfmanjm: with bools we usually do this...
```
struct {
  bool error:1;
  bool sorted:1;
};
```
and move it to the end of the class. This saves memory.

johnkingsley:
I just reviewed your changes @michael-moore0 
Looks good to me!

wolfmanjm:
closing as it is now in conflict with current edge. Also some requested changes appear to have never been made.

nebbian:
Would it be possible to sort based on modification time instead of file name?  I know that Marlin does this, and I find it really useful.  Modification time can (I think) fit into 4 bytes, instead of the many bytes required for file names.  This would cut down on memory usage tremendously.

PR #1038, opened 2016-10-12
modified timestamp format to HH:MM:SS.

michael-moore0:
Hopefully this PR will be a bit more strait forward than my last one.  I have a hard time thinking in seconds.  Whenever I would check the watch screen elapsed time, I would have to mentally do the calculations to convert the seconds into hours and minutes.  I decided to modify the watch screen and 'progress' command to display time in HH:MM:SS format.  I thought this might also be useful for others, so I'm submitting this PR in the hopes that it may be helpful.

### DIFF
--- a/src/modules/utils/panel/screens/3dprinter/WatchScreen.cpp
+++ b/src/modules/utils/panel/screens/3dprinter/WatchScreen.cpp
@@ -250,7 +250,7 @@ void WatchScreen::display_menu_line(uint16_t line)
             break;
         }
         case 1: THEPANEL->lcd->printf("X%4d Y%4d Z%7.2f", (int)round(this->pos[0]), (int)round(this->pos[1]), this->pos[2]); break;
-        case 2: THEPANEL->lcd->printf("%3d%% %2lu:%02lu %3u%% sd", this->current_speed, this->elapsed_time / 60, this->elapsed_time % 60, this->sd_pcnt_played); break;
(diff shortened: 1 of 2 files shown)


src/modules/utils/panel/screens/3dprinter/WatchScreen.cpp
```diff
@@ -250,7 +250,7 @@ void WatchScreen::display_menu_line(uint16_t line)
             break;
         }
         case 1: THEPANEL->lcd->printf("X%4d Y%4d Z%7.2f", (int)round(this->pos[0]), (int)round(this->pos[1]), this->pos[2]); break;
-        case 2: THEPANEL->lcd->printf("%3d%% %2lu:%02lu %3u%% sd", this->current_speed, this->elapsed_time / 60, this->elapsed_time % 60, this->sd_pcnt_played); break;
+		case 2: THEPANEL->lcd->printf("%3d%% %02lu:%02lu:%02lu %3u%% sd", this->current_speed, this->elapsed_time / 3600, (this->elapsed_time % 3600) / 60, this->elapsed_time % 60, this->sd_pcnt_played); break;
```
wolfmanjm: incorrect whitespace, violates the whitespace policy.

michael-moore0: Sorry, I think the white space was a copy/paste error.  I thought I had caught all of them.  On mine, I actually left off the 'sd' at the end and spaced in between.

If you're interested in keeping the formatting, I'm willing to change as necessary.  Otherwise, I'll just let it go.  Please let me know.  Thanks!

wolfmanjm: yea the SD can go, so long as it fits in 20 characters it is better than just the mins:secs I guess.

PR #1039, opened 2016-10-12
implemented the 'mkdir' command on the shell.

michael-moore0:
@arthurwolf requested that I implement the `mkdir` command for the Smoothieware shell.  Presented for your consideration is a basic `mkdir` implementation on the `SimpleShell`.  This function takes a single parameter, the path of the directory to be created.  It may be an absolute or relative path.  There is no `-p` or other special arguments/flags.  If the path to the folder to be created does not exist, the command will fail.


arthurwolf:
Awesome thanks ! I hope we get this into a shape to be merged soon !

arthurwolf:
Yay \o/

lautr3k:
Thanks a lot ! this is so cool!

PR #1047, opened 2016-10-20
added use-cases to the runaway temperature monitoring.

michael-moore0:
At the suggestion of @arthurwolf, I have separated runaway temperature monitoring into a class.  The purpose of this pull request is to introduce handling two new use-cases to the temperature monitoring.

First, the use case where the temperature has to be lowered in order to reach the target.  I often encounter this scenario as part of my normal workflow.  I will typically set the temperature lower than printing temperature in order to do a cold pull before changing filament (PLA for example would be 85C to cold pull, as opposed to 190C to print). Many times I do this whilst the hotend is still hot, so the temperature has to cool to reach the target. Sometimes I get distracted with other things and am not closely monitoring, waiting for the temperature to be reached.

Second, the use case where the temperature takes too long to get to the target.  An example of this would be when I print ABS I set the bed temperature to 95C.  I have a 12V system which takes around 15 minutes to get the bed from room temperature to 95C. Since the runaway control is not granular, I would need to set the runaway to 15 minutes universally so it would not get a false HALT due to the bed taking too long to heat. If I let my hotend heater run un-controlled for 15 minutes, it would likely start a fire.

The code for this pull request is primarily derived from https://github.com/Smoothieware/Smoothieware/pull/1045.  I also utilized some changes from https://github.com/Smoothieware/Smoothieware/pull/1046.  I have optimized memory usage so that this approach does not use any additional memory.  The IVARs for the `TemperatureMonitor` class have been optimized to fit within a single 32-bit word.  The IVAR for the monitor within the `TemperatureControl` class is a pointer, which is 32-bits.  So there are 64 bits of memory used, which is the same amount used in the original implementation.

### DIFF
--- a/src/modules/tools/temperaturecontrol/TemperatureControl.cpp
+++ b/src/modules/tools/temperaturecontrol/TemperatureControl.cpp
@@ -140,6 +140,7 @@ void TemperatureControl::load_config()
     this->designator          = THEKERNEL->config->value(temperature_control_checksum, this->name_checksum, designator_checksum)->by_default(string("T"))->as_string();
 
     // Runaway parameters
+    delete this->temperature_monitor;
     uint8_t range = THEKERNEL->config->value(temperature_control_checksum, this->name_checksum, runaway_range_checksum)->by_default(0)->as_number();
     uint16_t timeout = THEKERNEL->config->value(temperature_control_checksum, this->name_checksum, runaway_heating_timeout_checksum)->by_default(0)->as_number();
     this->temperature_monitor = new TemperatureMonitor(range, timeout);
--- a/src/modules/tools/temperaturecontrol/TemperatureControl.cpp
+++ b/src/modules/tools/temperaturecontrol/TemperatureControl.cpp
@@ -7,13 +7,13 @@
 
 // TODO : THIS FILE IS LAME, MUST BE MADE MUCH BETTER
 
-#include "libs/Module.h"
-#include "libs/Kernel.h"
+#include "Module.h"
+#include "Kernel.h"
 #include <math.h>
 #include "TemperatureControl.h"
 #include "TemperatureControlPool.h"
-#include "libs/Pin.h"
-#include "modules/robot/Conveyor.h"
+#include "Pin.h"
+#include "Conveyor.h"
 #include "PublicDataRequest.h"
 
 #include "PublicData.h"
@@ -27,6 +27,8 @@
 #include "PID_Autotuner.h"
 #include "SerialMessage.h"
 #include "utils.h"
+
+#include "TemperatureMonitor.h"
 
 // Temp sensor implementations:
 #include "Thermistor.h"
@@ -75,6 +77,7 @@ TemperatureControl::TemperatureControl(uint16_t name, int index)
     temp_violated= false;
     sensor= nullptr;
     readonly= false;
+    temperature_monitor= NULL;
 }
 
 TemperatureControl::~TemperatureControl()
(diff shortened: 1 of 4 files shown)


wolfmanjm:
This does not provide some of the extra robustness I added to my version.
Kudos for getting the memory footprint down though.
When maintaining temp one false reading could trigger a halt whereas in mine it needs to be for a few seconds.

src/modules/tools/temperaturecontrol/TemperatureControl.cpp
```diff
@@ -137,8 +140,12 @@ void TemperatureControl::load_config()
...
 
     // Runaway parameters
-    this->runaway_range           = THEKERNEL->config->value(temperature_control_checksum, this->name_checksum, runaway_range_checksum)->by_default(0)->as_number();
-    this->runaway_heating_timeout = THEKERNEL->config->value(temperature_control_checksum, this->name_checksum, runaway_heating_timeout_checksum)->by_default(0)->as_number();
+    delete this->temperature_monitor;
+    uint32_t range = THEKERNEL->config->value(temperature_control_checksum, this->name_checksum, runaway_range_checksum)->by_default(0)->as_number();
+    if(range > 63) range = 63;
+    uint32_t timeout = THEKERNEL->config->value(temperature_control_checksum, this->name_checksum, runaway_heating_timeout_checksum)->by_default(0)->as_number();
+    if(timeout > 511) timeout = 511;
+    this->temperature_monitor = new TemperatureMonitor(range, timeout);
```
wolfmanjm: why create this at all if runaway is disabled?

wolfmanjm:
To be honest I prefer my version (surprise!)
In set_desired_temp we already know what the previous temp was and what the new requested tempo is, we could have set the state there for it going up or down. And do this without an extra pointer to a new class or keeping the last temperature around.
Overall I think this is less robust than my PR.

wolfmanjm:
Also why is breaking this out into a separate class a good design decision? It uses more code space and IMO does not simplify anything.

michael-moore0:
Hey Jim.

>When maintaining temp one false reading could trigger a halt whereas in mine it needs to be for a few seconds.

This is not correct.  On heating or cooling state, the temperature has to stall for the timeout length before it will HALT.

wolfmanjm:
I think yo are wrong read my comment more carefully I was not refering to the heating up or cooling down states.

michael-moore0:
Granted.  I didn't modify the maintaining state part of the original code and had apparently overlooked changes you made to that state case in your mods.  I'll incorporate your changes.

michael-moore0:
>In set_desired_temp we already know what the previous temp was and what the new requested tempo is, we could have set the state there for it going up or down. And do this without an extra pointer to a new class or keeping the last temperature around.

The primary reason for keeping the last temp is to handle the second use case I mentioned in my original comment in this PR.  With your code, you get a max timeout of 8:32.  My bed takes 15 minutes or more to get to 95C for ABS.  Besides, even if the timeout were long enough, it would be too long to run the hot end heater uncontrolled without a fire risk.

michael-moore0:
>Also why is breaking this out into a separate class a good design decision? It uses more code space and IMO does not simplify anything.

Primarily because @arthurwolf suggested that I do that (in the other PR) and nobody countered.  I suppose there is no major advantage other than slightly cleaning the temperature control code and keeping a separation between control and monitoring.

wolfmanjm:
Here is the thing.. I really hate to reject work you did based on Arthur asking you to do something, but he knew I was working on fixing this, and I prefer my implementation. I still need to add a robust lower temp scenario which was a new requirement.
Adding a class is not a good solution or design goal. I am not sure why it was suggested, and as I had already closed that issue and submitted a PR of my own I presumed the issue was closed.

wolfmanjm:
I am also looking at removing all this code and just using the temperature switch code to replace it. That already does 90% of what is needed to monitor the temps, it just needs to add a halt action, and some timeouts. If nothing else it should be used to halt if the temp goes too high. I am still looking at that. It has a better state control for temps going up or going down and edge triggers etc.

wolfmanjm:
I'd like to say that this is not a bad implementation and I think you did a good job. However given the choice between my version and yours I'll pick mine, not because it is significantly better, it is different, but achieves the same goal, and as I have to maintain the code I am obviously going to prefer my code over someone elses. It is unusual to be in a position where someone else submits a PR to fix the same issue at the same time especially when there is already a PR to fix a given issue.
I have plans to improve my PR and increase the timeout periods without increasing the memory usage too which would be harder to do in yours. I may even take some ideas from yours :)

wolfmanjm:
I also plan on having a separate heat up and cool down timeout which should solve your issue with slow cool down.

wolfmanjm:
I propose PR #1046 as an alternative to this. It uses no extra memory and I believe handles the same use cases both heating up timeout and independent cooling down timeout.

wolfmanjm:
> Since the runaway control is not granular, I would need to set the runaway to 15 minutes universally so it would not get a false HALT due to the bed taking too long to heat. If I let my hotend heater run un-controlled for 15 minutes, it would likely start a fire.

This is wrong, each temperature control has its own timeout so hotend and bed have different timeouts.

wolfmanjm:
This fundamentally changes what the timeout value is measuring, and would not be backward compatible with what people already have set.
The current setting is the total timeout for how long it takes to reach a certain temp, whereas this change it is a timeout for how long it has stopped trending towards a certain temp (either up or down). This value would need to be a much smaller value to be safe. It also no longer protects the system from the original failure mode it was designed to protect against which is the thermistor falling out. In that case if the thermistor was close to the heater block i would still trend higher thus resetting the timeout potential indefinitely until the system caught fire. The original method is far safer for this failure mode.

PR #1050, opened 2016-10-25
modified watch screen to display extruder when printing.

michael-moore0:
I have found that when a print is in progress, I have more interest in the Extruder position than the X and Y positions (which are changing too rapidly).  I modified the watch screen to display the X, Y and Z positions when the printer is idle, and the Extruder and Z positions when printing.

I present this for your consideration, in case anyone else would find it useful.

### DIFF
--- a/src/modules/utils/panel/screens/3dprinter/WatchScreen.cpp
+++ b/src/modules/utils/panel/screens/3dprinter/WatchScreen.cpp
@@ -22,6 +22,7 @@
 #include "SwitchPublicAccess.h"
 #include "checksumm.h"
 #include "TemperatureControlPool.h"
+#include "ExtruderPublicAccess.h"
 
 
 #include <math.h>
@@ -29,6 +30,8 @@
 #include <string>
 #include <stdio.h>
 #include <algorithm>
+
+#define extruder_checksum CHECKSUM("extruder")
 
 using namespace std;
 static const uint8_t icons[] = { // 16x80 - he1, he2, he3, bed, fan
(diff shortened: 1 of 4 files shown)


wolfmanjm:
This is not my personal preference, I prefer to see XYZ. 
However if there was a config option to select the display then I think this maybe a preference for some people. I think you could use a boolean and config option to select the display you prefer (use it in a packed struct so it uses no extra memory :)

src/modules/utils/panel/screens/3dprinter/WatchScreen.cpp
```diff
@@ -249,7 +252,17 @@ void WatchScreen::display_menu_line(uint16_t line)
             }
             break;
         }
-        case 1: THEPANEL->lcd->printf("X%4d Y%4d Z%7.2f", (int)round(this->pos[0]), (int)round(this->pos[1]), this->pos[2]); break;
+        case 1:
+            if ( THEPANEL->is_playing() ) {
+                pad_extruder_t rd;
+                float extruder_pos = (PublicData::get_value( extruder_checksum, (void *)&rd )) ? rd.current_position : -1;
```
michael-moore0: Sure, that would not be a problem.

src/modules/utils/panel/screens/3dprinter/WatchScreen.cpp
```diff
@@ -249,7 +252,17 @@ void WatchScreen::display_menu_line(uint16_t line)
             }
             break;
         }
-        case 1: THEPANEL->lcd->printf("X%4d Y%4d Z%7.2f", (int)round(this->pos[0]), (int)round(this->pos[1]), this->pos[2]); break;
+        case 1:
+            if ( THEPANEL->is_playing() ) {
```
michael-moore0: I can do that.  Two questions:
1.  I was thinking to add the option as a sub-item under `panel` (such as `panel.display_extruder`), since if you have no panel, displaying the extruder value is irrelevant.  Would that be acceptable?
2.  Should I add that config as an example to any of the sample config files?

wolfmanjm:
panel.display_extruder is fine and you would then set the associated bit in the watch screen. as only panel reads the config file.
Yea it could be added as an example panel setting in the snippets, I was thinking of moving panel out to the snippetes folder as that is where I use mine from. But it could also be added as a commented out panel setting, but mostly it just needs to be documented in the wiki with an explanation of what it does.

michael-moore0:
@wolfmanjm I added a `panel.config` to the Snippets.  I just pulled the panel section from the Azteeg X5 Mini v3 sample config and appended the new option.  I put the flag inside the `Panel.cpp`, as that is where the config is read.  It seemed more straight forward that way, but let me know if you'd rather the option flag be in the `WatchScreen`.

wolfmanjm:
Nice and saves some memory :) you are getting the idea now ;)

PR #1052, opened 2016-10-27
Feature/temperature range

JesperKrogPoulsen:
Added support for specifying ranges for the set_and_wait m-code. 

Example:
M109 S100 : Waits until temperature is > 100 as before this change
M109 S0 : Turns off heater as before this change
M109 R50 : Turns off heater and waits until the temperature is below 50
M109 S50 R100 : Waits until temperature is between 50 and 100, either warming or cooling, depending on current temperature

Use cases:
I have added PSU control to my config and can turn off the printer with M81. I would like to do that at the end of a gcode file, to turn off the printer when the print is done. However, I do not want to turn off the printer when the temperature is too high, as heat might creep up in the cold end and jam the filament. With this change, I can add the following lines to my gcode file:

M109 R50
M81

to let the printer cool off the hotend to below 50, before turning off the power.

Another use case is multi-filament extruders/hotends. There are hotends that accept multiple filaments and change filament by retracting one filament and extruding the other. Since there is only one hotend, the filaments has to be of the same type. With this change, it is possible to let the print-head cool down before continuing, allowing for filaments of different types and temperature ranges. 
For example, if you have 3 filaments that works in the 200-210 range, 220-230 range and one that works in the 240-250 range. Use M109 S220 R230 to set the temperature when changing to the second filament, and the printer will pause until inside that range, no matter which of the other two filaments were used before.

This is untested, though, as I do not have such hotend/extruder on my printer.


JesperKrogPoulsen:
One more comment. The idea to use of 'R' to indicate range comes from http://reprap.org/wiki/G-code#M109:_Set_Extruder_Temperature_and_Wait

wolfmanjm:
This is what the temperature switch was used for http://smoothieware.org/temperatureswitch there is even an example of how to do this.
I am not sure we need two methods to do the same thing.

wolfmanjm:
Ok This may solve a different use case though, but needs to be tested for that.
I prefer the TemperatureSwitch for turning off the PSU when it cools down as it does not lock the system up while you wait for it. I do not like M109 in general as it does that. With the temperature switch you can arm and disarm it and continue jogging while it monitors the temperature.

src/modules/tools/temperaturecontrol/TemperatureControl.cpp
```diff
@@ -347,17 +355,25 @@ void TemperatureControl::on_gcode_received(void *argument)
                         }
 
                         this->waiting = true; // on_second_tick will announce temps
-                        while ( get_temperature() < target_temperature ) {
+                        float t = get_temperature();
+                        while ( t < target_temperature || (r != UNDEFINED && t > r) ) {
```
wolfmanjm: This is not very robust, there are many situations where the user can set values here that will cause it to wait indefinitely. I would add some sanity checks to make sure that not only is r < t but also that it is at least a few degrees less.

JesperKrogPoulsen:
Hm. I agree on your view on the non-blocking feature of the switch. 

I hadn't considered someone setting v=r. Good catch. I suggest changing line 341 to something along the line of `if (r-v < 3)`. (Sorry, don't know how to quote properly from the pull request, like you did). This gives a 3 degrees window. Also, after some thought, I want to change 342 to `return`, instead of invalidating the range. I think it is better to do nothing than doing something different than what the user expects. Is there any chance that this will be accepted? If not, I don't want to waste time, yours and mine,  fixing this.

wolfmanjm:
Yes with these changes it will be accepted. I may need to clean it up afterwards though ;)

JesperKrogPoulsen:
Pushed. Feel free to clean up, but please comment what you find needs to be changes. I am a SQL/perl/python-developer by trade and I need to brush up my c++-knowledge from old times.

Nutz95:
Hello! i bring here my little contribution to this pull request. 
I've also been interested by this feature.
I'm using the M109 RXX and M190 RXX before using M81 and it's working as expected on my side.
at least , to me this feature can be merged on edge branch

I haven't tested the M109 S220 R230 side of the feature however.

Nutz95 (approved):
Hello! i bring here my little contribution to this pull request.
I've also been interested by this feature.
I'm using the M109 RXX and M190 RXX before using M81 and it's working as expected on my side.
at least , to me this feature can be merged on edge branch

I haven't tested the M109 S220 R230 side of the feature however.

wolfmanjm:
Superceded by #1279

PR #1058, opened 2016-11-08
Hotend and bed can be preheated separately

nmaggioni:
Some heated beds take significantly longer to heat up rather than most hotends, so I often find myself turning on the former a minute or two before the latter. I guess that having the possibility to preheat a single element could benefit many users.

I created a subscreen in **Prepare > Pre Heat...**, with the options to preheat everything (as to maintain backwards compatibility, the old behavior is just one more button press away), preheat the hotend only, and preheat the bed only.

_Do you think that this could be a valuable addition to the fw?_

+ The feature was tested on my _Smoothieboard 5X v1.0a _ and [_RepRapDiscount GLCD_](http://smoothieware.org/panel#toc6).


wolfmanjm:
Thank you for your submission. I feel this is not very useful, and in fact could easily be done as a custom menu entry. I also do not like that it complicates the menu structure for no real gain.

wolfmanjm:
It can also be done with the existing temperature screen. I think duplication of functionality is not really a good idea given how short of flash we are.

arthurwolf:
If this feature can be implemented by doing a custom menu entry, please add an example for those menu entries to the wiki.

Issue #918, opened 2016-04-25
Temperatureswitch suspends working when hitting print


dintid:
Posted here after suggestions by wolfmanjm

I have learned this is expected behaviour as the code is right now, as M109/190 is supossed to suspend/stop the execution of Gcodes.
Since the switch can execute Gcodes it is under this umbrella.

**My request** is for the switches to be either exempt from the above restrictions, or the addition of a new switchtype where gcodes are not possible.

**Example:** many other system critical fan usages I'd like to use
**Desired effect by using temperature switch:** Heatsink hotend fan (Calling it HHF from now on) turns on and off when hotend thermistor reaches 50c.
**Reason:** To avoid noisy HHF unless the hotend is actually hot and to avoid forgetting to turn it on if controlled manually.

**Symptoms - unwanted behaviour**
If temperature is below 50c on hotend when I hit print, the HHF does not start untill the print itself starts. Same for FAN.
If temperature is above 50c on hotend, when I hit print, the HHF shuts off and does not start untill the print itself starts. Same for FAN.
If print is stopped, the HHF fan stops as well!

Google-Group discussion over here:
https://groups.google.com/forum/#!topic/smoothieware-support/GQOyoaneDqw

wolfmanjm:
What I plan to fix is the temperature switch will turn on the fan (or off) when the set temperature is reached regardless of if there is an M109 active.
I do not plan on adding any new switch types or changing the way switches currently work.
This is solely focused on the one use case of temperatureswitch (which is already processed in on_idle) to be able to toggle the switch pin on/off as required in on_idle.
If the referenced switch  is set to issue a gcode on toggle, then it will still have to wait for M109 to finish.
The current issue is that switches are always toggled in on_main, which is blocked by M109/M190. This would not be an issue if there were no M109/M190 in the gcode file.
FWIW M109/M190 work exactly as they are supposed to work. So there will be no changes there.

qharley:
I like this. Currently a workaround is to manually turn on the fan before
heating...

dintid:
Awesome. The alternative ideas were just if they were easier to implement.
Solution would Work perfect in the usecase specificed.

@qharley one problem is that if you turn on the fan specified in a switch, the fan will be turned off once you hit print.
Usecase for me in this instance is to get the fans to Work autonomously without need to use gcodes at all.

wolfmanjm:
[...]

PR #1072, opened 2016-12-15
Fix to allow temperature switch to turn on a fan even if blocked with…

wolfmanjm:
… heat and wait

  turns fan on in on_idle, but there must be no output_on_command or output_off_command defined in the switch.

Fixes issue #918


tiaanv:
awesome. will test. Thx J

tiaanv:
Tested, and working 100% as designed.  Thx Jim

PR #1073, opened 2016-12-15
Repair morgan homing on new movement code

qharley:
This pull request fixes homing on polar bots like Morgan Scara by honouring the request to disable arm_solution. It is done by bypassing any request to arm_solution actuator_to_cartesian and cartesian_to_actuator while the disable_arm_solution flag is set.

When not set, it will not affect other systems.  The flag gets set when is_morgan is set. and disables shortly after movement.

In order for a Pull request to be accepted it must meet certain standards and guidelines.

In addition to following the [Coding-Standards](http://smoothieware.org/coding-standards) it is suggested the following be adopted.

* There must be a detailed description of the change, why it was added and what the use case is, and a link to the documentation to use it in the wiki if applicable.
* Only one feature per pull request, if you change multiple files with unrelated changes it will be rejected.
* Do not make gratuitous changes. If your coding style is different from the one adopted by most of the Smoothie codebase, do not make changes to suit your style. Peoples styles differ, and the new code you write may reflect your coding style, that is ok, but changing existing code to match your style is not ok. Doing so will delay the acceptance of your code, or have it rejected entirely.
* Maintain backwards compatibility. Do not make changes that will break other peoples systems, it is not acceptable to force everyone to change their config files to accommodate your change. If necessary add new configs to enable your change if it is breaking backward compatibility.
* Do not make changes that are specific to your setup. Smoothie needs to work in many different environments.
* Do not make changes that customize messages specific to your company or implementation. You are welcome to do this in your own fork and binary, it is not ok to try to merge that into the upstream repository.
* Test your changes is as many different configurations as you can, eg different arm solutions. In the Pull request state what tests have been done and what tests need to be done.
* Do not commit changes to the FirmwareBin directory or any binary files in the repository.
* Do not reformat the Config Samples to suit your style.
* Add examples of your config entries to the ConfigSamples/Snippets folder, only change existing config samples if absolutely necessary
* If you must correct existing comments (like correct spelling mistakes or typos) then submit them as separate pull requests
* Make sure your Pull request is against the current edge branch, and passes the Travis CI.
* All comments and code must be in English, this includes names of functions.

It may seem like a pain to follow these rules, but it takes the maintainers valuable time and effort to review and test pull requests, these guidelines help reduce that effort.

### DIFF
--- a/src/modules/robot/Robot.cpp
+++ b/src/modules/robot/Robot.cpp
@@ -237,7 +237,12 @@ void Robot::load_config()
     // initialise actuator positions to current cartesian position (X0 Y0 Z0)
     // so the first move can be correct if homing is not performed
     ActuatorCoordinates actuator_pos;
-    arm_solution->cartesian_to_actuator(last_milestone, actuator_pos);
+    if(!disable_arm_solution)
+        arm_solution->cartesian_to_actuator(this->last_machine_position, actuator_pos);
+    else
+        for (size_t i = X_AXIS; i <= Z_AXIS; i++)
+            actuator_pos[i] = last_machine_position[i];
+
     for (size_t i = 0; i < n_motors; i++)
         actuators[i]->change_last_milestone(actuator_pos[i]);
 
@@ -322,7 +327,11 @@ int Robot::print_position(uint8_t subcode, char *buf, size_t bufsize) const
 
         // get machine position from the actuator position using FK
         float mpos[3];
+        if(!disable_arm_solution)
         arm_solution->actuator_to_cartesian(current_position, mpos);
+            else
+        for (size_t i = X_AXIS; i <= Z_AXIS; i++)
+            mpos[i] = current_position[i];
 
         if(subcode == 1) { // M114.1 print realtime WCS
             // FIXME this currently includes the compensation transform which is incorrect so will be slightly off if it is in effect (but by very little)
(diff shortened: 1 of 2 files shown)


wolfmanjm:
As before I reject the changes to Robot, they are not required and way too hacky for me. I'd accept the changes to Endstops I think.

wolfmanjm:
What you need to do rather than hacking Robot and making it so messy is understand what is calling these from the endstops and not call those routines when is_scara is set. I explained this in the IRC channel last night.
Basically there is a lack of understanding of what endstops is doing when it resets axis, and that is really what needs to be fixed not scatter gunning hacks all over Robot.cpp. (Which BTW are totally incorrect and will cause severe issues down the road especially if you were to try to use kill/halt).

Basically You don't want to be calling the reset_position_from_current_actuator_position  for scara.

qharley:
As before I will try again.  I like Smoothieware, and I intend to keep using it.
I will restore robot to it former unhacked self, and limit the hacks to endstops.

wolfmanjm:
I am not entirely convinced that something like a scara can be homed without direct access to actuators and a special homing sequence. As there is a way to do direct actuator access (see SimpleShell test raw) I suspect that will be the most robust way to do it, and keep your own fork that can still be easily merged with upstream edge. the Homing would be in your own module probably the scara calibrate module even.

PR #1076, opened 2016-12-16
Repair SCARA homing on new ABC endstops.cpp

qharley:
Scara homing requires homing in the actuator space.  This pull request uses the axis space temporarily to perform homing.

As per guidelines by wolfmanjm


src/modules/tools/endstops/Endstops.cpp
```diff
@@ -106,6 +106,7 @@ enum STATES {
 Endstops::Endstops()
 {
     this->status = NOT_HOMING;
+    THEROBOT->disable_arm_solution = false;
```
wolfmanjm: please remove this it is not needed and can cause issues.

PR #1089, opened 2017-01-09
Removing web interface to save flash

arthurwolf:
... and force users to install better ones to SD.

This replaces the current web interface that is stored in flash with a page welcoming new network users, and pointing them to useful places.


arthurwolf:
If this is alright, I'll also remove the rest of the current webif files ( javascript etc ).

wolfmanjm:
It needs to be compiled into the flash just changing the index.html does not work.

arthurwolf:
Right, I missed that.
I'm going to do another PR with that done.
What do I do with the rest of the now un-used html/js stuff ? delete it, move it to an "old" folder ?

wolfmanjm:
just use the same PR, and yes delete the old files too.

arthurwolf:
Alright, updated the files, ran the scripts, recompiled the firmware ( and firmware-cnc ) bins, updated the md5sum.
If I forgot something please tell me I'll do my best.

wolfmanjm:
FYI you should NEVER update the firmwareBin folder. Please do not do that in future. I have a script that does that.

mvadu:
Why is this 3 year old PR still open? Can't we remove the conflicting changes and go ahead? in the last 3 years most if not all users are either using [Octoprint](http://smoothieware.org/octoprint) or [ESP3d](http://smoothieware.org/wifi) for UI. What is preventing from going ahead with this PR?

PR #1090, opened 2017-01-10
Allow Delta Grid ZProbe to save and load multiple grid files

pyjamasam:
In order to make life with my multiple magnetic effector plates easier I wanted the ability to save and load multiple grid compensation files.

By adding an additional identifier (Fn) to the M375 and M374(.1) gcodes you can now specify a grid file to load.

When effector plates are changed an appropriate matching grid file can be loaded.

If no Fn identifier is added to the command the original behaviour of loading and saving/deleting the original file is maintained.  (Also using F0 explicitly loads and saves the original file).

### DIFF
--- a/src/modules/tools/zprobe/DeltaGridStrategy.cpp
+++ b/src/modules/tools/zprobe/DeltaGridStrategy.cpp
@@ -95,11 +95,12 @@
 #define do_home_checksum             CHECKSUM("do_home")
 #define is_square_checksum           CHECKSUM("is_square")
 
-#define GRIDFILE "/sd/delta.grid"
+#define GRIDFILE_BASE "/sd/delta.grid"
 
 DeltaGridStrategy::DeltaGridStrategy(ZProbe *zprobe) : LevelingStrategy(zprobe)
 {
     grid= nullptr;
+    loaded_grid_file_number = 0;
 }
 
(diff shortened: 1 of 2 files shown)


wolfmanjm:
F is not a valid parameter as it is always used and reserved for feed rate. 

My feeling is this is not a widely needed feature, and as it significantly complicates the code base for a feature that maybe only one person needs it should not be accepted. However your fork could be added to the third party branches in the wiki in case someone wants that feature. 
Thank you

pyjamasam:
Good point on F, I'll switch that up.  Thanks for the guidance on that.

No worries on not accepting the pull request.  I thought I would contribute it just incase it was deemed valuable.

Thanks for taking a look at it either way.

wolfmanjm:
A alternative would be to use a command line command so you can pass in a file name...
So something like..

`load-grid /sd/mygrid1.dat`

may be cleaner to implement too.

arthurwolf:
Hey !

I added this to : http://smoothieware.org/third-party-branches
This way, users this is useful to can test/use it, and if it becomes popular enough, this will be a great argument for integration into Smoothie's mainstream. It's possible we don't have space in the v1 firmware, but if so it'll also be much easier to integrate into v2 later on.

Thanks for the code !

Cheers.

PR #1093, opened 2017-01-15
Minor cosmetic config changes

arthurwolf:
Added some config option informations, made a few minor cosmetic changes, added some links, removed a few options users shouldn't mess with anyway


arthurwolf:
I think this is very innocuous, if there are no objections I'll merge it soon, and then work on doing the same thing for the delta config file.

PR #1094, opened 2017-01-16
Fix the z-probe delta grid code for cartesian printers.

FluffyKaon:
Fixes the G29 command using the delta grid leveling strategy on cartesian printers (is_square = true).
Before the fix only the center of my printer was probed, after the fix all the points are probed as expected.


lordofhyphens:
I'm not sure what this change is supposed to do, and I see everything get probed :/

wolfmanjm:
good catch, thank you.
@lordofhyphens it fixes the test probe G29 which I guess no one tested yet ;)

lordofhyphens:
Ah. Herpderp. I never use G29 ^^;

PR #1097, opened 2017-01-20
initial optimization of G0-3 by calling robot directly and sending ok…

wolfmanjm:
… before planning


wolfmanjm:
Although this may help it is better to use fast-stream instead which bypasses the entire issue.

PR #1102, opened 2017-01-26
Added AD597 thermocouple board support

henriz:
Added support for the AD597 based thermocouple board as sold by, amongst others, reprapworld.com. See http://reprap.org/wiki/ExtThermoCouple_1.0. Basically the module is the same as the AD8495 module, only the calculation from sensor value to temperature differs.

This relies on the board getting 5v, and the SIG pin of the board connected to an analogue input with pulldown (ie. 1.30v or 1.31v).

Config values:

temperature_control.hotend.sensor                 ad597 
temperature_control.hotend.ad597_pin              1.31v 
temperature_control.hotend.ad597_offset           0

Note that the offset is merely a calibration value, it is not used as with the AD8495. Suitable pins are 1.30 and 1.31.

### DIFF
--- a/ConfigSamples/Snippets/AD597-thermocouple-board.config
+++ b/ConfigSamples/Snippets/AD597-thermocouple-board.config
@@ -1,0 +1,3 @@
+temperature_control.hotend.sensor                 ad597 
+temperature_control.hotend.ad597_pin              1.31v 
+temperature_control.hotend.ad597_offset           0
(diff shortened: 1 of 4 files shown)


wolfmanjm:
As there is apparently only one line change then a whole new file is unnecessary, why not just add an option that says which ADxxx it is and use the one line that is different.

henriz:
I totally agree, however, that would also involve changing existing AD8495 module. Basically I was told not to do that.

The correct way would still be an own module, but a base class which omits the calculation routine, and then derivates for every sensor which just implements the calculation.

I didn’t want to change existing code (also was basically told not to) so added it as separate module. It has the lowest impact.

henriz:
Also, don't forget where the objects are instantiated (in temperaturecontrol.cpp).  It would add code there, add code to the module and then also the existing config and module have to be changed, as you can't set sensor type=AD8495 and then differentiate for AD597 and so on.

In short, although I totally agree code files with only one line really different feels wrong, in this case, it _is_ the cleanest solution.

To get the puricly perfect routine, the themperaturecontroll.cpp should be rewritten so that it has no knowledge of what class it's actually instantiating.

PR #1117, opened 2017-02-10
rectangular z grid probe

wolfmanjm:
as per PR #1116

### DIFF
--- a/src/modules/tools/zprobe/CartGridStrategy.cpp
+++ b/src/modules/tools/zprobe/CartGridStrategy.cpp
@@ -1,0 +1,547 @@
+/*
+ This code is derived from (and mostly copied from) Johann Rocholls code at https://github.com/jcrocholl/Marlin/blob/deltabot/Marlin/Marlin_main.cpp
+ license is the same as his code.
+
+    Summary
+    -------
+    Probes grid_size points in X and Y (total probes grid_size * grid_size) and stores the relative offsets from the 0,0 Z height
+    When enabled every move will calculate the Z offset based on interpolating the height offset within the grids nearest 4 points.
+
+    Configuration
+    -------------
+    The strategy must be enabled in the config as well as zprobe.
+
+      leveling-strategy.delta-grid.enable         true
+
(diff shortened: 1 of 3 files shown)


pflugj:
I have tested this new "cartesian-grid" and it seems to do the job ;-)
...well done and thanks ;-)

However I must say that I kind of liked to set the origin to the center. 
I built a pretty big CoreXY machine (>50x50x40) and 
I let the machine do a grid-leveling before every print.
The built platform is some 8.4mm security glass and should be pretty straight.
 
With the setting of the orgin to the center I usually did only a 30x30cm grid with 3x3 points
With the new "cartesian-grid" I have to set the Origin to the bottom left, to get the grid to work 
and have to run the full size grid over at least 45x45cm and therefore I also selected more grid points (5x5).

Maybe it is still possible to do the center setup with the cartesian ?
and do the cartesian-grid from the center?  

.. or do I miss something as usuall?

Thanks again
Jürgen

qharley:
We had it in Zgrid. Easy addition

wolfmanjm:
You can still set center 0,0 if you prefer that. the issue is what the machine coordinates are set to, and those need to be 0,0 in corner left. To set 0,0 center just set up WCS (as explained in the wiki) the scan will still scan the entire grid though starting at 0,0. 
You setup WCS with either G92 or G10 L2 or L20.
The scan size can be changed using the X and Y parameters to the G31, but will be relative to the 0,0 in machine coordinates as scans are always performed in machine coordinates.
If you really want 0,0 in the center then use delta grid, it will still works, although the is_square has been deprecated.
It is not easy to go back to using 0,0 as the center in the new rectangular grid so i don't think we will be doing that, it was a specific request we eliminate that requirement.

wolfmanjm:
FWIW there is no reason to run a grid scan every print, that is why we allow it to be saved and reloaded.

Also if you have a flat bed then there is no reason to use a grid at all you should use the threepoint leveling as that is far more efficient and does not slow the entire system down during a print. whereas zgrid will slow down due to the 1mm segmentation required.

PR #1124, opened 2017-02-14
Configure grid size on the fly with I J params of G31/G32

vbolshakov:
While working with rectangular but not square board having x and y steps equal not good.


.gitignore
```diff
@@ -22,3 +22,4 @@ debug
 /win_install.log
 /linux_install.log
 /src/default_excludes.mk
+.vscode
```
wolfmanjm: You are not allowed to change or check in the .gitignore file. Also all files must terminate with a \n

wolfmanjm (changes requested):

src/modules/tools/zprobe/CartGridStrategy.cpp
```diff
@@ -229,32 +257,32 @@ bool CartGridStrategy::load_grid(StreamOutput *stream)
...
+    stream->printf("grid loaded, grid: (%f, %f), size: %d x %d\n", x_size, y_size, load_grid_x_size, load_grid_y_size);
     fclose(fp);
     return true;
 }
 
-bool CartGridStrategy::probe_grid(int n, float x_size, float y_size, StreamOutput *stream)
+bool CartGridStrategy::probe_grid(int n, int m, float x_size, float y_size, StreamOutput *stream)
 {
-    if(n < 5) {
+    if((n < 5)||(n < 5)) {
```
wolfmanjm: typo n should be m

src/modules/tools/zprobe/CartGridStrategy.cpp
```diff
@@ -229,32 +257,32 @@ bool CartGridStrategy::load_grid(StreamOutput *stream)
...
         stream->printf("Need at least a 5x5 grid to probe\n");
         return true;
     }
 
     float initial_z = findBed();
     if(isnan(initial_z)) return false;
 
-    float x_step= x_size / n;
-    float y_step= y_size / n;
+    float x_step= x_size / (n -1);
```
wolfmanjm: why? this is wrong IMO

wolfmanjm:
This is ok, but needs a few fixes. Plus the config needs to be backward compatible with existing config settings. Which it currently is not. as a rule you cannot break config compatibility with previous versions of config settings.

src/modules/tools/zprobe/CartGridStrategy.cpp
vbolshakov: This is like
#define AUTO_BED_LEVELING_GRID_X ((RIGHT_PROBE_BED_POSITION - LEFT_PROBE_BED_POSITION) / (grid_size - 1)) in original code
x_step is delta, so with n grid points delta = length / (n - 1)

src/modules/tools/zprobe/CartGridStrategy.cpp
```diff
@@ -83,7 +84,8 @@
 #include <cmath>
 #include <fastmath.h>
 
-#define grid_size_checksum           CHECKSUM("size")
```
wolfmanjm: This needs to be kept for backward comptibility

wolfmanjm (changes requested):

src/modules/tools/zprobe/CartGridStrategy.cpp
```diff
@@ -196,17 +212,29 @@ bool CartGridStrategy::load_grid(StreamOutput *stream)
         return false;
     }
 
-    uint8_t size;
+    uint8_t load_grid_x_size, load_grid_y_size;
     float x, y;
 
-    if(fread(&size, sizeof(uint8_t), 1, fp) != 1) {
+    if(fread(&load_grid_x_size, sizeof(uint8_t), 1, fp) != 1) {
```
wolfmanjm, line 251: This breaks compatibility with existing saved grid files. Not quite sure how you would fix this.

src/modules/tools/zprobe/CartGridStrategy.cpp
vbolshakov: Fixed - use this value only if x or y not defined

src/modules/tools/zprobe/CartGridStrategy.cpp
vbolshakov: Use different filename? If grid is square then save and load to/from original name. If not square than  use new filename.

vbolshakov: Done.

src/modules/tools/zprobe/CartGridStrategy.cpp
vbolshakov: Done.

.gitignore
wolfmanjm: You need to remove this from the pull request.

src/modules/tools/zprobe/CartGridStrategy.cpp
```diff
@@ -106,7 +110,16 @@ CartGridStrategy::~CartGridStrategy()
 
 bool CartGridStrategy::handleConfig()
 {
-    grid_size = THEKERNEL->config->value(leveling_strategy_checksum, cart_grid_leveling_strategy_checksum, grid_size_checksum)->by_default(7)->as_number();
+
+    uint8_t grid_size = THEKERNEL->config->value(leveling_strategy_checksum, cart_grid_leveling_strategy_checksum, grid_size_checksum)->by_default(7)->as_number();
+    current_grid_x_size = configured_grid_x_size = THEKERNEL->config->value(leveling_strategy_checksum, cart_grid_leveling_strategy_checksum, grid_x_size_checksum)->by_default(0)->as_number();
+    if(0 == configured_grid_x_size){
```
wolfmanjm: This is inefficient. Just put grid_size in the by_default

src/modules/tools/zprobe/CartGridStrategy.cpp
wolfmanjm: No it is specifically different for a reason so leave it the way it was please.

src/modules/tools/zprobe/CartGridStrategy.cpp
```diff
@@ -277,12 +319,13 @@ bool CartGridStrategy::handleGcode(Gcode *gcode)
             // first wait for an empty queue i.e. no moves left
             THEKERNEL->conveyor->wait_for_idle();
 
-            int n = gcode->has_letter('I') ? gcode->get_value('I') : 0;
+            int n = gcode->has_letter('I') ? gcode->get_value('I') : 7;
```
wolfmanjm: This is wrong too. it should default to the config specified grid size.

wolfmanjm: this should probably default to the grid size in the config not 7.

src/modules/tools/zprobe/CartGridStrategy.cpp
```diff
@@ -402,14 +445,26 @@ bool CartGridStrategy::doProbe(Gcode *gc)
     if(gc->has_letter('X')) x_size = gc->get_value('X'); // override default probe width, will get saved
     if(gc->has_letter('Y')) y_size = gc->get_value('Y'); // override default probe length, will get saved
 
+    if(gc->has_letter('I')) current_grid_x_size = gc->get_value('I'); // override default grid x size
+    if(current_grid_x_size > configured_grid_x_size){
+        gc->stream->printf("Grid X size bigger than configerd. Unable to allocate memory\n");
```
wolfmanjm: typo, and wrong error message, it is not allocating memory here.

src/modules/tools/zprobe/CartGridStrategy.cpp
```diff
@@ -402,14 +445,26 @@ bool CartGridStrategy::doProbe(Gcode *gc)
...
+        return false;
+    }
+
+    if(gc->has_letter('J')) current_grid_y_size = gc->get_value('J'); // override default grid y size
+    if(current_grid_y_size > configured_grid_y_size){
+        gc->stream->printf("Grid Y size bigger than configerd. Unable to allocate memory\n");
```
wolfmanjm: typo

src/modules/tools/zprobe/CartGridStrategy.h
```diff
@@ -28,14 +28,15 @@ class CartGridStrategy : public LevelingStrategy
...
+    bool probe_grid(int n, int m, float x_size, float y_size, StreamOutput *stream);
 
     float initial_height;
     float tolerance;
 
     float *grid;
     std::tuple<float, float, float> probe_offsets;
-    uint8_t grid_size;
+    uint8_t configured_grid_x_size, configured_grid_y_size;
+    uint8_t current_grid_x_size, current_grid_y_size;
```
wolfmanjm: this is using 4 words when it used 1. I do not like wasting memory. This needs to be compacted into a struct.
```
struct {
 uint8_t configured_grid_x_size:8;
 uint8_t configured_grid_y_size:8;
... etc
};
```

wolfmanjm (changes requested):

src/modules/tools/zprobe/CartGridStrategy.cpp
vbolshakov: Get this from original version:
            if(n == 0) n = 7;
            probe_grid(n, x, y, gcode->stream);

Will fix with configured values

src/modules/tools/zprobe/CartGridStrategy.cpp
vbolshakov: Done.

src/modules/tools/zprobe/CartGridStrategy.cpp
vbolshakov: Done.

src/modules/tools/zprobe/CartGridStrategy.h
vbolshakov: Done.

src/modules/tools/zprobe/CartGridStrategy.cpp
vbolshakov: Done

.gitignore
vbolshakov: Don;t know how... without full rewrite histore of commits and lost this conversation.

src/modules/tools/zprobe/CartGridStrategy.cpp
```diff
@@ -402,14 +446,26 @@ bool CartGridStrategy::doProbe(Gcode *gc)
     if(gc->has_letter('X')) x_size = gc->get_value('X'); // override default probe width, will get saved
     if(gc->has_letter('Y')) y_size = gc->get_value('Y'); // override default probe length, will get saved
 
+    if(gc->has_letter('I')) current_grid_x_size = gc->get_value('I'); // override default grid x size
+    if(current_grid_x_size > configured_grid_x_size){
+        gc->stream->printf("Grid X size bigger than configured. Chande configuration.\n");
```
wolfmanjm: typo

src/modules/tools/zprobe/CartGridStrategy.cpp
```diff
@@ -402,14 +446,26 @@ bool CartGridStrategy::doProbe(Gcode *gc)
...
+        return false;
+    }
+
+    if(gc->has_letter('J')) current_grid_y_size = gc->get_value('J'); // override default grid y size
+    if(current_grid_y_size > configured_grid_y_size){
+        gc->stream->printf("Grid Y size bigger than configured. Chande configuration.\n");
```
wolfmanjm: typo

wolfmanjm:
not sure why you keep saying G28 when G28 is not involved at all. it is G31/G32

wolfmanjm (approved):

src/modules/tools/zprobe/CartGridStrategy.cpp
vbolshakov: Done.

src/modules/tools/zprobe/CartGridStrategy.cpp
vbolshakov: Done.

vbolshakov:
Yes! You right G29, not G28.

vbolshakov:
Yes. Before it go to bottom right corner while 0,0 is in the middle of square. Now it just start from 0,0

vbolshakov:
Revert coordinates format in DEBUG messages. Previously pushed to wrong branch.

vbolshakov:
It is full set of my additions to CartGridStrategy:
1. non square grid with different steps on x and y
2. ability to change grid on the fly
3. ability to probe not only from 0,0 byt from any point
4. human redable formatting of table of levels, so human can understand it more easyli

it is packed in one request as you told me before

wolfmanjm:
When we document this in the Wiki we need to point out that if not using 0,0 as the limit, but using a sub rectangle, then moving outside that will be disastrous as there is no extrapolation for points outside the defined rectangle. (We could add extrapolation as in the delta grid, but I suspect it is not needed for the use case).

vbolshakov:
Yes! you absolutely right. I explicitly disable extrapolation because got several Z axis movement MAX while navigating outside of measured region.
Also, may be it will be better to explicitly convert from negative width and length (that correctly processed now) to start point offset and positive width nd length.
For example - start point -10,-10, width and length also -10,-10. It is geometrically correct, but not easy to understanding, so we can convert it to -20,-20 as start point and 10,10 as width and length. Only in humanreadable mode.
When code will be pulled i will prepare wiki with images for all cases.

src/modules/tools/zprobe/CartGridStrategy.cpp
```diff
@@ -15,7 +15,8 @@
 
     The size of the grid can be set with...
 
-      leveling-strategy.rectangular-grid.size        7
+      leveling-strategy.rectangular-grid.grid_x_size        7
+      leveling-strategy.rectangular-grid.grid_y_size        7
```
lordofhyphens: I would like to support the old size config syntax here in this documentation.

src/modules/tools/zprobe/CartGridStrategy.cpp
```diff
@@ -229,41 +287,41 @@ bool CartGridStrategy::load_grid(StreamOutput *stream)
...
+        float y = y_start + y_step * c;
         for (int r = 0; r < n; ++r) {
-            float x = x_step * r;
+            float x = x_start + x_step * r;
             float z = 0.0F;
             float mm;
             if(!zprobe->doProbeAt(mm, x, y)) return false;
             z = zprobe->getProbeHeight() - mm;
-            stream->printf("%8.4f ", z);
+            stream->printf("%1.4f ", z);
```
lordofhyphens, line 331: why was the total width reduced to 1 here?

src/modules/tools/zprobe/CartGridStrategy.cpp
```diff
@@ -277,12 +335,28 @@ bool CartGridStrategy::handleGcode(Gcode *gcode)
...
             THEKERNEL->conveyor->wait_for_idle();
 
-            int n = gcode->has_letter('I') ? gcode->get_value('I') : 0;
-            float x = x_size, y = y_size;
-            if(gcode->has_letter('X')) x = gcode->get_value('X'); // override default probe width
-            if(gcode->has_letter('Y')) y = gcode->get_value('Y'); // override default probe length
-            if(n == 0) n = 7;
-            probe_grid(n, x, y, gcode->stream);
+            int n = gcode->has_letter('I') ? gcode->get_value('I') : configured_grid_x_size;
+            int m = gcode->has_letter('J') ? gcode->get_value('J') : configured_grid_y_size;
```
lordofhyphens, line 346: So J is now compulsory for this gcode? This doesn't follow very easily from the way that grid size is overridden. Example:
Probing a 5x5 grid, saved config grid is 7x7. 
We're now required to specify I5 J5, instead of just I5 like before. Specifying I5 only would get a grid of 5x7, which is probably not is expected. 

There should be an error check here at least if grid_size is defined or completely to avoid this pitfall if both I and J are compulsory for probing to work.

src/modules/tools/zprobe/CartGridStrategy.cpp
```diff
@@ -396,11 +470,32 @@ float CartGridStrategy::findBed()
 bool CartGridStrategy::doProbe(Gcode *gc)
 {
     gc->stream->printf("Rectangular Grid Probe...\n");
+
+    if(only_by_two_corners){
+        if(gc->has_letter('X') && gc->has_letter('X') && gc->has_letter('A') && gc->has_letter('B')){
+            x_start = gc->get_value('X'); // override default probe start point, will get saved
+            y_start = gc->get_value('Y'); // override default probe start point, will get saved
+            x_size = gc->get_value('A'); // override default probe width, will get saved
+            y_size = gc->get_value('B'); // override default probe length, will get saved
```
lordofhyphens: From what I understood of the documentation, isn't A,B defined as the size of the rectangle? This is being used as the coordinates of the other two points of the rectangle (which will probably cause some confusion). Either fix the documentation or change the implementation to reflect the x_size and y_size being relative coordinates from the start points.

src/modules/tools/zprobe/CartGridStrategy.cpp
```diff
@@ -396,11 +470,32 @@ float CartGridStrategy::findBed()
...
+        if(gc->has_letter('Y')) y_size = gc->get_value('Y'); // override default probe length, will get saved
+    }
+    
     setAdjustFunction(false);
     reset_bed_level();
 
-    if(gc->has_letter('X')) x_size = gc->get_value('X'); // override default probe width, will get saved
-    if(gc->has_letter('Y')) y_size = gc->get_value('Y'); // override default probe length, will get saved
+    if(gc->has_letter('I')) current_grid_x_size = gc->get_value('I'); // override default grid x size
+    if(gc->has_letter('J')) current_grid_y_size = gc->get_value('J'); // override default grid y size
```
lordofhyphens, line 489: Adding comment here (same as previous override) so that this doesn't get missed.

src/modules/tools/zprobe/CartGridStrategy.cpp
```diff
@@ -451,24 +546,27 @@ bool CartGridStrategy::doProbe(Gcode *gc)
...
-    float left = (1 - ratio_y) * z1 + ratio_y * z2;
-    float right = (1 - ratio_y) * z3 + ratio_y * z4;
-    float offset = (1 - ratio_x) * left + ratio_x * right;
-
-    if(inverse)
-        target[Z_AXIS] -= offset;
-    else
-        target[Z_AXIS] += offset;
+    if ((std::min(x_start, x_start + x_size) <= target[X_AXIS]) && (target[X_AXIS] <= std::max(x_start, x_start + x_size)) && 
+        (std::min(y_start, y_start + y_size) <= target[Y_AXIS]) && (target[Y_AXIS] <= std::max(y_start, y_start + y_size))) {
```
lordofhyphens: See above for the note on A,B values (x_size and y_size). According to this, there will be a distance between y_start and y_size that isn't probed and yet is calculated for compensation?

Is x_size and y_size (from AB config) absolute or relative parameters to x_start?

lordofhyphens (changes requested):
Some confusion regarding x_size and y_size and some comments at the top for grid_size.

src/modules/tools/zprobe/CartGridStrategy.cpp
wolfmanjm: AFAIK this is backward compatible so the old config should continue to work. That was the intention to have it be totally backward compatible with no config changes needed.

src/modules/tools/zprobe/CartGridStrategy.cpp
wolfmanjm: I wanted it left as 1 which was the old way it was. this means that when it is printed there are no preceding spaces, but does not limit the pre period length.

src/modules/tools/zprobe/CartGridStrategy.cpp
wolfmanjm: Agreed both need to be optional. As I read it they are.

src/modules/tools/zprobe/CartGridStrategy.cpp
vbolshakov: A and B relative to start point. I tried both variants and this was easier for my test group (2 coworkers that use my CNC for PCB milling).

src/modules/tools/zprobe/CartGridStrategy.cpp
vbolshakov: “Two corners” not right name for this mode, because it use only one corner and rectangle size. I take it from CAD i'm using. Can you suggest new name? True two corner mode not easy to use...

src/modules/tools/zprobe/CartGridStrategy.cpp
vbolshakov: Also, may be it will be better to explicitly convert from negative width and length (that correctly processed now) to start point offset and positive width and length.
For example - start point -10,-10, width and length also -10,-10. It is geometrically correct, but not easy to understanding, so we can convert it to -20,-20 as start point and 10,10 as width and length. Only in humanreadable mode.

src/modules/tools/zprobe/CartGridStrategy.cpp
lordofhyphens: What I meant I'd that the note here omits the old config, which is incorrect.

src/modules/tools/zprobe/CartGridStrategy.cpp
```diff
@@ -349,14 +424,14 @@ bool CartGridStrategy::handleGcode(Gcode *gcode)
 
 // These are convenience defines to keep the code as close to the original as possible it also saves memory and flash
 // set the rectangle in which to probe
-#define LEFT_PROBE_BED_POSITION (0)
+#define LEFT_PROBE_BED_POSITION (x_start)
 #define RIGHT_PROBE_BED_POSITION (x_size)
 #define BACK_PROBE_BED_POSITION (y_size)
```
lordofhyphens: Ensure that the x_size and y_size used here are correct. They are absolute points.

lordofhyphens (changes requested):

src/modules/tools/zprobe/CartGridStrategy.cpp
vbolshakov: They used previously as size:
#define LEFT_PROBE_BED_POSITION (0)
#define RIGHT_PROBE_BED_POSITION (x_size)
#define BACK_PROBE_BED_POSITION (y_size)
#define FRONT_PROBE_BED_POSITION (0)

vbolshakov: May be someone (wolfmanjm?) must write small naming guide - what names best for what things - width and heigth or with and length and so on...

lordofhyphens: They're usable as size because absolute is the same as relative when the start point is 0.

You need to ensure that your rectangle is using the values you think it is (effectively a find-replace). Since you shadowed the names I can't tell if the x_size used there is a local variable or the member variable.

vbolshakov: There two changes:
First i remove legacy from DeltaGrid - where left/rigth and front/back calculated from radius or width/length of table.
Then i add size changeable on the fly.
I think that this define can be removed. And may be chant to "this->" when using default values.

vbolshakov: Actually, there is only one place, where local and class variable can be misused - G28 processing and probe_grid.
But i change all local variables to different name and all significant places with class variable to "this->"

wolfmanjm:
Please stop making changes, they are simply making the code worse and worse. I am close to the point of rejecting this now. Only fix things that a re broken, and check with me first.

src/modules/tools/zprobe/CartGridStrategy.cpp
```diff
@@ -335,8 +336,8 @@ bool CartGridStrategy::handleGcode(Gcode *gcode)
             // first wait for an empty queue i.e. no moves left
             THEKERNEL->conveyor->wait_for_idle();
 
-            int n = gcode->has_letter('I') ? gcode->get_value('I') : configured_grid_x_size;
-            int m = gcode->has_letter('J') ? gcode->get_value('J') : configured_grid_y_size;
+            int n = gcode->has_letter('I') ? gcode->get_value('I') : (gcode->has_letter('J') ? gcode->get_value('J') : configured_grid_x_size);
```
wolfmanjm: This is some of the worst code I have seen in a while! it is totally obfuscated and ugly.

src/modules/tools/zprobe/CartGridStrategy.cpp
```diff
@@ -489,8 +490,8 @@ bool CartGridStrategy::doProbe(Gcode *gc)
     setAdjustFunction(false);
     reset_bed_level();
 
-    if(gc->has_letter('I')) current_grid_x_size = gc->get_value('I'); // override default grid x size
-    if(gc->has_letter('J')) current_grid_y_size = gc->get_value('J'); // override default grid y size
+    current_grid_x_size = gc->has_letter('I') ? gc->get_value('I') : (gc->has_letter('J') ? gc->get_value('J') : current_grid_x_size); // If I provided update x grid size, if not try J.
+    current_grid_y_size = gc->has_letter('J') ? gc->get_value('J') : (gc->has_letter('I') ? gc->get_value('I') : current_grid_y_size); // If J provided update y grid size, if not try I.
```
wolfmanjm: I do not really agree with this change. I and J must be optional but if one is left out the other should revert to the default in config not to the one specified.
@lordofhyphens is this really what you wanted?

wolfmanjm (changes requested):

src/modules/tools/zprobe/CartGridStrategy.cpp
vbolshakov: Reverted.

src/modules/tools/zprobe/CartGridStrategy.cpp
vbolshakov: Reverted.

src/modules/tools/zprobe/CartGridStrategy.cpp
vbolshakov: There is a note about "leveling-strategy.rectangular-grid.size"

 The size of the grid can be set with...
      leveling-strategy.rectangular-grid.size        7
    or
      leveling-strategy.rectangular-grid.grid_x_size        7
      leveling-strategy.rectangular-grid.grid_y_size        7
    this is the X and Y size of the grid, it must be an odd number, the default is 7 which is 49 probe points

wolfmanjm (approved):

wolfmanjm:
Can you update the wiki please.

vbolshakov:
Yes!

triffid:
This looks like a great example of how to get things merged into Smoothie :1st_place_medal: :+1:

PR #1125, opened 2017-02-15
Stating the obvious : creating a code of conduct.

arthurwolf:
This is something that can be pointed at when people cross the line as happens sometimes. It might be useful to *some* people for who this isn't obvious, though I doubt it.


arthurwolf:
I think this is pretty un-controversial, merging.

PR #1144, opened 2017-03-03
Add ZGrid compatibility to CartesianGrid

qharley:
Adds functions to cartgrid in order to make it usable for machines with long beds, like Morgan.
* Ability to specify different X and Y grid resolution
* Ability for head to park for manual attachment of probe
* One line change in ZProbe: return move only in Z direction to prevent stalling in undefined positions created by irregular bed shapes (halfmoon bed on Morgan)

All the changes are backwards compatible, and will have no adverse affects on current CartGrid users.

### DIFF
--- a/src/modules/tools/zprobe/ZProbe.cpp
+++ b/src/modules/tools/zprobe/ZProbe.cpp
@@ -238,8 +238,8 @@ bool ZProbe::run_probe_return(float& mm, float feedrate, float max_dist, bool re
         if(fr > this->fast_feedrate) fr = this->fast_feedrate; // unless that is greater than fast feedrate
     }
 
-    // absolute move back to saved starting position
-    coordinated_move(save_pos[0], save_pos[1], save_pos[2], fr, false);
+    // absolute move back to saved starting position: Only in Z direction
+    coordinated_move(NAN, NAN, save_pos[2], fr, false);
 
     return ok;
 }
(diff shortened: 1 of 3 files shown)


src/modules/tools/zprobe/ZProbe.cpp
```diff
@@ -238,8 +238,8 @@ bool ZProbe::run_probe_return(float& mm, float feedrate, float max_dist, bool re
         if(fr > this->fast_feedrate) fr = this->fast_feedrate; // unless that is greater than fast feedrate
     }
 
-    // absolute move back to saved starting position
-    coordinated_move(save_pos[0], save_pos[1], save_pos[2], fr, false);
+    // absolute move back to saved starting position: Only in Z direction
+    coordinated_move(NAN, NAN, save_pos[2], fr, false);
```
wolfmanjm, line 242: Why Change this?

wolfmanjm:
This is a duplicate submission to PR #1124, So as that was first I was goin gto accept that one, he has made subsequent changes too. We cannot have two competing versions.

Sorry.

qharley:
Fine. When will that one be added?

wolfmanjm:
once @lordofhyphens has reviewed and tested it. This is not a comment on your submission but there "can only be one" ;) and he has been working on that PR for some time, and it covers use cases like PCB milling.

qharley:
Seems that I will just have to be content with having to hack at the
software every time anyone changes something. Not as much fun as it sounds
like

src/modules/tools/zprobe/ZProbe.cpp
qharley: If the probe position is in a limited area on Morgan, the probe returns very slowly.  Since the movement is only in Z, there is no reason not to only return the Z to the saved position.

PR #1173, opened 2017-04-19
Allow multiple extruders to use the same temperature control.

hmflash:
This change allows multiple extruder modules to use the same temperature control module.

Currently, there is an implicit association based on module name, so that `extruder.xxx` is automatically paired with `temperature_control.xxx`.  Currently, for multi extruder single hotend configurations, the temperature of the hotend can only be changed when T0 is selected.  While I can work around this by post processing my gcode, it seems like the the firmware should be able to change the hotend temp of the active extruder when multiple extruders share the same hot end.

This change adds a new configuration option to the extruder module allowing the user to specify the temperature_control module to use.  If the configuration is omitted, the default value of the extruder module name is used making it backwards compatible.  A full example would be:

```
# hotend1 uses the same temperature_control as hotend
extruder.hotend2.temperature_control        hotend
```

I have tested the change and verified the following behavior with two extruders and a single temperature_control

```
; extruder.hotend2.temperature_control config omitted
M104 S150 T0; hotend target changes to 150
M104 S155 T1; hotend target stays at 150
M104 S160 T0; hotend target changes to 160
```

```
; extruder.hotend2.temperature_control hotend specified
M104 S150 T0; hotend target changes to 150
M104 S155 T1; hotend target changes to 155
M104 S160 T0; hotend target changes to 160
```

I wanted to get your feedback on the change.  It was easiest to change the tool manager to contain Extruder* instead of Tool* which does couple the two objects more closely.  However, the only place the tool manager is constructed is in the ExtruderMaker, so the coupling seemed reasonable.

Still need to do do docs but wanted some initial feedback.

### DIFF
--- a/src/modules/tools/extruder/Extruder.cpp
+++ b/src/modules/tools/extruder/Extruder.cpp
@@ -24,6 +24,7 @@
 #include "PublicDataRequest.h"
 #include "StreamOutputPool.h"
 #include "ExtruderPublicAccess.h"
+#include "utils.h"
 
 #include <mri.h>
 
@@ -51,6 +52,7 @@
 #define x_offset_checksum                    CHECKSUM("x_offset")
 #define y_offset_checksum                    CHECKSUM("y_offset")
 #define z_offset_checksum                    CHECKSUM("z_offset")
+#define temperature_control_checksum         CHECKSUM("temperature_control")
 
 #define retract_length_checksum              CHECKSUM("retract_length")
 #define retract_feedrate_checksum            CHECKSUM("retract_feedrate")
@@ -124,6 +126,13 @@ void Extruder::config_load()
     if(filament_diameter > 0.01F) {
         this->volumetric_multiplier = 1.0F / (powf(this->filament_diameter / 2, 2) * PI);
     }
+
+    // Get associated temperature_control name, defaults to extruder name
+    string s = THEKERNEL->config->value(extruder_checksum, this->identifier, temperature_control_checksum)->by_default("" )->as_string();
+    if (s.empty())
+        this->temperature_control_name = this->identifier;
+    else
+        this->temperature_control_name = get_checksum(s);
 
     // Stepper motor object for the extruder
     stepper_motor = new StepperMotor(step_pin, dir_pin, en_pin);
--- a/src/modules/tools/extruder/Extruder.h
+++ b/src/modules/tools/extruder/Extruder.h
@@ -28,6 +28,7 @@ class Extruder : public Tool {
         void select();
         void deselect();
         float get_e_scale(void) const { return volumetric_multiplier * extruder_multiplier; }
+        uint16_t get_temperature_control_name(void) const { return temperature_control_name; }
 
     private:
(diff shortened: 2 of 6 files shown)


wolfmanjm:
I have mixed feelings.

One temperature control and two extruders currently works perfectly well as is, so I do not see any need to add this.
I also feel very strongly that Tx is a control word as defined by the NIST spec. So using it as a parameter is wrong.
We have a hack that allows the control word anywhere in the line but that will also switch the tool, but that is IMO what should happen.

It is sad you have to post process your gcode, but may I suggest you ask the people who generate the gcode to fix it so they correctly use T0/T1 as a control word which is modal?

So my vote is to reject this change, as it may also have side effects for CNC use etc where Tn would need to be handled correctly per the NIST spec.

qharley:
Looked like such a promising feature.  ( I have my bowden toting Morgan 3D Printers hat on at the moment )

hmflash:
This change doesn't have anything to do with processing T0/T1 as a parameter and none of the Tx handling code was touched.  While two extruders one temperature control works, it is currently not possible to change the temperature unless T0 is the currently selected tool.  This change just alters how a temperature control module determines if it is currently selected in response to M104 SXXX.

On dual extruder, single temp control, given the following gcode:

```
T1
M104 S200
```

The `M104 S200` does not change the hotend temperature because T1 is the currently selected tool and the temp control module is bound to T0.  I would argue the temperature should change the hotend temp since T1 is really the active hotend the stepper is trying to push filament through.

FWIW, my post processing script is basically, find all M104 lines, and insert a T0 before and restore the tool after.

wolfmanjm:
Oh that is much easier to fix without all this code :)
We just make sure that a single extruder is always selected, that is pretty easy to do, in fact I thought it already did that.
So your "fix" means that T0 and T1 is still modal?
so if you used T1 on an M104 command it would switch to the second extruder?

hmflash:
Yes, T0 and T1 are still modal. `M104 S200 T1` would activate tool T1.

wolfmanjm:
Ok I fixed this issue in a much simpler way. See PR #1174, please test it let me know if it works for you.

I believe this was always meant to be the way ot worked, it may have been broken along the way.

wolfmanjm:
Basically now you no longer need to have T0 or T1 on a set temp command if you only have one heater.
May want the gcode generators allow that option to simplify things.
That would also solve the issue of T1 being modal when used as a parameter which I am not willing to change.
Really it should be used as a command and never used as a parameter.!

wolfmanjm:
Actually reviewing the code it should have worked as it was, as it checks if the named temperature control is managed or not. the problem was you did not setup the configuration correctly. You should not have named the temperature control the same as either of your extruders, then it would have worked.
This is so because the temperature control you have is not linked to either extruder (or linked to both).
I suspect the wiki should elaborate on that.
Anyway my PR makes it so the requirement to name the temperature control differently is no longer required in this particular circumstance. However there were no code changes really required, just a config change and maybe some more docs in the wiki.

PR #1174, opened 2017-04-19
Fix issue where a single temperature control with multiple extruders …

wolfmanjm:
…needs to have tool 0 selected.

Now temp setting will always work regardless of which extruder is set.


hmflash:
I can't test this right now, but I think it won't work because a heated bed temperature control module will put the count at two and the single bool will thus be false.  Realizing that if you name the temp control module different from the extruder tools will fix this, I think just updating the documentation is the correct solution.

wolfmanjm:
This is not needed and won't work either, good catch ;) Tha is what happens when you bang out code at 3am :)

PR #1179, opened 2017-04-28
plan9: Update error string names

jkoelker:
Several of the error names are int values defined elsewhere. Prefix them
with `P9_` to prevent `plan9.cpp:27:12: error: expected unqualified-id
 before numeric constant`.


wolfmanjm:
Thank you for doing this, even though plan9 is no longer officially supported in the system, thi sis good for anyone wanting to enable it.

PR #1184, opened 2017-05-10
Fix missing zprobe_offset

nbrlt:
I added missing zprobe_offset values in the find_bed function to avoid the probe looking for the bed outside of the print surface. Now puts correctly the probe at 0,0.


wolfmanjm:
Thank you. This was a good find.

Issue #1203, opened 2017-07-13
Pause before probing?


DjDemonD:
Hi all,

I now market and sell an open-source piezo Z-probe. More details can be found online at reprap forum, just see the general forum main page, I won't leave any spam here.

What we are interested in is having a configurable pause before a probing dive takes place. RRF already has this as DC42 and T3P3 are making a nozzle contact probe based on piezo like we do, and the M558 command now has an R (recovery) parameter which is pause (in seconds) after positioning over a probing point before diving to probe. A value of 0.2sec is usually enough. It allows mechanical noise to settle before probing an reduces/eliminates false triggers from this very sensitive probe.

These things are becoming quite popular now as they are by far the most convenient and accurate z-probe available.

Would you consider implementing it for the benefit of the now hundreds of people using piezo z probes? I am not certain if your firmware already has such a switch that can be used or if the code already builds in a recovery before probing, but would be very grateful if you would consider it.

Kind regards
DjDemonD/Simon
Precisionpiezo.co.uk

wolfmanjm:
I have no plans to add it, but you are welcome to add it and submit a pull request.

DjDemonD:
Many thanks I'll put that put to some of the users. I am willing to send a free module as a thank you to anyone who can make a fork and add this feature.

wolfmanjm:
#1204 fixes this

PR #1204, opened 2017-07-16
Dwell before probing

sdavi:
Adds a configurable dwell time before performing a probe. Implements issue #1203. 

Added configuration option:
`zprobe.dwell_before_probing            0.2    #Dwell time in seconds before probing.`

An additional parameter added to M670:
`M670 D0.2`

Default value is 0.  

Tested on a delta by running a single probe (G30), delta grid leveling strategy (G31) and delta calibration (G32).

### DIFF
--- a/src/modules/tools/zprobe/ZProbe.cpp
+++ b/src/modules/tools/zprobe/ZProbe.cpp
@@ -43,6 +43,7 @@
 #define gamma_max_checksum       CHECKSUM("gamma_max")
 #define max_z_checksum           CHECKSUM("max_z")
 #define reverse_z_direction_checksum CHECKSUM("reverse_z")
+#define dwell_before_probing_checksum CHECKSUM("dwell_before_probing")
 
 // from endstop section
 #define delta_homing_checksum    CHECKSUM("delta_homing")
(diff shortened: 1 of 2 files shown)


src/modules/tools/zprobe/ZProbe.cpp
```diff
@@ -191,6 +193,8 @@ bool ZProbe::run_probe(float& mm, float feedrate, float max_dist, bool reverse)
     // save current actuator position so we can report how far we moved
     float z_start_pos= THEROBOT->actuators[Z_AXIS]->get_current_position();
 
+    if(dwell_before_probing > 0) safe_delay_ms(dwell_before_probing*1000);
```
wolfmanjm: can you change this to `> .0001F` to avoid the float issues?

sdavi: Fixed.

adrianjobst:
Thanks for the implementation. I recently bought a precision piezo so this and #1203 got my attention. I tested it but this will not really work. I think it should be better if the delay is before the first if condition. Otherwise vibrations from any previous movement will lead to cancellation before the delay can be executed

markkahler:
I have bought a precision piezo that is mounted to my end effector/E3D HotEnd. The printer I have made is a core-xy. I have added the following to the config.txt zprobe area: **zprobe.dwell_before_probing 1**
When doing a grid probe, during the initial Z movement of the print bed, the probe/piezo is not being triggered and the E3D Nozzle is correctly detecting the bed at the x0,y0 position. But when the end effector moves to the next probe position, there is a flashing/false detection from the piezo. As a result, before the next probe starts, I get the following message: **Probe failed to complete, check the initial probe height and/or initial_height settings**
Please can you change the G32 function to only start monitoring the probe after the 'dwell_before_probing' timeout before the z movement.

sdavi:
@adrianjobst has already submitted a pull request #1239 to address this

PR #1206, opened 2017-07-27
Module for reading a pin using G-code command

bullestock:
This adds a new `readpin` module which I have found useful for debugging. Sample usage:
```
readpin.d1.enable                            true
readpin.d1.pin                               2.11
readpin.d1.gcode                             M700
```
Now, issuing `M700` prints the current state of pin 2.11 to the console.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -29,6 +29,7 @@
 #include "modules/utils/killbutton/KillButton.h"
 #include "modules/utils/PlayLed/PlayLed.h"
 #include "modules/utils/panel/Panel.h"
+#include "modules/utils/readpin/ReadPinPool.h"
 #include "libs/Network/uip/Network.h"
 #include "Config.h"
 #include "checksumm.h"
@@ -190,6 +191,11 @@ void init() {
     #ifndef NO_UTILS_MOTORDRIVERCONTROL
     kernel->add_module( new MotorDriverControl(0) );
     #endif
+
+    auto rpp = new ReadPinPool();
+    rpp->load_tools();
+    delete rpp;
+
     // Create and initialize USB stuff
     u.init();
 
--- a/src/modules/utils/readpin/ReadPin.cpp
+++ b/src/modules/utils/readpin/ReadPin.cpp
@@ -1,0 +1,59 @@
+#include "libs/Kernel.h"
+#include "ReadPin.h"
+#include "libs/nuts_bolts.h"
+#include "libs/utils.h"
+#include "Config.h"
+#include "SlowTicker.h"
+#include "libs/SerialMessage.h"
+#include "libs/StreamOutput.h"
+#include "checksumm.h"
+#include "ConfigValue.h"
+#include "StreamOutputPool.h"
+
+using namespace std;
+
+#define gcode_checksum               CHECKSUM("gcode")
+#define pin_checksum                 CHECKSUM("pin")
+
+ReadPin::ReadPin() {
+}
+
+ReadPin::ReadPin(uint16_t name) {
+    name_checksum = name;
+}
+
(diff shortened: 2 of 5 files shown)


src/main.cpp
```diff
@@ -190,6 +191,11 @@ void init() {
     #ifndef NO_UTILS_MOTORDRIVERCONTROL
     kernel->add_module( new MotorDriverControl(0) );
     #endif
+
+    auto rpp = new ReadPinPool();
```
wolfmanjm, line 195: This needs to be protected with `#ifndef NO_UTILS_READPIN` like all the other modules.

wolfmanjm (changes requested):

src/modules/utils/readpin/ReadPin.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+    // Settings
+    on_config_reload(this);
+}
+
+// Get config
+void ReadPin::on_config_reload(void*) {
+    my_pin.from_string(THEKERNEL->config->value(readpin_checksum, name_checksum, pin_checksum )->by_default("nc")->as_string())->as_input();
+    if (my_pin.connected())
+        my_pin.pull_none();
```
wolfmanjm, line 37: This should not be required it should use the standard pin definitions.

bullestock: @wolfmanjm you mean the pull_none() call? I kinda forgot about that, when I first tried the pin apparently had a pull-up.

src/modules/utils/readpin/ReadPin.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+ReadPin::ReadPin() {
+}
+
+ReadPin::ReadPin(uint16_t name) {
+    name_checksum = name;
+}
+
+void ReadPin::on_module_loaded() {
+    THEKERNEL->slow_ticker->attach(5, this, &ReadPin::pin_tick);
```
wolfmanjm, line 26: This is an absolute waste of CPU resources, just read the pin on demand not 5 times a second all the time.

wolfmanjm (changes requested):

src/modules/utils/readpin/ReadPin.cpp
bullestock: Yeah, I don't know what I was thinking there...

src/modules/utils/readpin/ReadPin.cpp
wolfmanjm: Yes, it is pullup by default you would define the pin as P1.2- to disable the pullup.
http://smoothieware.org/configuring-smoothie#advanced-pin-configuration

Zaaphod:
I have been interested in doing this as well.  I see there is output for 'Pin is not connected'  Is smoothie able to tell if an input is floating?  Or does that mean the pin is not defined?

wolfmanjm:
TBH this is not needed there is already a  `switch` command, and you can define a switch in config for an input instead, so I see this as a duplication of functionality  that uses up valuable flash space.

Zaaphod:
I thought the input switch could only be used to turn an output on or off,  can you somehow query the status of the switch to the host with it?

wolfmanjm:
Yes, you can set a switch to be an input and read it with `switch name`

Zaaphod:
Awesome that works!  I don't see it documented here http://smoothieware.org/switch unless I'm missing it.  It would be a good thing to add to examples,  I've had occasion to want the host to do something based on the condition of an external input and this would save the extra hardware I've been using.

bullestock:
Right, if that had been documented I wouldn't have bothered.

PR #1208, opened 2017-08-20
Use stepper acceleration for extruder only moves

hmflash:
For extruder only moves (retracts) use the extruders's acceleration and not min(extruder.acceleration, default_acceleration).

I have a geared bowden extruder that can run reasonably high accelerations (5000mm/s^2).  I have cura and slic3r configured to change my motion acceleration from 2000 for infill and 800 for perimeters.  I noticed that the `M204 S800` gcode was effecting my retraction speeds.  The current smoothie implementation uses the minimum of default_acceleration and the stepper's acceleration when scheduling moves.  This means my extruder acceleration is being dropped to 800.  The following gcode demonstrates the issue:

```
G91
M204 E5000 ; Extruder acceleration 5000
M204 S10 ; Silly slow acceleration
G0 X50 F150000 ; Accelerates slowly
G0 E-4 F2400 ; Expect to accelerate quickly, but is slow
```

With this change, the extruder only moves will just use the stepper's configured acceleration and not the minimum.

### DIFF
--- a/src/modules/robot/Robot.cpp
+++ b/src/modules/robot/Robot.cpp
@@ -1218,7 +1218,9 @@ bool Robot::append_milestone(const float target[], float rate_mm_s)
             float ma =  actuators[actuator]->get_acceleration(); // in mm/sec²
             if(!isnan(ma)) {  // if axis does not have acceleration set then it uses the default_acceleration
                 float ca = fabsf((d/distance) * acceleration);
-                if (ca > ma) {
+                // For moves with XYZ component, use minimum of default_acceleration and actuator acceleration
+                // For E only moves (retract), use the actuator acceleration
+                if (ca > ma || auxilliary_move) {
                     acceleration *= ( ma / ca );
                 }
             }


hmflash:
Forgot to mention, even without changing the acceleration via M204, if your smoothie config has:

```
acceleration 1000
extruder.hotend.acceleration 5000
```

Currently, retracts accelerate at 1000, my change has them accelerate at 5000.

wolfmanjm:
I see this was possibly a bug in the existing code, I'll see if there is  better way to fix it. Unfortunately this PR does not take into account auxiliary moves are not just extruder only moves (at least they are on 3d printers but not on CNC).

I think the intention in the original code was to do what you want but it used acceleration to compare with by mistake.  The intention is to use the scaled smaller of any acceleration, but in the case of only one actuator moving it would use the acceleration for that actuator. As the default is usually larger than any individual actuator acceleration it was not noticed.

wolfmanjm:
FWIW a workaround for now is to set the default acceleration to the maximum you want for E, then set the individual XYZ accelerations lower. This should do what you want with the current edge code.

src/modules/robot/Robot.cpp
```diff
@@ -1218,7 +1218,9 @@ bool Robot::append_milestone(const float target[], float rate_mm_s)
             float ma =  actuators[actuator]->get_acceleration(); // in mm/sec²
             if(!isnan(ma)) {  // if axis does not have acceleration set then it uses the default_acceleration
                 float ca = fabsf((d/distance) * acceleration);
-                if (ca > ma) {
+                // For moves with XYZ component, use minimum of default_acceleration and actuator acceleration
+                // For E only moves (retract), use the actuator acceleration
+                if (ca > ma || auxilliary_move) {
```
wolfmanjm, line 1223: The issue here is if we move A and B axis (which are both auxiliary) then it will end up with the larger acceleration of the two which would be incorrect.

wolfmanjm:
I think you need to use the workaround, and we need to update the docs to say default acc needs to be the largest of any axis accel.

Fixing this properly is turning out to be a lot harder than first thought. But the workaround should work.

The complexity is that you don't simply use the lowest acceleration, it has to be scaled by the ratio of the move of that axis.

hmflash:
Is asking the actuator if it is an extruder a better fix?  Something like this:

```
--- a/src/modules/robot/Robot.cpp
+++ b/src/modules/robot/Robot.cpp
@@ -1214,12 +1214,15 @@ bool Robot::append_milestone(const float target[], float rate_mm_s)
 
         // adjust acceleration to lowest found, for now just primary axis unless it is an auxiliary move
         // TODO we may need to do all of them, check E won't limit XYZ.. it does on long E moves, but not checking it could exceed the E acceleration.
-        if(auxilliary_move || actuator < N_PRIMARY_AXIS) {
+        if(auxilliary_move && actuators[actuator]->is_extruder()) {
+            float ma =  actuators[actuator]->get_acceleration(); // in mm/sec²
+            if(!isnan(ma)) {
+                acceleration = ma;
+            }
+        } else if(auxilliary_move || actuator < N_PRIMARY_AXIS) {
             float ma =  actuators[actuator]->get_acceleration(); // in mm/sec²
             if(!isnan(ma)) {  // if axis does not have acceleration set then it uses the default_acceleration
                 float ca = fabsf((d/distance) * acceleration);
                 if (ca > ma) {
                     acceleration *= ( ma / ca );
                 }
```

Using the workaround causes different acceleration to be computed for diagonal moves because of the axial component scaling.  The default acceleration is not scaled to each axis.  I don't know if this is desired or not.

Case 1: If the machine acceleration is set to 1000, a diagonal move will accelerate at 1000.

Case 2: If the machine acceleration is set to say 5000, and the X and Y steppers are set to accelerate at 1000, a diagonal move will accelerate at 1414.

hmflash:
Have another thought.  Let me get IRC setup so I'm not putting random thoughts in closed PRs.

PR #1209, opened 2017-08-21
updating to latest version 

dawsqeq:
* Minor cosmetic config changes

Added some config option informations, made a few minor cosmetic changes, added some links, removed a few options users shouldn't mess with anyway

* Fix the z-probe delta grid code for cartesian printers.

* Fix bug that did not allow the enable pin for a motor to be set to nc (as needed with SPI drivers)
Fix error messages in SPI drivers module
Fix HALT setting enable in SPI drivers in on_idle so it does not crash

* update firmware.bin

* Adapting config changes from cartesian to delta

Changes ( cosmetic ones ) were made to the cartesian config, I am porting them to the delta file. This was done using diff software, and does not change the content of the file just it's appearance.
I also added a delta grid configuration example.

* Updating laser config to the latest format

* Updating laser config to the latest format in delta config

* added example script of fast stream over USB

* Create fast-stream.py

example of how to fast stream over USB serial to smoothie

* check for halt condition

* handle target being in halt state

* fix type

* Redefine delta_grid when using is_square to be the half diagonal of the square defined by the larger of x_max or y_max.

* Clean up is_square config handling for delta grid
  max_x and max_y MUST be defined now when is_square is set

* when is_square is set return error if dimensions changed between config and saved grid

* use 64 bit fixed point for stepticker

* fix zprobe to deteect of config of LS failed.

* fix config entries for delta grid

* update firmware.bin

* fix switch momentary/toggle config setting

* Initial commit for cart-grid, or cartesian-grid.
Assumes that 0,0 is in the corner.

* Fixed descriptive text.

* fix some stuff modify other stuff

* change width length to x_size, y_size

* remove is_square from delta grid, but return a deprecated error instead for now

* Calculate the compensation grid over +X and +Y only (lower-left 0,0), instead of centered 0,0.

* remove uneeded routine in cart grid

* remove unused rotuiens in cart grid

* Update CartGridStrategy.cpp

* allow rectangular grid to use G32 as well as G31

* Grid size on the fly

* Grid size on the fly

* Compatibility fix

* Fix save and load to/from file

* G28 default grid size from configuration. Error messages for bigger that configured grid sizes.

* Compact grid sizes into struct

* revert .gitignore

* Typos fixed

* Stating the obvious : creating a code of conduct.

This is something that can be pointed at when people cross the line as happens sometimes. It might be useful to *some* people for who this isn't obvious, though I doubt it.

* Wor If current grid size NxM less or equal to configured.

* Start probing from (0;0) by removing navigation to bottom right corner left from DeltaGrid

* Add prebuild binary for hires

* add hores build

* Revert back coordinates format in DEBUG messages

* make default for get_axis size 3 not n_primary_axis

* Allow larger strings for M114 and get pos
check for sane values of number of axis

* add some more block debugging

* more debug stuff

* remove block debug

* Fix a bug where very slow feedrates could cause the step rate to be less than the stepticker period.
  Potential fix used here is to make sure that the minimum ticks for acceleration is 1. Not sure if that may have consequences.
  This showed up as a problem doing G1 X1 F1 where it would step at high rates and finish immediately

* Update PULL_REQUEST_TEMPLATE.md

* revert c10aa2c8c4e34fd77734da623d4022e680856760

* Added leveling region definition using start point and x and y size

* Added human redable format of current grid display

* Revert DEBUG format once again

* Fix edge case where accleration is too high for stepticker
 - fixes the G1 X1 F1 issue

* Revert "Fix edge case where accleration is too high for stepticker"

This reverts commit 3d2dd8f9485ed642f8809e678366b242aa1fd7df.

* fix null ptr deref in config porcessing

* free up old config value when a duplicate is found.

* release the include directive if found in config

* fix initial Z used for G98 drilling cycle

* I and J optional. If only one provided it will use it as value for second.

* Documetation correction

* Separate local and global variables

* Rever I/J omitting handling logic

* increase extruder position DP to 4

* Added support for PT100 using E3D amplifier.

* make M911.x more robust in advanced motor driver

* cap current sent to i2c to 255

* fix bug in setting jog feed rate from panel (cnc mode)

* reduce block size by removing the acceleration_per_tick and decleration_per_tick from block
refactor the prepare called from planner.
fix the block size reported by mem command

* fix reported block size from mem command

* minor optimization

* update pre built binaries

* added sample TMC26X spresadcycle setting

* update firmware.bin

* Add kill_button_toggle_enable to config settings
  - enable for an estop lock style button, halt will be cleared when unlocked and released

* Update CartGridStrategy.cpp

fix typo

* add keyboard interrupt handling to fast-stream.py
reset the current positions on clearing a HALT, this fixes any discrepancy when an abrupt abort is issued

* fix what happens when we abort the fast-stream

* add keyboard interrupt handling to fast-stream.py
reset the current positions on clearing a HALT, this fixes any discrepancy when an abrupt abort is issued

* remove unused variable in Kernel

* Junction deviation tweak

In the original code, junction deviation was not calculated at values less than 0.95 (~18.2 degrees).  Any gcode file that allows the machine to reach max speed in a straight line and then goes into a turn consisting of line segments less than 18 degrees of change between them will keep the machine at full speed throughout the turn.  If 10 very short segments are placed together in this manner, the machine can complete a 180 degree movement change at max set speed without decelerating through the turn.  The 0.95 value has been changed to 1.0 allowing the max junction speed calculation to work with any line segments that are not co-linear.  This makes the code match the description in the comment.

This also fixes an issue described in the smoothie forums: http://forum.smoothieware.org/forum/t-2146070/smoothie-takes-a-corner-too-fast.

* Update Planner.cpp

* changed to <= 0.9999

* fix typo

* optimize 64bit fixed point usage

* move the tick_info array into AHB0 to save memory

* add test for out of memory when allocating tickinfo
add tickinfo size to mem

* update hires firmware images

* Allocate block queue blocks in AHB0, and tickinfo in main memory

* change allocation to new[]

* cover other allocatin in heapring

* refactor...
  rename HeapRing to BlockQueue
  remove templates and make it Block specific
  as it is now only used for the Planner queue and allocates its queue in AHB0

* update bins

* cleanup header files

* Optimize stack usage for probe commands

* optimize print_position to use string and reduce stack usage.

* update hires bins

* Add an alarm if homing did not hit the endstops
fix ALARM: messages for grbl (it uses a space)
Attempt to fix halt condition detection in conveyor

* Allow ? to show realtime coordinates when homing

* rename bins

* change goto park issues in mm or inches depends on setting instead of switching to mm mode for the move

* fix some inchmode issues when homing

* fix the limit stop direction report

* fix arc_target size for G2/G3

* u[date bin

* only check max speeds for 3 axis

* fix another typo

* Make the etst pmoves not wait for idle between iterations, makes it smoother.

* FIx issue where a single temperature control with multiple extruders needs to have tool 0 selected.
Now temp setting will always work regardless of which extruder is set.

* allow get pos to show A axis for E1 and B for E2, M114 still shows E instead

* Revert "FIx issue where a single temperature control with multiple extruders needs to have tool 0 selected."

This reverts commit 1c9a010d073fe74288c035c790a03d296af15ad9.

* remove unused bool in temp control

* update firmware.bin

* update the latest bin versions

* Update abc-endstop.config

* plan9: Update error string names

Several of the error names are int values defined elsewhere. Prefix them
with `P9_` to prevent `plan9.cpp:27:12: error: expected unqualified-id
 before numeric constant`.

* remove old extruder config settings #defines

* when doing zprobes save state first then restore it

* update firmware-latest.bin

* Fix missing zprobe_offset

I added missing zprobe_offset values in the find_bed function to avoid the probe looking for the bed outside of the print surface. Now puts correctly the probe at 0,0.

* Add an AutoPushPop and RAII style to push state and pop state on exit of scope
Use above for the test commands so we don;t leave G0 feedrate set incorrectly

* add command 'switch name' to query the state of a named switch

* do not send > prompt for a non telnet client connection

* So streaming over network dos not hog all of main loop we now issue one net command per main loop iteration
this allows another telnet connection and/or a USB serial connection to actually get a response

* fix momentary input switch to reflect input state not toggle it

* reduce stack usage for cart grid G32 and zprobe

* Fix coordinated move

* remove debug printf

* updat firmnware

* Allow ABC axis to be homed on a delta

* update katest bin

* fix G28 A0 for deltas, should be able to home just the A (or B or C) axis

* initial attempt to implement a feedhold.
  this acts like the planner queue is full, but will not pause immediately as the existing stuff in the planner queue will need to complete
  in grbl-mode ! will pause and ~ will resume

* allow feed hold to be enabled independently of grbl mode, so in grbl mode it is set in 3d printer mode it is not set by default

* fix comment typo

* Add ABC or E to the ? query

* add zprobe.max_z config option (falls back to gamma_max if not defined)
fix delta homing for A only so Z does not move back when only homing A(BC)

* update firmware-latest.bin

* change prepare screen menu 'set home' to 'set X0Y0' to more properly reflect what it does and is supposed to do

* Fix rectangular grid find bed, which was broken

* fix error message

* add new_status_format for grbl1.1-like status format which is easier to parse and adds ABC axis to MCS, and F: and S: and L:

* Update config

* Update config

* Update config

* Update config

* Update config

* Update config

* close and reopen file every 400 bytes to avoid corruption

* also fix cart grid

* update firmware-latest

* Dwell before probing

Adds a configurable delay before probing

* Updated float comparison

In order for a Pull request to be accepted it must meet certain standards and guidelines.

In addition to following the [Coding-Standards](http://smoothieware.org/coding-standards) it is suggested the following be adopted.

* There must be a detailed description of the change, why it was added and what the use case is, and a link to the documentation to use it in the wiki if applicable.
* Do not commit changes to the FirmwareBin directory or any binary files in the repository. (this is hard to reverse once done)
* Only one feature per pull request, if you change multiple files with unrelated changes it will be rejected.
* Do not make gratuitous changes. If your coding style is different from the one adopted by most of the Smoothie codebase, do not make changes to suit your style. Peoples styles differ, and the new code you write may reflect your coding style, that is ok, but changing existing code to match your style is not ok. Doing so will delay the acceptance of your code, or have it rejected entirely.
* Maintain backwards compatibility. Do not make changes that will break other peoples systems, it is not acceptable to force everyone to change their config files to accommodate your change. If necessary add new configs to enable your change if it is breaking backward compatibility.
* Do not make changes that are specific to your setup. Smoothie needs to work in many different environments.
* Do not make changes that customize messages specific to your company or implementation. You are welcome to do this in your own fork and binary, it is not ok to try to merge that into the upstream repository.
* Test your changes is as many different configurations as you can, eg different arm solutions. In the Pull request state what tests have been done and what tests need to be done.
* Do not reformat the Config Samples to suit your style.
* Add examples of your config entries to the ConfigSamples/Snippets folder, only change existing config samples if absolutely necessary
* If you must correct existing comments (like correct spelling mistakes or typos) then submit them as separate pull requests
* Make sure your Pull request is against the current edge branch, and passes the Travis CI.
* All comments and code must be in English, this includes names of functions.

It may seem like a pain to follow these rules, but it takes the maintainers valuable time and effort to review and test pull requests, these guidelines help reduce that effort.


wolfmanjm:
I am sorry, but we have to reject this HUGE pr.

If you read the submsion guidelines for pull requests it says ONE change per PR with minimal changes. This is way too hard for the devs to go through and check.

PR #1211, opened 2017-09-01
Fix truncation of command output from panel probe screen on large displays

DavidJRobertson:
Bug: only a maximum of three lines of output are displayed on the panel's probe screen, even when the output is longer than can fit in three lines and there is more space on the display.


wolfmanjm:
Thank you.

PR #1215, opened 2017-09-03
Add remote temperature ExternalSensor support

ipittau:
This feature adds the support to a new "remote" temperature sensor, not directly connectected to Smoothieboard.
If you, for some reason, need more temperature sensors or to use sensors not supported by Smoothieboard you can use the ExternalSensor type and push the temperature, read externally, to Smoothieboard using a configurable M-command.
With ExternalSensor you can still take advantage of all features and automatisms of temperature control system of smoothie firmware.

A new config, set_temp_m_code, has been added to temperature_control in order to configure the M-command to set the temperature on the sensor.

Set the temperature is simple, you only have to call the configured M-command with the float value of the temperature. 
You maybe want to call this command constantly, to update the value frequently.

**How to configure an External Sensor:**
> temperature_control.heater.enable                true              
> temperature_control.heater.sensor                ExternalSensor     
> temperature_control.heater.set_temp_m_code       933 

**How to set the temperature:** 
> M933 S37.1

**What changed:**
- Added the ExternalSensor class: it simply stores the float value and returns it on get_temperature
- Added the set_temp_m_code on TemperatureControl: default value 933
- Added the management of configurable M-Code on TemperatureControl: basically calls set_temperature on the sensor.
- An emulated_sensor flag has been added to TemperatureControl to avoid call set_temperature on internal sensors.
- Added the config snippets examples

### DIFF
--- a/ConfigSamples/Snippets/external_sensor.config
+++ b/ConfigSamples/Snippets/external_sensor.config
@@ -1,0 +1,19 @@
+## Chamber heater configuration
+temperature_control.heater.enable                true               #
+temperature_control.heater.sensor                ExternalSensor     # Emulated sensor, the temperature is provided by UART
+temperature_control.heater.set_temp_m_code       933                # M code used to set the current temperature
+temperature_control.heater.designator            R                  #
+temperature_control.heater.max_temp              75                 # Set maximum temperature - Will prevent heating above
+temperature_control.heater.readings_per_second   1                  # Reduced sampling
+
+## Heated chamber configuration
+temperature_control.chamber.enable               true               #
+temperature_control.chamber.heater_pin           1.19               #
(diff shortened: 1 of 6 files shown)


ipittau:
Unfortunately testing the feature I bring up a side effect.
When the G-code 109 (set-and-wait) is issued, the temperature_control module blocks the G-code-dispatcher, preventing the execution of new commands. 
The temperature_control module, waits until the temperature reaches the target, but with we cannot update the temperature if we cannot push G-code commands.

A solution found is to:
-  avoid blocking the g-code-dispatcher
-  wait for the target on main loop
-  pause and resume the player of print files, to avoid going forward on printing process.

Of course better solutions are welcomed

wolfmanjm:
This is a show stopper and none of the suggested remedies are appropriate.

PR #1220, opened 2017-09-20
Windows DFU Support

DavidJRobertson:
Using DFU during development is convenient, however currently the `make upload` script does not work on Windows. This PR adds the dfu-util.exe binary and tweaks the makefile slightly so that `make upload` will work on Windows machines.

Note: In order for dfu-util to work with Smoothie on Windows, you need to install the appropriate driver (WinUSB). By far the easiest way to do this is to use a tool called Zadig (http://zadig.akeo.ie/). You will need to use Zadig to install the WinUSB driver twice: once for the "Smoothie DFU" device which is present during normal operation, and once for the lone "Smoothie" device which appears after you run `make upload`.

TODO: Could avoid this driver trouble by implementing WCID


wolfmanjm:
I'd prefer to NOT include any windows binaries here.

This belongs in the wiki with pointers to the exe files and instructions.

But thank you for the contribution

wolfmanjm:
However I'll look at incorporating the Make file changes once i check them out more closely.
I don't use windows neither do most developers so it is hard to test this out.

triffid:
Nice, I figured this should be possible when I wrote the DFU bootloader, just couldn't find any good information on the _how_.

Thanks for working it out!

This information definitely belongs in the wiki, and wolfmanjm is right - system tools don't belong in the source code repository, or we would already have the compile toolchain in here ;)

PS: would be better to replace cp with $(COPY) and define it appropriately rather than flip the entire command including filenames into a variable

PR #1221, opened 2017-09-23
Implemented basic FTP server

DavidJRobertson:
The current RFC913 SFTP server seems like an odd choice given the lacking support for RFC913 as it is an abandoned protocol.

This PR implements a basic RFC959 FTP server. This enables the use of a standard FTP client e.g. filezilla. As FTP has widespread support, this is likely to be much more useful than the current SFTP.

### DIFF
--- a/src/libs/Network/uip/Network.cpp
+++ b/src/libs/Network/uip/Network.cpp
@@ -23,6 +23,7 @@
 #include "webserver.h"
 #include "dhcpc.h"
 #include "sftpd.h"
+#include "ftpd.h"
 
 #ifndef NOPLAN9
 #include "plan9.h"
@@ -55,6 +56,7 @@ Network::Network()
     ethernet = new LPC17XX_Ethernet();
     tickcnt= 0;
     sftpd= NULL;
+    ftpd= NULL;
     hostname = NULL;
     plan9_enabled= false;
     command_q= CommandQueue::getInstance();
@@ -316,6 +318,9 @@ void Network::setup_servers()
 
     // sftpd service, which is lazily created on reciept of first packet
     uip_listen(HTONS(115));
+    
+    // ftpd service, which is lazily created on reciept of first packet
+    uip_listen(HTONS(FTP_CONTROL_PORT));
 }
 
 extern "C" void dhcpc_configured(const struct dhcpc_state *s)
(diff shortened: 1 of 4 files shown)


wolfmanjm:
Very nice, thank you.

However we do not have enough flash or memory to incorporate this into the mainline, which is why it was not done originally.

However this will almost certainly make it into v2 if there is not already a ftp server in nuttx.

thanks again. Nice work!

I think this should be linked in the wiki under third party branches, so anyone who has fewer modules enabled can enable this.

wolfmanjm:
Would there be a way to free up the ftp server when the connection is closed?
That would mitigate to some extent the huge amount of memory this uses once it has been activated.
We may still get out of memory crashes on initial connection though. I don't think there is anyway to avoid that as most people have very little memory to spare especially while printing.

wolfmanjm:
Actually I am re opening this as it is so good, hopefully it will get some testing and we can see how close to the edge it takes our memory usage ;)

DavidJRobertson:
The `control_conn_state` struct only exists in memory for the duration of a connection and the `data_conn_state` struct only exists for the duration of LIST/RETR/STOR commands. These structs contain the IO buffers which should be the majority of the memory consumption.

DavidJRobertson:
Yes, it does appear that using the FTP server while printing can crash the board due to running out of memory. With my setup, I was able to connect and list the root directory, but listing the contents of /sd caused a crash.

wolfmanjm:
I think that we need to lock out any ftp connections while the printer is active, we need to do this so people don't crash in the middle of a print. Also we probably need to inhibit prints while ftp is active, although I have no idea how to do that. The former is quite easy as we just make sure the system is_idle() before accepting a connection and reject the connection if it is not idle. We need to release the ftp server when the connection closes too to release all the memory.

DavidJRobertson:
Yes, I think you're right. Also, as is it can be painfully slow for large files. I'm not sure there's any easy solution as the limiting factor seems to be that you only get 346 bytes of TCP payload per packet due to the size of the uip buffer, and the structure of uip means the window size is limited to that number too. Lots of round trips for rather little data.

wolfmanjm:
As this crashes the system due to memory constraints it is being rejected. I am sorry it was a great patch, but the V1 smoothie just does not have the resources to handle this.

PR #1222, opened 2017-09-23
Download dfu-util automatically during install on Windows

DavidJRobertson:
Following from #1220

I have removed the binary `dfu-util.exe` and instead modified `win_install.cmd` to download a copy on install.


wolfmanjm:
I am still not happy with downloading windows binaries from shady sources due to viruses.
I'd prefer to leave it as instructions on the wiki and allow people to risk downloading executables for windows at their discretion.

DavidJRobertson:
I'm unsure why downloading dfu-util is any different in that regard to downloading gcc. 

The file is checked against an MD5 in the install script so if the file available at that URL were to change in the future, it would cause the script to fail.

DavidJRobertson:
I have added installation instructions to the wiki: http://smoothieware.org/compiling-smoothie#flashing-via-dfu-usb-device-firmware-update

However, the makefile will still have to be adjusted to replace the call to cp with one to copy.

DavidJRobertson:
#1223

PR #1227, opened 2017-10-08
Implement a post enable delay for stepper drivers

NeilBetham:
For a little background on this change, i'm currently running a Smoothieboard 4x v1.1 with GeckoDrives, G203V, as the stepper drivers. I've run into an issue where just after a rising edge on the enable signal the drives won't step for a short period of time and then they will start stepping at the speed of the move,
and subsequently stalled, and then ramp down to a stop. This gets to be especially annoying when trying to home the machine as you have to either wait for the home procedure to fail or reset the machine then do a couple short moves to clear the enable issue first then home the machine. With the added delay between enable and the beginning of stepping there is enough time to allow the driver to initialize and step properly.

For this pull request i'd like to get some feedback on how I implemented the change. Specifically:
* Is there a better way to halt execution then calling `wait`?
  * I'm not as familiar with architecture of the main run loop, from what I could tell there didn't seem to be an easy way to halt execution of movement specific code. Mainly I'd rather have the uC still running services like USB and network.
* Is there a better spot to put the wait?
  * Right now each stepper channel has to wait for the given time period instead of a one shot after every channel has been enabled. This isn't so bad considering it only happens after the drivers get enabled initially, but it's not optimal. I did look into putting the wait else where but by the design the only spot that really knows the state of the enable is the `StepperMotor` class.

I also added the wait time to the configuration so the delay can be adjusted to suit the needs of different external drivers. I have tested this with and without a wait and it solves the issue i'm running into.

Thanks,
Neil

### DIFF
--- a/src/libs/StepperMotor.cpp
+++ b/src/libs/StepperMotor.cpp
@@ -9,9 +9,15 @@
 #include "Kernel.h"
 #include "MRI_Hooks.h"
 #include "StepTicker.h"
+#include "checksumm.h"
+#include "Config.h"
+#include "ConfigValue.h"
 
 #include <math.h>
 #include "mbed.h"
+
+#define stepper_driver_enable_delay_checksum    CHECKSUM("stepper_driver_enable_delay")
+
 
 StepperMotor::StepperMotor(Pin &step, Pin &dir, Pin &en) : step_pin(step), dir_pin(dir), en_pin(en)
 {
@@ -29,6 +35,7 @@ StepperMotor::StepperMotor(Pin &step, Pin &dir, Pin &en) : step_pin(step), dir_p
     acceleration= NAN;
     selected= true;
     extruder= false;
+    enable_delay= THEKERNEL->config->value(stepper_driver_enable_delay_checksum)->by_default("0.0")->as_number();
 
     enable(false);
     unstep(); // initialize step pin
@@ -59,10 +66,21 @@ void StepperMotor::on_enable(void *argument)
     uint32_t bm= (uint32_t)argument;
     if(bm == 0x01) {
         enable(true);
-
     }else if(bm == 0 || ((bm&0x01) == 0 && ((bm&(0x02<<motor_id)) != 0)) ) {
         enable(false);
     }
+}
+
+void StepperMotor::enable(bool state)
+{
+    bool state_rising_edge = (state == true) && (state != is_enabled());
+    en_pin.set(!state);
+    if (enable_delay){
+      if(state_rising_edge){
+        wait(enable_delay);
+      }
+    }
+
 }
 
 void StepperMotor::change_steps_per_mm(float new_steps)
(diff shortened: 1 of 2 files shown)


wolfmanjm:
No calling wait is not allowed at all.

src/libs/StepperMotor.cpp
```diff
@@ -59,12 +66,23 @@ void StepperMotor::on_enable(void *argument)
...
     }
 }
 
+void StepperMotor::enable(bool state)
+{
+    bool state_rising_edge = (state == true) && (state != is_enabled());
+    en_pin.set(!state);
+    if (enable_delay){
+      if(state_rising_edge){
+        wait(enable_delay);
```
wolfmanjm, line 80: Wait is not a call allowed ever anywhere in smoothie as it blocks everything

wolfmanjm (changes requested):

NeilBetham:
That's kinda what I figured. Any suggestions on alternate ways to add a delay after an enable?

arhi:
> Any suggestions on alternate ways to add a delay after an enable?

Why don't you just configure alpha_en_pin, beta_en_pin etc.. in the config (they can all be linked to a same pin or separate ones) and then use M17 and M18 to enable/disable drivers... so instead of doing it in "stepper loop" you do it in G-code... you setup your startup code to something like 

M17 ; enable motors
G4 P100 ; wait 100ms for drivers to initialize
G28 ; home
...

wolfmanjm:
Yes I think it is not needed to modify smoothie at all for this.

Simply enable the steppers in the gcode file you are about to run, then do a pause...

```
M17
G4 P2
etc
```

src/libs/StepperMotor.cpp
```diff
@@ -29,6 +35,7 @@ StepperMotor::StepperMotor(Pin &step, Pin &dir, Pin &en) : step_pin(step), dir_p
     acceleration= NAN;
     selected= true;
     extruder= false;
+    enable_delay= THEKERNEL->config->value(stepper_driver_enable_delay_checksum)->by_default("0.0")->as_number();
```
wolfmanjm, line 38: Stepper motors are not modules and therefore are not allowed to read the config or set stuff from the config.

NeilBetham:
I was hoping to avoid having to write GCode or macros to fix this issue as it still leaves the onus on the user to fix the problem.

wolfmanjm:
It really is not a general problem that needs to be solved IMO. The gcode workaround for this edge condition seems suitable, as you can have slicers/cad cam gcode generators insert the code into the start up automatically.

I will close this PR, as it is not really appropriate, and there are too many errors in it to be accepted into mainline.
Thank you for the contribution though.

NeilBetham:
Also, it might be worth it to put some guidelines or checking script for what components and functionality can be used where so that people new to the project can get a sense of what rules exist for implementation before submitting pulls.

NeilBetham:
For anyone finding this after the fact, I've reworked the changes such that it no longer blocks execution while waiting for the drivers to enable and only executes the delay once instead of per driver. You can find the changes [here](https://github.com/NeilBetham/Smoothieware/commit/178cc220b66b75cd2bbd5ef801138412595b8170).

PR #1230, opened 2017-10-24
Added COPYING file

oddstr13:
This file is required for properly applying the GNU GPLv3 to the project.
It is a plain copy of https://www.gnu.org/licenses/gpl-3.0.txt, but please verify this yourself.

This also has the added benefit of having GitHub recognize the project license.
Screenshot

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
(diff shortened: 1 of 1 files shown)


wolfmanjm:
@arthurwolf ?

arthurwolf:
I don't like having two of the license it's redundant. But if Github recognizes based on that file, I guess it's overall a good idea. 
I'd personally be ok with merging.

PR #1231, opened 2017-10-25
Add M211.1 support for traditional endstop hard limit manipulation

newt0ns:
# Add M211 support for traditional endstops.

This was needed for a machine I was converting (Stratasys Dimension Elite). The z probe is deployed by moving the head into specific positions at the back of the machine. This involves moving the probe towards X-min to deploy the probe, and X-max to retract it. The retract point lives just beyond the X-Max trigger area (optical sensor with a plate that triggers ~25mm before physical limits). So there was a need to temporarily disable the x-max check for every print, but only for this one function.

Usage example, partial g-code start script:
M80 ;power on
**M211 S1 X1 Y1 Z0** ;Ensure endstops are enabled before homing. Machine homes to X/Y max and Z min
G91 ;relative positioning
G0 Z10 F5000 ;move Z down from wherever it is
G90 ;absolute positioning
G28 X0; home X axes
G0 X100 F12000 ; center X to clear probe storage post if probe was left deployed
G28 Y0 Z0 ;home y
G1 X-30 Y240 F5000 ;deploy probe
G0 X100 F15000 ; center X
T0 ;Ensure we're using T0 offsets
G32 ;probe bed
G0 X100 Y240 F15000 ;center x, move y to probe storage area
**M211 S0 X1** ;Disable x-max endstop
G1 X235 F5000 ;Store probe (this is just beyond the xMax endstop)
G0 X100 F15000 ;center x
**M211 S1 X1** ;enable x-max endstop again
G28 Z0

### DIFF
--- a/src/modules/tools/endstops/Endstops.cpp
+++ b/src/modules/tools/endstops/Endstops.cpp
@@ -1098,23 +1098,24 @@ void Endstops::on_gcode_received(void *argument)
                 gcode->stream->printf(" will take effect next home\n");
                 break;
 
-            case 211: { //Enable or disable endstop(s) hard limit
-                if(gcode->has_letter('S')) { //Alter endstop behavior    
-                        for(auto& e : endstops) {
-                            if(gcode->has_letter(e->axis)) {
-                                if(e->is_max_stop == gcode->get_value(e->axis)) {  //Are we configuring a min or max endstop?
-                                    string name;
-                                    name.append(1, e->axis).append(e->is_max_stop ? "_max" : "_min");
-                                    gcode->stream->printf("%s: %s ", name.c_str(), (e->limit_enable?"enabled":"disabled"));
-                                    THECONVEYOR->wait_for_idle();
-                                    e->limit_enable = gcode->get_value('S')>0;
-                                    if(!limits_enabled && e->limit_enable) {
-                                        register_for_event(ON_IDLE);
-                                        limits_enabled = true;
+            case 211: //Enable or disable endstop(s) hard limit
+                if(gcode->subcode == 1) { //M211.1 for hardware endstop manipulation
+                    if(gcode->has_letter('S')) { //Alter endstop behavior
+                            for(auto& e : endstops) {
+                                if(gcode->has_letter(e->axis)) {
+                                    if(e->is_max_stop == (gcode->get_value(e->axis)>0)) {  //Are we configuring a min or max endstop?
+                                        string name;
+                                        name.append(1, e->axis).append(e->is_max_stop ? "_max" : "_min");
+                                        THECONVEYOR->wait_for_idle(); //Wait for movement to complete before altering endstops
+                                        e->limit_enable = (gcode->get_value('S')>0);
+                                        gcode->stream->printf("%s: %s ", name.c_str(), (e->limit_enable?"enabled":"disabled"));
+                                        if(!limits_enabled && e->limit_enable) {
+                                            register_for_event(ON_IDLE);
+                                            limits_enabled = true;
+                                        }
                                     }
                                 }
                             }
-                        }
                     }
                     else { //Print current status
                         for(auto& e : endstops) {
@@ -1122,9 +1123,9 @@ void Endstops::on_gcode_received(void *argument)
                             name.append(1, e->axis).append(e->is_max_stop ? "_max" : "_min");
                             gcode->stream->printf("%s: %s ", name.c_str(), (e->limit_enable?"enabled":"disabled"));
                         }
-                    }
+                    }                    
                     gcode->add_nl = true;
-                }
+                }                
                 break;
 
             case 306: // set homing offset based on current position
--- a/src/modules/tools/endstops/Endstops.cpp
+++ b/src/modules/tools/endstops/Endstops.cpp
@@ -154,7 +154,6 @@ bool Endstops::load_old_config()
         ENDSTOP_CHECKSUMS("gamma")    // Z
     };
 
-    bool limit_enabled= false;
     for (int i = X_AXIS; i <= Z_AXIS; ++i) { // X_AXIS to Z_AXIS
         homing_info_t hinfo;
 
@@ -207,7 +206,7 @@ bool Endstops::load_old_config()
 
             // limits enabled
             info->limit_enable= THEKERNEL->config->value(checksums[i][LIMIT])->by_default(false)->as_bool();
-            limit_enabled |= info->limit_enable;
+            limits_enabled |= info->limit_enable;
         }
 
         homing_axis.push_back(hinfo);
@@ -218,7 +217,7 @@ bool Endstops::load_old_config()
 
     get_global_configs();
 
-    if(limit_enabled) {
+    if(limits_enabled) {
         register_for_event(ON_IDLE);
     }
 
@@ -241,8 +240,6 @@ bool Endstops::load_old_config()
 // Get config using new syntax supports ABC
 bool Endstops::load_config()
 {
-    bool limit_enabled= false;
-
     std::array<homing_info_t, k_max_actuators> temp_axis_array; // needs to be at least XYZ, but allow for ABC
     {
         homing_info_t t;
@@ -294,7 +291,7 @@ bool Endstops::load_config()
 
         // are limits enabled
         pin_info->limit_enable= THEKERNEL->config->value(endstop_checksum, cs, limit_checksum)->by_default(false)->as_bool();
-        limit_enabled |= pin_info->limit_enable;
+        limits_enabled |= pin_info->limit_enable;
 
         pin_info->is_max_stop = THEKERNEL->config->value(endstop_checksum, cs, limit_position_checksum)->by_default("min")->as_string() == "max";
 
@@ -366,7 +363,7 @@ bool Endstops::load_config()
     // sets some endstop global configs applicable to all endstops
     get_global_configs();
 
-    if(limit_enabled) {
+    if(limits_enabled) {
         register_for_event(ON_IDLE);
     }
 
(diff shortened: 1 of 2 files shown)


wolfmanjm:
There are some issues with this implementation that I think need to be cleared up before it is merged...

1. If limits are not initially enabled at config, then M211 will not actually enable the requested axis.
2. There can be min and max limits set, however this gets confused by homing and limits by using the homing direction to decide which one to set, when homing direction has no relation to limit switches. There needs to be a way to enable min and max limits and not consult homing direction. (Maybe use a bitmap so S1 is min S2 is max and S3 is min and max, or X1, X2 ,X3 to select min, max, min&max.
3. A Minor issue is that the formatting is not the same as the rest of the smoothie code.

You seem a little confused between homing and limits I commented the lines, but basically enabling or disabling limits for homing is not required. You are not enbling or disabling endstops as the comments imply, you are just enabling/disabling hard limits (if they were enabled in the first place) which does not affect homing at all.

wolfmanjm:
Additionaly M211 should  wait for idle so it is synchronized with previous moves, your example uses G4  P500 (M400 would be more appropriate) however most commands in smoothie that require to be synchronous call wait_idle() there should not be a requirement to use M400

src/modules/tools/endstops/Endstops.cpp
```diff
@@ -1085,6 +1085,36 @@ void Endstops::on_gcode_received(void *argument)
                 gcode->stream->printf(" will take effect next home\n");
                 break;
 
+            case 211: { //Enable or disable endstop(s)
```
wolfmanjm: this is not enabling endstops it simply enables or disables hard limits.

wolfmanjm (changes requested):

src/modules/tools/endstops/Endstops.cpp
```diff
@@ -1085,6 +1085,36 @@ void Endstops::on_gcode_received(void *argument)
...
 
+            case 211: { //Enable or disable endstop(s)               
+                    if(gcode->has_letter('S')) //Alter endstop behavior
+                    {
+                        bool enable = gcode->get_value('S')>0;
+    
+                        for(auto& h : homing_axis) {                            
+                            if(gcode->has_letter(h.axis))
+                            {
+                                if(h.home_direction == !gcode->get_value(h.axis)) //Are we configuring a min or max endstop?
```
wolfmanjm: homing direction has nothing to do with hard limits

newt0ns:
Great points, will fix now. On vaca so I'm trying to get my repo cleaned up and my patches applied and offered back for others. Did these a while ago, so apparently I'll need to be more careful about ensuring these weren't rushed just to get my printer doing what I needed.....

1:  Will add a registration check and register for the event if not already

2:  Totally agree, not sure why I assumed it was the way to go. Adding a simple flag is_max_limit set during config to tell which direction this stop is used for, seems easy enough to do with the old syntax, will try to find a clean way with the new (maybe loop through all of the endstops after reading the config to try to determine which end it lives on via homing directiona? Seems we may need to add a direction definition in the config to better define these.).

3: Will check, thought I was matching the other gcode handlers in endstops, but will verify.

wolfmanjm:
I'll reiterate;) homing direction has absolutely NOTHING to do with the limits. You need to be able to explicitly define which limit switch you are enabling (or disabling). Adding another flag is not needed.
There are homing axis and there are endstops, any endstop can be set as a limit regardless of whether it is used for homing or not. Please do not confuse homing with limits they are two totally different beasts.

newt0ns:
Totally understand that, but what we need for the M211 call is the ability to know which end of the axis that endstop resides on. The old style config has that ability (you can do a "info->is_max_stop = j == MAX_PIN;" in load_old_config() and you should be able to determine direction (untested, but seems like it should work). The new style config is ambiguous to end stop direction, as far as I can see. There is no way to determine if it's a min or max. So setting a M211 S0 X1 would not know which x axis stop was the max stop we are trying to disable (without guessing by checking if it's a homing direction endstop or not).

So something like this seems to be in order, unless there's something blatantly obvious I'm missing (quite likely):
// new config syntax
// endstop.xmin.enable true
// endstop.xmin.pin 1.29
// endstop.xmin.axis X
// endstop.xmin.homing_direction home_to_min
// endstop.xmin.limit_enable true
// endstop.xmin.limit_position min
//
// endstop.xmax.enable true
// endstop.xmax.pin 1.27
// endstop.xmax.axis X
// endstop.xmax.limit_enable true
// endstop.xmax.limit_position max

wolfmanjm:
Sorry this is not the solution, I gave you the preferred solution I'd prefer that solution (where you explicitly set the min or max in M211. I'm rejecting the addition of more config syntax, sorry,

src/modules/tools/endstops/Endstops.cpp
```diff
@@ -86,6 +96,7 @@ enum DEFNS {MIN_PIN, MAX_PIN, MAX_TRAVEL, FAST_RATE, SLOW_RATE, RETRACT, DIRECTI
 #define max_travel_checksum                CHECKSUM("max_travel")
 #define retract_checksum                   CHECKSUM("retract")
 #define limit_checksum                     CHECKSUM("limit_enable")
+#define limit_position_checksum            CHECKSUM("limit_position")
```
wolfmanjm: Rejecting the addition of an unneeded extra config option.

src/modules/tools/endstops/Endstops.cpp
```diff
@@ -1085,6 +1098,35 @@ void Endstops::on_gcode_received(void *argument)
                 gcode->stream->printf(" will take effect next home\n");
                 break;
 
+            case 211: { //Enable or disable endstop(s) hard limit
+                if(gcode->has_letter('S')) { //Alter endstop behavior    
+                        for(auto& e : endstops) {
+                            if(gcode->has_letter(e->axis)) {
+                                if(e->is_max_stop == gcode->get_value(e->axis)) {  //Are we configuring a min or max endstop?
```
wolfmanjm: You can continue to do this in your fork if you wish. However it will not be accepted as a PR if it is continued to be done this way.. Sorry. There should be no link to if it is min or max, this needs to be explicitly set in the M211 syntax by one of the methods I have already suggested

src/modules/tools/endstops/Endstops.cpp
```diff
@@ -186,14 +197,16 @@ bool Endstops::load_old_config()
             // add index to the homing struct if this is the one used for homing
             if((hinfo.home_direction && j == MIN_PIN) || (!hinfo.home_direction && j == MAX_PIN)) hinfo.pin_info= info;
 
+            info->is_max_stop = j == MAX_PIN;
```
wolfmanjm, line 200: Whether a limit is min or max has nothing to do with the homing direction, you need to totally disconnect this concept of homing direction with limit switches

src/modules/tools/endstops/Endstops.cpp
wolfmanjm: is_max_stop is a bool it is not good practice to compare with a float. I have to think about if we need this is_max_stop or not. up until now limits didn't care if they were min or max, it was irrelevant. We determined which one was hit by the direction the carriage was moving when one was triggered.

newt0ns:
So, I'm a little confused by what it is you say you want. As I am already following a similar syntax you suggested. The problem I have is defining the endstop mapping, there is only an axis it is mapped to, there is no direction attached to that endstop (min or max). Thus I was originally (and incorrectly) trying to use homing directions to determine what we were using, I have since move to a config option for the new syntax, and using the MAX_PIN checksum on the old style config. I'm all ears for a better method to determine which end of the axis the endstop is attached. 

The syntax for M211, as I have implemented it allows for specification of exactly which endstop you are trying to manipulate. I'm using the same syntax as listed here: http://reprap.org/wiki/G-code#M211:_Disable.2FEnable_software_endstops

If there is another spec we want to adhere to, I'd be happy to. As it is now, if you want to enable the x-max endstop it would be:

M211 S1 X1

to disable:

M211 S0 X1

to enable the x-min and y-max:

M211 S1 X0 Y1

wolfmanjm:
I retracted a couple of comments as I did not immediately see what you were doing. There seem to still be paces you are mixing limits with homing direction. Also how you determine which endstop to enable is dubious.

src/modules/tools/endstops/Endstops.cpp
newt0ns: That was my mistake, I previously tested min or max via:  "gcode->get_value(e->axis)!=0" to convert to a bool, not sure when I undid that...

Fixing now, will wait to hear if there's a better way to determine min/max location of the stop before any further commits.

wolfmanjm:
Ok I see what you are doing now. The M211 syntax is fine. The way you are doing it is not fine.

I can live (for now) with the extra config syntax to say which limit is min and max although I am not happy with that as it was not needed before and complicates the config and is no longer backwards compatible. if i think of a better way (or if it is even needed) I'll comment.

wolfmanjm:
also the M211 needs to be changed to M211.1 as M211 is used to enable/disable soft limits in the soft limit PR #1149 
M211.1 would be for hard limits, and M211 would remain for soft limits.

newt0ns:
Made the requested changes, cannot test for a few days though (should have packed a board ;)).

src/modules/tools/endstops/Endstops.cpp
wolfmanjm: I guess this is the only way to do it with the old syntax

src/modules/tools/endstops/Endstops.cpp
```diff
@@ -1098,33 +1098,34 @@ void Endstops::on_gcode_received(void *argument)
...
-                                    e->limit_enable = gcode->get_value('S')>0;
-                                    if(!limits_enabled && e->limit_enable) {
-                                        register_for_event(ON_IDLE);
-                                        limits_enabled = true;
+            case 211: //Enable or disable endstop(s) hard limit
+                if(gcode->subcode == 1) { //M211.1 for hardware endstop manipulation
+                    if(gcode->has_letter('S')) { //Alter endstop behavior
+                            for(auto& e : endstops) {
+                                if(gcode->has_letter(e->axis)) {
+                                    if(e->is_max_stop == (gcode->get_value(e->axis)>0)) {  //Are we configuring a min or max endstop?
```
wolfmanjm: I still do not like this, the code is obfuscated and not clear what you are doing. Also it is still not safe to compare a float with 0. To do that safely you need to use gcode->get_uint(..).
I had to read this several times to understand what it is supposed to do, and as such it means others will not have a clue! Sometimes it is better to be verbose rather than do some neat coding trick, especially when others have to read the code.

newt0ns: Understood, will verbosify and uint it.

newt0ns:
Ok, updated, tested code with 'old' config syntax, works as expected. Should be able to get the new syntax tested tomorrow.

src/modules/tools/endstops/Endstops.cpp
```diff
@@ -1101,27 +1101,29 @@ void Endstops::on_gcode_received(void *argument)
...
-                                        gcode->stream->printf("%s: %s ", name.c_str(), (e->limit_enable?"enabled":"disabled"));
-                                        if(!limits_enabled && e->limit_enable) {
-                                            register_for_event(ON_IDLE);
-                                            limits_enabled = true;
-                                        }
+                        bool enable_limit = gcode->get_uint('S')>0;
+                        for(auto& e : endstops) {
+                            if(gcode->has_letter(e->axis)) {
+                                bool is_max_axis_endstop = gcode->get_uint(e->axis)>0; //M211.1 S1 X1 would enable x-max endstop limit, M211.1 S1 X0 would enable x-min endstop
+                                if(e->is_max_stop == is_max_axis_endstop) {
```
wolfmanjm, line 1108: this IMO is no better wrt understandability, the main issue is doing an equality between two booleans when really it should be a logic test. I admit this is concise, but it is obfuscated. No need to change it, but understand my issue.

newt0ns:
Have been printing with this for several days on the new config syntax as well, no issues.

src/modules/tools/endstops/Endstops.cpp
newt0ns: Understood, thanks

wolfmanjm:
closing this as it no longer merges cleanly.

PR #1234, opened 2017-11-18
Align get_pid_settings() with report_settings()

cilynx:
In the handler for M958 in `SpindleControl`, a function,
`get_pid_settings()`, is called which does not appear to exist beyond
its virtual initialization also in `SpindleControl`.  Because this
function does not really exist, M958 does not return any PID data.

`PWMSpindleControl` contains a function `report_settings()`, which
prints the PID data one would expect from M958.  Replacing
`SpindleControl`'s references to the non-existent `get_pid_settings()`
with `report_settings()` causes M958 to return PID data as expected.


wolfmanjm:
Thanks

PR #1235, opened 2017-11-18
Feature/spindle max pwm

cilynx:
Treadmill motor controllers are becoming increasingly popular in DIY
machines.  One such popular controller is the mc2100 which expects a
duty cycle of 0-85% for 0-100% speed.  To keep the PWM signal below the
controller maximum and thus prevent the controller from going into an
error state when higher speeds than physically possible are requested,
we need a config option for the ceiling on the PWM signal.

I tested this for backwards compatibility with my original configuration
without this new parameter included and it works as expected, using 100%
as the maximum duty cycle.

I tested setting `spindle.max_pwm` to 0.85 (above the hard-coded 0.5
starting point) and 0.25 (below the hard-coded 0.5 starting point) and
confirmed expected function for both via M957 output and viewing the
physical PWM output with a scope.

I added a sample `spindle.config` showing this new parameter in context
with the other spindle parameters I could find with a quick grep of the
code and search through the docs.

### DIFF
--- a/ConfigSamples/spindle.config
+++ b/ConfigSamples/spindle.config
@@ -1,0 +1,34 @@
+# Spindle control settings
+
+spindle.enable            true           # [Default false]   Set this to false to disable the spindle module
+
+# PWM spindle settings
+
+spindle.type              pwm            # [Default pwm]     Spindle module mode (pwm, analog, modbus)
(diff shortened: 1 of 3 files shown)


cilynx:
Whoops -- I put the `spindle.config` in the `ConfigSamples/` root instead of `ConfigSamples/Snippets/` where other similar samples live.  @wolfmanjm -- do you want to move it or would you like another PR?

Issue #1069, opened 2016-11-27
Spindle does not turn off when ALARM occurs


arthurwolf:
#### What were you doing?

Using the spindle module to control a VFD via Gcode

#### What did you expect to happen?

ALARM occured and I expected the spindle to stop. I killed a 50 euro probe not realizing it was still on when starting the job again :)

#### Branch & Commit or Version of Smoothieware

Any recent edge version

( this is mostly a note to myself, I hope to have time to fix this soon )

wolfmanjm:
fixed by #1236

PR #1236, opened 2017-11-19
Fix/spindle off on halt

cilynx:
When a machine enters a `HALT` state for any reason, the spindle should
stop for safety along with everything else.  This commit changes the
default behavior to `turn_off` the spindle `on_halt`.  I added an
optional `spindle.ignore_on_halt` config parameter that defaults to
`false` in case anyone needs the previous less-safe behavior.

I believe this addresses #1069.

I confirmed functionality on `PWMSpindleControl` via `M957` as well as a
physical scope while causing `HALT` via `M112` or `^X` over the serial
interface.

I found some weirdness in `AnalogSpindleControl` around `report_speed`,
but I'll address that in another PR.  I did confirm that `turn_off` is
being called `on_halt` as it should be and the PWM duty cycle is reduced
to 0 even though `M957` (`report_speed`) continues to report that the
spindle is moving.

Without having the equipment for the other side of the RS485 connection,
I can't do much to test `ModbusSpindleContol`, but
`HuanyangSpindleControl`, the only supported Modbus VFD right now,
implements `turn_off`, so I don't see any reason it shouldn't work fine
with this change.

### DIFF
--- a/ConfigSamples/Snippets/spindle.config
+++ b/ConfigSamples/Snippets/spindle.config
@@ -1,6 +1,7 @@
 # Spindle control settings
 
 spindle.enable            true           # [Default false]   Set this to false to disable the spindle module
+#spindle.ignore_on_halt    true           # [Default false]   Don't stop the spindle on HALT.  Not recommended unless you really know what you're doing.
 
 # PWM spindle settings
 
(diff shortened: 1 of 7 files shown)


wolfmanjm:
I am surprised this works, as you need to register for the ON_HALT event, and I do not see that anywhere.
Also there seems to be some code cleanup going on and generally we prefer that in a separate PR.

wolfmanjm:
Ok nevermind, due to the afore mentioned code cleanup i missed where the ON_HALT was being registered. That is why we prefer code cleanup as  a separate issue as it obscures what the actual cited change is.

cilynx:
Apologies -- will separate moving forward.

PR #1239, opened 2017-12-06
fix / dwell before probing

adrianjobst:
moved dwell delay time to the beginning of the run_probe function, so the if clause will not cancel the function before the dwell ever happened.

Imho this is necessary otherwise vibrations from any previous movement will lead to cancellation before the delay can be executed when you use super sensitive sensors like the precision piezo solution.

I tested this on my hypercube evolution with precision piezo probe.


sdavi:
Yep, good catch. This is where the test should be moved to.

PR #1240, opened 2017-12-08
A heater will wait until a temperature is reached, even when cooling down, if 'R' is used rather than 'S' for set_and_wait_m_code commands

michaelnew:
**Basic Idea**
Currently if you use a "set temp and wait" command (heating the bed with `M190 S50` for example) the printer will wait until the temperature is reached before proceeding to the next line of gcode. If you follow that command with `M190 S40` the printer will set the bed temperature to 40C but it will **not** wait for the temperature to be reached before proceeding. The logic is essentially `while (current_temp < target_temp) idle; `, so if the temperature is already higher than the target temp it just moves on immediately.

This PR adds support for commands like `M190 R40`. If the current temperature is lower than the target it will behave the same way as `M190 S40`, but if the current temperature is higher than the target temperature it will wait until the heater cools down to 40C before proceeding. This is how Marlin behaves, at least according to the docs. I don't have a Marlin printer to test on.

**Use Case**
I needed this because I have an OctoPrint plugin that removes the part from the bed and then starts a new print, but it can only remove the part once the bed has cooled down. I'm sure there are other scenarios where this would be important, such as using a multi-material single extruder setup where you'd want to explicitly wait for a cool down before switching materials.

**Other Thoughts**
It might make sense to just have 'S' commands wait for a heater to cool down before proceeding and not even bother with the 'R' thing. Conceptually that makes more sense to me since I would expect "set temperature and wait" to wait for that temperature to be reached, even if it's cooling down. The current behavior is really this (using bed heater examples, but the same applies for the hotend):

`M140 S50` Set the bed temperature and proceed immediately.
`M190 S50` Set the bed temperature. Proceed immediately if the current temp is equal or higher, otherwise wait.
`M190 R50` Set the bed temperature and and wait until it's reached (added by this PR).

I imagine though that people do make use of the "set temp and wait, but only if the heater is too cold" logic, so changing the behavior of an existing command could cause some issues. I also think there's some value in keeping things consistent with other flavors of gcode, like Marlin.

**Implementation Details**
Right now the heater is turned off completely for a 'cooldown and wait' command, and then set to the target temperature after it's been reached. I originally tried setting `set_desired_temperature` to the exact target, but the heater would float a couple degrees higher than the target and never quite reach it. It might make more sense to change the PID logic for when it's cooling down rather than just turning off the heater, but I don't know enough about that to mess with it.

This approach could be totally off; I just grabbed the code and made changes because I needed this to work. I'm happy to test or try out other ideas. I do however think this is a pretty essential feature, no matter how it's implemented.

### DIFF
--- a/src/modules/tools/temperaturecontrol/TemperatureControl.cpp
+++ b/src/modules/tools/temperaturecontrol/TemperatureControl.cpp
@@ -325,7 +325,7 @@ void TemperatureControl::on_gcode_received(void *argument)
                 }
             }
 
-        } else if( ( gcode->m == this->set_m_code || gcode->m == this->set_and_wait_m_code ) && gcode->has_letter('S')) {
+        } else if ( (gcode->m == this->set_m_code || gcode->m == this->set_and_wait_m_code ) && (gcode->has_letter('S') || gcode->has_letter('R')) ) {
             // this only gets handled if it is not controlled by the tool manager or is active in the toolmanager
             this->active = true;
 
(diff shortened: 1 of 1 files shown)


wolfmanjm:
This was submitted quite a while back and caused issues so is marked as needs testing. #1052

We have the temperature switch module which usually handles most use cases for cool down.

Thanks for the submission sorry it is being rejected as a duplicate.

wolfmanjm:
look at PR #1052

michaelnew:
I don't think the temperature switch module handles either of the use cases I mentioned, unless I'm misunderstanding what it does. I don't want to switch something off or on after a target temperature is reached; I want it to wait until the bed or hotend is cool enough before proceeding to the next piece of gcode. For me that's a script to remove a part from the bed, but for others it could be switching materials or something else along those lines. I also need to be able to set that target temperature based on different materials or for different parts, and that's not something can do when I have to edit a config file on the SD card.

I'm looking through that other PR. It looks like it does the same thing with the addition of allowing for a temperature range, so that's pretty cool, but it's going to run into the same problem I had to work around here. The temperature will likely not hit the target when it's cooling down unless the heater is switched off. My printer kept getting stuck waiting indefinitely until I made that change.

I also don't think that quietly aborting the heat command if the range is narrow is a good idea at all:
```
if (r-v < 3) { // Invalid range if r < v
    return;
}
```
As a user it would be very hard to understand why a heat command was ignored, and the print would likely proceed at the wrong temperature, which is not something I ever want to happen.

PR #1246, opened 2017-12-24
a bit more convenient messages to return when empty or incorrect comm…

alpha6:
A bit more convenient messages to return when empty or incorrect commands are received.

The current behavior is slightly incorrect and confusing. PR fixes that.

- If empty line was received: do nothing. It's a trash in tty in 99.99% cases. 
- If no command in N: there is a100% error. Return an error message to tty.
- If no g-code or console command was received: show what is wrong in the input line and say "ok".

### DIFF
--- a/src/modules/communication/GcodeDispatch.cpp
+++ b/src/modules/communication/GcodeDispatch.cpp
@@ -60,9 +60,8 @@ void GcodeDispatch::on_console_line_received(void *line)
     int ln = 0;
     int cs = 0;
 
-    // just reply ok to empty lines
+    // empty line is just an incorrect input in most cases. Just skip it.
     if(possible_command.empty()) {
-        new_message.stream->printf("ok\r\n");
         return;
     }
 
@@ -114,8 +113,10 @@ try_again:
 			if(lnsize != string::npos) {
 				possible_command = possible_command.substr(lnsize);
 			}else{
-				// it is a blank line
+				// Blank line with the N prefix it's an error.
+                new_message.stream->printf("Error: Empty command with N!.\r\n");
 				possible_command.clear();
+                return;
 			}
 
         } else {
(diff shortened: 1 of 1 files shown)


src/modules/communication/GcodeDispatch.cpp
```diff
@@ -60,9 +60,8 @@ void GcodeDispatch::on_console_line_received(void *line)
     int ln = 0;
     int cs = 0;
 
-    // just reply ok to empty lines
+    // empty line is just an incorrect input in most cases. Just skip it.
     if(possible_command.empty()) {
-        new_message.stream->printf("ok\r\n");
```
wolfmanjm, line 65: Absolutely NOT, this is NOT the reprap spec. EVERY line MUST return OK.

src/modules/communication/GcodeDispatch.cpp
```diff
@@ -114,8 +113,10 @@ void GcodeDispatch::on_console_line_received(void *line)
 			if(lnsize != string::npos) {
 				possible_command = possible_command.substr(lnsize);
 			}else{
-				// it is a blank line
+				// Blank line with the N prefix it's an error.
+                new_message.stream->printf("Error: Empty command with N!.\r\n");
```
wolfmanjm, line 117: This also is NOT what reprap hosts expect.

wolfmanjm (commented):
This breaks compatibility and expected behavior with all the hosts out there. - sorry it is rejected.

wolfmanjm:
FWIW the responses are not supposed to be "convenient" they are supposed to conform to a protocol that most reprap hosts understand. This is mostly undocumented. blank lines however MUST return OK otherwise the ping pong protocol breaks down. Returning random messages also tends to confuse hosts.

alpha6:
Could you give me a link to the spec? I spent some time searching and did not found anything like it.

wolfmanjm:
the reprap wiki has some info but is incomplete, and/or wrong. You need to actually look at the major hosts and see what they expect. So look at pronterface and octoprint code base to see how they talk to the various boards.

wolfmanjm:
Also smoothie currently works very well with the supported hosts, so no changes to the protocol will be made.

Issue #1255, opened 2018-01-10
Bug in the progress command


obujor:
Hi,

I've noticed a bug in the _progress_ command when playing large gcode files (more than 2 millions of lines) from the sd on a Smoothieboard 4XC v1.1 controlling my laser diode engraver.

From 0% to 70% the progress was working great, after 70% instead of 71% I'm getting 1%, 2% and so on.
This is the output of _version_ command:

> Build version: edge-4f622ed, Build date: Dec 22 2017 21:38:05, MCU: LPC1769, System Clock: 120MHz  CNC Build 3 axis

I'm sending commands to the board from the _imrahil-ui_.

This's a list of progress outputs (Notice that _progress -b_ command is working properly):

> file: /sd/memi-15-4000.gcode, 0 % complete, elapsed time: 00:02:27, est time: 04:41:53
> file: /sd/memi-15-4000.gcode, 2 % complete, elapsed time: 00:04:42, est time: 03:29:09
> SD printing byte 1363816/58961451
> file: /sd/memi-15-4000.gcode, 3 % complete, elapsed time: 00:06:08, est time: 03:13:19
> SD printing byte 2287582/58961451
> file: /sd/memi-15-4000.gcode, 6 % complete, elapsed time: 00:10:55, est time: 02:49:21
> file: /sd/memi-15-4000.gcode, 6 % complete, elapsed time: 00:11:51, est time: 02:47:48
> file: /sd/memi-15-4000.gcode, 9 % complete, elapsed time: 00:16:08, est time: 02:37:58
> file: /sd/memi-15-4000.gcode, 12 % complete, elapsed time: 00:20:41, est time: 02:28:46
> SD printing byte 7313401/58961451
> file: /sd/memi-15-4000.gcode, 12 % complete, elapsed time: 00:21:51, est time: 02:26:21
> file: /sd/memi-15-4000.gcode, 13 % complete, elapsed time: 00:23:09, est time: 02:24:05
> file: /sd/memi-15-4000.gcode, 18 % complete, elapsed time: 00:30:06, est time: 02:13:35
> file: /sd/memi-15-4000.gcode, 21 % complete, elapsed time: 00:34:11, est time: 02:08:01
> file: /sd/memi-15-4000.gcode, 22 % complete, elapsed time: 00:36:00, est time: 02:05:51
> file: /sd/memi-15-4000.gcode, 24 % complete, elapsed time: 00:38:50, est time: 02:02:39
[...]

PR #1256, opened 2018-01-11
fix progress when large files are being played

wolfmanjm:
addresses Issue #1255

### DIFF
--- a/src/modules/utils/player/Player.cpp
+++ b/src/modules/utils/player/Player.cpp
@@ -339,10 +339,10 @@ void Player::progress_command( string parameters, StreamOutput *stream )
                 est = (file_size - played_cnt) / bytespersec;
         }
 
-        unsigned int pcnt = (file_size - (file_size - played_cnt)) * 100 / file_size;
+        float pcnt = (((float)file_size - (file_size - played_cnt)) * 100.0F) / file_size;
         // If -b or -B is passed, report in the format used by Marlin and the others.
         if (!sdprinting) {
-            stream->printf("file: %s, %u %% complete, elapsed time: %02lu:%02lu:%02lu", this->filename.c_str(), pcnt, this->elapsed_secs / 3600, (this->elapsed_secs % 3600) / 60, this->elapsed_secs % 60);
+            stream->printf("file: %s, %u %% complete, elapsed time: %02lu:%02lu:%02lu", this->filename.c_str(), (unsigned int)roundf(pcnt), this->elapsed_secs / 3600, (this->elapsed_secs % 3600) / 60, this->elapsed_secs % 60);
             if(est > 0) {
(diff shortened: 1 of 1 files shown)


obujor:
Hi,
in my opinion to prevent 32 integer overflow we can just rewrite the expression in another way.
This is the original line:
`unsigned int pcnt = (file_size - (file_size - played_cnt)) * 100 / file_size;`

The factor causing integer overflow is _100_ because _file_size_ will hardly be so big to overflow integer, that will be a file of about 4GB of size, almost 200 milions of line of code.

So I think instead of using _float_ we can write the expression that way (using less memory than float):
`unsigned int pcnt = (file_size - (file_size - played_cnt)) / file_size * 100;`

And thus the factor which may cause overflow could be only _file_size_ but for reasons listed above this is very difficult to happen even because I don't think smoothie supports huge files of about 4GB.

I hope this can be helpful.

triffid:
That's not how integer division works. It'll always return zero unless played_cnt gets larger than file_size.

Since this operation doesn't need much precision, why not do something like `unsigned pcnt = played_cnt / (file_size / 100)` as long as file_size is larger than a megabyte or so?

obujor:
Oh yes, you're right, it's been a while since I did not write C* code.
I like your solution.

wolfmanjm:
I already fixed it with a float (which is the same size as an int anyway).

obujor:
> which is the same size as an int anyway

I was not aware of this, sorry for my  useless comments.

karabas:
As I see on LCD Display percentage  does not work correctly. It somewhere overflows after about 40 hours of printing and starts from zero. I think LCD code was not changed.
I think we need something similar with percent_complete
```
void Player::on_get_public_data(void *argument)
{
    PublicDataRequest *pdr = static_cast<PublicDataRequest *>(argument);

    if(!pdr->starts_with(player_checksum)) return;

    if(pdr->second_element_is(is_playing_checksum) || pdr->second_element_is(is_suspended_checksum)) {
        static bool bool_data;
        bool_data = pdr->second_element_is(is_playing_checksum) ? this->playing_file : this->suspended;
        pdr->set_data_ptr(&bool_data);
        pdr->set_taken();

    } else if(pdr->second_element_is(get_progress_checksum)) {
        static struct pad_progress p;
        if(file_size > 0 && playing_file) {
            p.elapsed_secs = this->elapsed_secs;
  //          p.percent_complete = (this->file_size - (this->file_size - this->played_cnt)) * 100 / this->file_size;
			p.percent_complete = (((float)this->file_size - (this->file_size - this->played_cnt)) * 100.0F) / this->file_size;
            p.filename = this->filename;
            pdr->set_data_ptr(&p);
            pdr->set_taken();
        }
    }
}

```

PR #1279, opened 2018-03-05
Add Marlin compatible M109 Rxxx support

crpalmer:
Support set-and-wait of an exact temperature as per the Marlin implementation.

If both S and R are specified then the temperature specified by the S parameter
takes precedence.

When processing M109 R<T> the current temperature is used to determine if we
need to heat or cool.  If the current temperature is less than T then it is
equivalent to M109 S<T>.  If the current temperature is higher than T then
we want until the temperature drops to T.

----

Note: This differs from the existing open pull request (1052) significantly.  That request looks to be implementing a "set and wait for range" and/or "turn off heater and wait for temperature".  This approach is simpler and semantically matches what people might expect coming from other firmware.

### DIFF
--- a/src/modules/tools/temperaturecontrol/TemperatureControl.cpp
+++ b/src/modules/tools/temperaturecontrol/TemperatureControl.cpp
@@ -325,7 +325,8 @@ void TemperatureControl::on_gcode_received(void *argument)
                 }
             }
 
-        } else if( ( gcode->m == this->set_m_code || gcode->m == this->set_and_wait_m_code ) && gcode->has_letter('S')) {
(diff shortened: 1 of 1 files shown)


michaelnew:
Tested this out a bit today. The problem is that if the target temperature is significantly cool enough it will never be reached, because the heater is kicked on as the temp gets close. 
image

I believe the heater needs to be turned off until the target temperature is reached. I submitted a PR a couple years ago that did it this way: https://github.com/Smoothieware/Smoothieware/pull/1240/files

wolfmanjm:
No one has tested this in 2 years, so I presume there is no interest in it. Closing for now.

PR #1280, opened 2018-03-06
Allow Probe tangent to diagonal surface

Zaaphod:
Allow Probe in multiple axes to approch tangent to a diagonal surface.  There is no reason to not allow this, and if the user does not want to do a multi axis probe, they can simply supply only one axis as argument.  If the user supplies multiple axis, they expect a diagonal move, not the other axis to be ignored.

Diagonal probing is more accurate, especially at steep angles. it also allows a host program to probe the entire shape of a part accurately and adjust to new angles as it progresses around the part.


wolfmanjm:
This is wrong. Unused axis need to be set to nan. I'm closing this and I'll fix it myself.

Thanks

Issue #1291, opened 2018-03-26
G2/G3 do not always perform full circles


Zaaphod:
#### What were you doing?
Running the following Gcodes:
G0X7.86Y13.96M400 
M400
G1Z0
G2X7.86Y13.96I-0.11J-0.06     

#### What did you expect to happen?
A full circle to be performed

#### What happened instead?
no circle was performed.

#### Branch & Commit or Version of Smoothieware

edge-e9942d1, Build date: Jan 1 2018 22:24:20, 
CNC build

#### Hardware model you are using (Make, and version)

Smoothieboard 0x

#### Host and Operating System running the Host

Production Automation Gcode host

#### Link to and console logs

2018/03/25 07:13pm COM3: --->N15G0X7.86Y13.96M400*113
2018/03/25 07:13pm COM3: <---ok
2018/03/25 07:13pm COM3: --->N16M400*48
2018/03/25 07:13pm COM3: <---ok
2018/03/25 07:13pm COM3: --->N17G1Z0*84
2018/03/25 07:13pm COM3: <---ok
2018/03/25 07:13pm COM3: --->N18G2X7.86Y13.96I-0.11J-0.06*2
2018/03/25 07:13pm COM3: <---ok
    
#### Link to running config and M503 results

Config - https://pastebin.com/F4U55Yuu
M503 - https://pastebin.com/ktuYRg57

#### Additional information:
This is a known bug and has been happing a very long time.  The quick fix has been to break full circles into 2 arcs, however trying to interface with a new cam program is making breaking the circles into arcs time consuming so I'm now trying to solve this issue.

here are more examples:
G0X18.038253Y9.539614M400
G1Z0
G2X18.038253Y9.539614I-0.108253J-0.0625
Does not perform the circle

G0X18.0383Y9.54M400
G1Z0
G2X18.0383Y9.54I-0.1083J-0.062886
Does not perform the circle

G0X18.0383Y9.54M400
G1Z0
G2X18.0383Y9.54I-0.1083J-0.063
DOES perform the circle

[...]

PR #1293, opened 2018-03-30
Improvements to G2/G3 Arcs

Zaaphod:
This PR is to fix the issue documented #1291.  It was found that the atan2 function has limitations which can cause a full circle to be ignored.  The conversion after it could sometimes fix it, however the conversion also had the possibility of creating a full circle when a tiny arc segment was intended, and it also was unable to fix all occurrences.   
Both these issues are solved by first detecting if the command is a true full circle, and if so, simply setting the angular_travel  to the correct value.  By eliminating full circles from the atan2 formula, it now is very reliable, anything from the tiniest arc segment to the largest imaginable arc that is not a full circle calculate accurately.  
Since the atan2 formula is now reliable, the section after it that was used as a correction after it now simply needs to perform an adjustment to shift the output of atan2.  atan2 can produce a result from pi to -pi (excluding -pi) however for our purposes, we need counterclockwise arcs to be from 0 to 2pi and clockwise arcs to be from -2pi to 0.  The adjustment is able now achieve this without errors because the atan2 calculation works fine if there are no full circles (full circles have identical start and end points and everything cancels out).  With massively extreme input like a start and end point only 0.00001mm apart with a radius over 2000 for example, the atan2 formula will produce a result of 0, it is important that if it produces 0, it stays 0, otherwise one would get a gigantic full circle.  The current adjustment makes sure that a 0 result from the atan2 formula will remain 0 for these extreme edge cases.

### DIFF
--- a/src/modules/robot/Robot.cpp
+++ b/src/modules/robot/Robot.cpp
@@ -1461,11 +1461,8 @@ bool Robot::append_arc(Gcode * gcode, const float target[], const float offset[]
     float r_axis1 = -offset[this->plane_axis_1];
     float rt_axis0 = target[this->plane_axis_0] - center_axis0;
     float rt_axis1 = target[this->plane_axis_1] - center_axis1;
+    float angular_travel = 0;
 
-    // Patch from GRBL Firmware - Christoph Baumann 04072015
-    // CCW angle between position and target from circle center. Only one atan2() trig computation required.
-
-    float angular_travel = atan2f(r_axis0 * rt_axis1 - r_axis1 * rt_axis0, r_axis0 * rt_axis0 + r_axis1 * rt_axis1);
     gcode->stream->printf("Mpos Plane_Axis_0: %8.34f\r\n", machine_position[this->plane_axis_0]);
     gcode->stream->printf("Mpos Plane_Axis_1: %8.34f\r\n", machine_position[this->plane_axis_1]);
     gcode->stream->printf("Offset Plane_Axis_0: %8.34f\r\n", offset[this->plane_axis_0]);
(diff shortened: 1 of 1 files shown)


src/modules/robot/Robot.cpp
```diff
@@ -1461,17 +1461,30 @@ bool Robot::append_arc(Gcode * gcode, const float target[], const float offset[]
...
+    float angular_travel = 0;
 
-    // Patch from GRBL Firmware - Christoph Baumann 04072015
-    // CCW angle between position and target from circle center. Only one atan2() trig computation required.
-    float angular_travel = atan2f(r_axis0 * rt_axis1 - r_axis1 * rt_axis0, r_axis0 * rt_axis0 + r_axis1 * rt_axis1);
-    if (plane_axis_2 == Y_AXIS) { is_clockwise = !is_clockwise; }  //Math for XZ plane is reverse of other 2 planes
-    if (is_clockwise) { // Correct atan2 output per direction
-        if (angular_travel >= -ARC_ANGULAR_TRAVEL_EPSILON) { angular_travel -= (2 * PI); }
+    //Check to see if we have a full circle, and if so, set angualr_travel.   
+    if ((this->machine_position[this->plane_axis_0]==target[this->plane_axis_0]) and
```
wolfmanjm, line 1467: please replace and with && as and is not a c++ thing

wolfmanjm:
Thanks. I'll fix the and :)

wolfmanjm:
Ok this breaks the fix I did before. I now get a huge circle on a tiny arc.

wolfmanjm:
I reverted this merge as it breaks a previous fix for small arcs.

src/modules/robot/Robot.cpp
Zaaphod: Opps thanks for catching that... I'm surprised I did not get a compiler error.   I've been working on my Pascal project a lot lately, it uses 'And'

PR #1296, opened 2018-03-31
Use arc_milestone

Zaaphod:
This is another attempt to address the issues in #1291 

It uses arc_milestone to track the positions given in gcodes to calculate arcs, this prevents error due to tiny movements that do not cause motion.  It also incorporates the pervious changes made by PR #1293, as I have not found an alternative method as efficient as this code to handle the other issues... I keep ending up with less efficient code that ends up with the same limitations.

This method will makes a distinction between full circles because they are precisely defined that way in Gcode accurate to floating point precision, and false full circles that are really microscopic arc segments that are further than float precision away, but less than machine resolution away.  It is important that the gcode file is generated with enough precision to be clear whether it is a tiny arc segment or a full circle as in this example:

Exporting 4 decimal places gives us this:
G1 X406.1208 Y172.6107
G2 X406.1210 Y172.6111 I49.9647 J-17.0159
because the G1 point is not exactly the same as the G2 defined end point, this is a very small arc segment.

if we only export 3 decimal places we would have:
G1 X406.121 Y172.611
G2 X406.121 Y172.611 I49.965 J-17.016
this is a full circle by definition because both start and end are identical, there is no possible way to distinguish this.

There is still a possibility that points smaller than floating point precision could be considered a full circle, because we have floating point variables.  Users should be careful with simulators, as many of them use doubles.  If there is an option in the simulator for precision to use, single floating point accuracy should be used for 100% compatibility.   I do not expect an issue due to this, because floating point accuracy can distinguish between 1000 and 1000.0001 and if you are generating arc segments smaller than 0.0001mm then you need to learn to clean up your CAD drawing before generating Gcodes... even if smoothie could process files with that small of an arc segment, it would run very poorly as it would be throttled by the sheer massive amount of coordinates being shoved into it.

One additional change was made in the calculation of  rt_axis0 and rt_axis1.  It is a minor change that improves accuracy of these values and catches some extreme edge cases operating a near float resolution.  The formula has not changed, but it is simply done all in one operation now
`float rt_axis0 = target[this->plane_axis_0] - this->arc_milestone[this->plane_axis_0] - offset[this->plane_axis_0];`
this does the calculation and stores the result with float precision
the original code calculated the center first, stored it as a float, then used the center to find  rt_axis0.  this resulted in a possible compound float resolution issue and had the possibility of resulting in a float other than expected.  The current formula seems to always produce the expected result, because it is only stored as a float once.  This very small change was necessary to correctly process a few extreme edge cases, and will not affect anything else.

### DIFF
--- a/src/modules/robot/Robot.cpp
+++ b/src/modules/robot/Robot.cpp
@@ -1457,17 +1457,19 @@ bool Robot::append_arc(Gcode * gcode, const float target[], const float offset[]
     }
 
     // Scary math
-    float center_axis0 = this->machine_position[this->plane_axis_0] + offset[this->plane_axis_0];
-    float center_axis1 = this->machine_position[this->plane_axis_1] + offset[this->plane_axis_1];
-    float linear_travel = target[this->plane_axis_2] - this->machine_position[this->plane_axis_2];
+    float center_axis0 = this->arc_milestone[this->plane_axis_0] + offset[this->plane_axis_0];
+    float center_axis1 = this->arc_milestone[this->plane_axis_1] + offset[this->plane_axis_1];
+    float linear_travel = target[this->plane_axis_2] - this->arc_milestone[this->plane_axis_2];
     float r_axis0 = -offset[this->plane_axis_0]; // Radius vector from center to current location
     float r_axis1 = -offset[this->plane_axis_1];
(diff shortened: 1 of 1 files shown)


wolfmanjm:
Thanks. seems to work on all my tests too.

PR #1297, opened 2018-03-31
Fix typos in comments of robot.cpp

Zaaphod:


wolfmanjm:
Sorry I consider these gratuitous changes. But would have accepted them if they did not collide with some changes I have in progress

PR #1298, opened 2018-04-02
Correct Arc Start position to current machine position

Zaaphod:
I've encountered an off the wall case where an arc did not perform as expected.  The correction is simple.  Here's the scenario that caused the issue.

Trying to re-mill a hole a little larger... the part was removed and then later put back in a vice... so I used a probe to find the center,  of the hole, then to find the right edge of the hole, then I did a G92 to the start coordinate of the hole so I could re-run the program from the middle.  the G92 correctly re-set the coordinate, to what should have been the start coordinate, but it was not used because it didn't move the machine.    I could have did the G92 to the center of the hole then a G0 to the edge of the hole and would not have had an issue... but I thought maybe someone else would / could do something similar and end up with inexplicable results, so I thought I better fix it.

This simple fix takes care of this issue without impacting anything else.

### DIFF
--- a/src/modules/robot/Robot.cpp
+++ b/src/modules/robot/Robot.cpp
@@ -1138,6 +1138,8 @@ void Robot::reset_position_from_current_actuator_position()
         actuators[i]->change_last_milestone(actuator_pos[i]); // this updates the last_milestone in the actuator
     }
     #endif
+    // needed to act as start of next arc command
+    memcpy(arc_milestone, machine_position, sizeof(arc_milestone));
 }
 
 // Convert target (in machine coordinates) to machine_position, then convert to actuator position and append this to the planner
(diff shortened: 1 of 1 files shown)


Zaaphod:
Just noticed arc_milestone was not initialized, had strange values in it after a reset..  Added the initialization to this PR

wolfmanjm:
G92 does not call the reset_position_from_current_actuator_position() so I do not see how this could have any effect. And this is what I was worried about having to sync this everywhere. This is only called after a home, and as you pointed out it is wrong to have an arc done without a G0/G1. So I am inclined to reject this.

Zaaphod:
I've looked at it... it was the probe that called reset_position_from_current_acuator_position() and that's why it worked... I can go either way on this... it IS technically wrong to do the arc without a G0/G1 first, and that's what I should have done.   I do think arc_milestone should be initialized though.

wolfmanjm:
G92 MUST be followed by a move of G0 or G1 otherwise the machine position will be wrong.
I'll add initialization though although based on your statement that G0 or G1 MUST precede a G2/G3 it should not be necessary.

A G2/G3 should never follow a probe without the G0/G1 so I think this is just operator error IMO.

Zaaphod:
Agreed: Operator error

wolfmanjm:
Turns out this was needed. However arc_milestone is not needed anymore so I have removed it entirely.

PR #1304, opened 2018-04-11
Fix for: "must define all preceding A, B, C axis as homing endstops"

markmaker:
This is the fix for the bug described here: http://smoothieware.org/6axis 

> NOTE Due to a bug you must define all preceding Axis as homing endstops, so if you want a C endstop to home then XYZAB need to also be defined not just XYZC. This will be fixed eventually.

This fix removes the requirement. Endstops on A, B, C can be defined with gaps.

I tested this on my 5 axis machine with A having no endstops defined (rotation axis) and B having endstops and limits enabled. Tested home_to_min, home_to_max. Tested HALT when max_travel exhausted. Also works when the A endstops are defined but the pins are set to nc. Tested with homing_order ZXYB. Tested homing individual axis using i.e. G28 B0. 

Thanks for a great open source firmware and board!

### DIFF
--- a/src/modules/tools/endstops/Endstops.cpp
+++ b/src/modules/tools/endstops/Endstops.cpp
@@ -632,14 +632,15 @@ void Endstops::home(axis_bitmap_t a)
 
     // potentially home A B and C individually
     if(homing_axis.size() > 3){
-        for (size_t i = A_AXIS; i < homing_axis.size(); ++i) {
-            if(axis_to_home[i]) {
+        for (size_t ia = A_AXIS; ia <= homing_axis.size(); ++ia) {
+            size_t i = homing_axis[ia].axis_index;
+            if (axis_to_home[i]) {
                 // now home A B or C
                 float delta[i+1];
                 for (size_t j = 0; j <= i; ++j) delta[j]= 0;
-                delta[i]= homing_axis[i].max_travel; // we go the max
-                if(homing_axis[i].home_direction) delta[i]= -delta[i];
-                THEROBOT->delta_move(delta, homing_axis[i].fast_rate, i+1);
+                delta[i]= homing_axis[ia].max_travel; // we go the max
+                if(homing_axis[ia].home_direction) delta[i]= -delta[i];
+                THEROBOT->delta_move(delta, homing_axis[ia].fast_rate, i+1);
                 // wait for it
                 THECONVEYOR->wait_for_idle();
             }
@@ -661,8 +662,9 @@ void Endstops::home(axis_bitmap_t a)
 
     // also check ABC
     if(homing_axis.size() > 3){
-        for (size_t i = A_AXIS; i < homing_axis.size(); ++i) {
-            if(axis_to_home[i] && !homing_axis[i].pin_info->triggered) {
+        for (size_t ia = A_AXIS; ia < homing_axis.size(); ++ia) {
+            size_t i = homing_axis[ia].axis_index;
+            if (axis_to_home[i] && !homing_axis[ia].pin_info->triggered) {
                 this->status = NOT_HOMING;
                 THEKERNEL->call_event(ON_HALT, nullptr);
                 return;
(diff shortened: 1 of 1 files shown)


src/modules/tools/endstops/Endstops.cpp
```diff
@@ -679,8 +681,8 @@ void Endstops::home(axis_bitmap_t a)
 
     // Move back a small distance for all homing axis
     this->status = MOVING_BACK;
-    float delta[homing_axis.size()];
-    for (size_t i = 0; i < homing_axis.size(); ++i) delta[i]= 0;
+    float delta[C_AXIS+1];
```
wolfmanjm, line 684: This is not an acceptable workaround. Hard coding the size of this array is incorrect. there is a reason it was set to the size of the homing_axis size. This may have unintended consequences.

wolfmanjm:
I do not like this "fix" at all I am afraid. But I do thank you for attempting to fix the issue, but it simply is not that simple otherwise it would have been fixed earlier.
What is not being considered here is other configurations that do not use the new syntax and backwards compatibility.

I will take another look at trying to fix this myself, but as it is a extreme edge case and is easily worked around by simply defining the intermediate endstops it may take a while.
I would recommend you keep this in your own fork for now as it seems to work for you.

src/modules/tools/endstops/Endstops.cpp
markmaker: Thank you for looking into this. Please reconsider your assessment. 

The selection of axes is governed by the axis_bitmap_t. I'm afraid this type is hard coded to 6 axis in Endstops.h. The only constant I could find is the C_AXIS macro from nuts_bolts.h (Grbl). Many things in Endstops.cpp are hardcoded to support only XYZABC. So I adapted to that. 

All axis movements (the delta vector) are governed by the fixed axis index values. So having a vector for all the axes there are is not a mistake. 

On the other hand the homing_axis collection list is just a list. The list position inside homing_axis is not to be mistaken as axis index. That's where the bug lies. 

To make things a bit more complicated, there is code at Endstops.cpp:336 to make sure X, Y, Z are always populated, albeit with nc pins to mark those as dummies. For X, Y, Z list positions in deed do correspond to axis index. This was then mistakenly also assumed for A, B, C.

However for the A, B, C axis the homing_axis[list_pos].axis_index is the correct mapping from the list position to the axis index. 

I really, really analyzed this. Please reconsider.

wolfmanjm:
can you test PR #1305 I think this will fix the issue correctly. Thank you

markmaker:
Again, please reconsider. I humbly disagree about backwards compatibility for reasons I described above. 

Please look at the finishing steps of the homing process **in the original code**. The "Move back a small distance for all homing axis" and "Start moving the axes towards the endstops slowly". There you clearly see the same homing_axis[list_pos].axis_index mapping taking place **in the original code**. There **too** the delta vector may contain zeroes for axes not homed at the moment (i.e. G28 with individual axes specified or singled out when having a homing order). 

I sized and zeroed the delta vector for all fixed axis indices. This is backwards compatible because there will be no homing_axis[c].axis_index entries pointing to the delta vector A, B, C components. So these vector entries will remain at zero. The robot will do the right thing (not move at zero) as it already did in the original code when one missed out X, Y, Z axes.

wolfmanjm:
I have fixed this correctly. 
I am afraid the code is more complex than you have grasped, and you have not considered cases where only XYZ axis are defined in other places like Robot. There are many interrelated bits of code spread out over smoothie, and in this case you simply have not covered all the edge cases.
My decision is final.

wolfmanjm:
Here is an example of where this would cause most other peoples firmware to crash...

You call `THEROBOT->delta_move(delta, feed_rate, C_AXIS+1);` this will cause delta move to crash in the case where only three axis are defined (the CNC build). And I think there are several other cases where this is true.
It is not unexpected for someone to not fully understand the intricacies of the code and interrelations, so please do not feel bad about it.

markmaker:
No problem, I see the pitfalls now. 

EDIT: see that you truncate by max_index.

BTW: how does one define the order of the module initialization i.e. how can you be sure, that THEROBOT->get_number_registered_motors() is already at its final value when the config is still being loaded? Is there an inherent "motors before endstops" order? Am I allowed to order and group in the config.txt freely?

wolfmanjm:
there is a reason for dynamically allocating the array, setting it to the maximum possible is a total waste of RAM which is in very short supply.
A quick look at main.cpp will show you why I can rely on the number of motors being setup before endstops needs it.

markmaker:
Thanks. 

Further comment at #1305.

PR #1305, opened 2018-04-11
Allow non contiguous definition of ABC homing endstops

wolfmanjm:


markmaker:
Just tested this on my machine. Works!

Many thanks, your very fast response - helps a lot with my project!

-Mark

P.S. Re RAM: I read a lot negative stuff about the use of std::vectors (and such) in MCUs as these containers grow dynamically one element by one which involves a lot of copying/moving and fragmentation (no MMU to help). Internally a geometric capacity buffer growth takes place (space for 1, 2, 4, 8 elements is allocated with gcc). For X, Y, Z you get 1 + 2 + 4 = 7 elements allocated, 3 freed in two fragments, 3 additional memcopies. 

So I did the test with the best case (just X, Y, and Z endstops configured):

std:vector (code as is):

> @mem
SENDING:mem
Unused Heap: 3180 bytes
Used Heap Size: 23164
Allocated: 15584, Free: 6172
Total Free RAM: 9352 bytes
Free AHB0: 11080, AHB1: 10440
Block size: 84 bytes, Tickinfo size: 280 bytes

Then changed to a std::array< ... 6>

> @mem
SENDING:mem
Unused Heap: 3316 bytes
Used Heap Size: 23028
Allocated: 15456, Free: 6172
Total Free RAM: 9488 bytes
Free AHB0: 10900, AHB1: 10440
Block size: 84 bytes, Tickinfo size: 280 bytes

Yep. 136 bytes saved although A, B, C are always allocated as dummies in the dumb array! 

I tried adding:
`homing_axis.reserve(max_index+1);`
at the original version of [Endstops.cpp:341](https://github.com/Smoothieware/Smoothieware/blob/31b817d717db963c207c80b54dc6818702fe1e99/src/modules/tools/endstops/Endstops.cpp#L341)

See also: [std::vector::reserve()](http://en.cppreference.com/w/cpp/container/vector/reserve)

But it does not help. 

>Unused Heap: 3160 bytes
Used Heap Size: 23184
Allocated: 15552, Free: 6208
Total Free RAM: 9368 bytes
Free AHB0: 11080, AHB1: 10440
Block size: 84 bytes, Tickinfo size: 280 bytes

Seems those guys were right.

wolfmanjm:
you would need to use void shrink_to_fit() not reserve! and I think your results may be bogus as AHB storage should not change but does, so I suspect there are some things going on affecting storage in between your tests.
I am well aware of the pros/cons of using vector vs array. and reallocations and lack of MMU in this context are irrelevant as the array is created once and not changed.. Each storage unit of a homing array is 32 bytes, adding 3 useless ones is still a waste, and there are other ways to shrink the vector size if it had over allocated, which I will look into if shrink_to_fit() does not work.

PR #1316, opened 2018-04-20
Update sample configs for proper debounce entry

cilynx:
While the Endstop module supports both `debounce_count` and
`debounce_ms`, the ZProbe module supports only `debounce_count`.

https://github.com/Smoothieware/Smoothieware/blob/edge/src/modules/tools/endstops/Endstops.cpp

https://github.com/Smoothieware/Smoothieware/blob/edge/src/modules/tools/zprobe/ZProbe.cpp

This commit updates all of the sample config files that included
`zprobe.debounce_count` to use `zprobe.debounce_ms` instead.


cilynx:
...er.  ZProbe module supports only `debounce_ms`....

wolfmanjm:
Thank you.

PR #1317, opened 2018-04-23
Add G38.4 & G38.5

Zaaphod:
This is to implement G38.4 and G38.5 as defined at [http://www.linuxcnc.org/docs/2.5/html/gcode/gcode.html#sec:G38-probe]

G38.4 and G38.5 are very useful for several probing operations.  

Many autoprobe schemes require that you probe again in a new direction from the probe recorded position, however you can't probe in any direction while the probe is engaged.  G38.4 & G38.5 solve this issue by backing away just enough to not be touching anymore, therefore a probe in the new direction can happen with a minimum of movement away from the part.  This is critical when probing shapes with tight inside angles where any excess backoff move could hit the opposite side of the angle

Also if you are probing at a fairly high speed, you will always over travel slightly due to debounce settings.  G38.4/G38.5 allows you do read the probe as you back away from the part very slowly until loss of probe contact, this is significantly faster than backing away a fixed distance and re-probing slowly and double probing every point, and then still needing to back away again so you can probe again.

Instead of:
`G91
G38.3X1F10
G0X-0.05
G38.3X0.1F1
G0X-0.05`

One could run:
`G38.3X1F10
G38.5X-0.1F1`
and record the points at which the probe looses contact with the part.   Notice now the probe is released so we can probe again without doing any relative movements.

These are just a few reasons why there is such a thing as G38.4 and G38.5, and I think smoothie would benefit by their inclusion.

### DIFF
--- a/src/modules/tools/zprobe/ZProbe.cpp
+++ b/src/modules/tools/zprobe/ZProbe.cpp
@@ -325,8 +325,8 @@ void ZProbe::on_gcode_received(void *argument)
 
     } else if(gcode->has_g && gcode->g == 38 ) { // G38.2 Straight Probe with error, G38.3 straight probe without error
         // linuxcnc/grbl style probe http://www.linuxcnc.org/docs/2.5/html/gcode/gcode.html#sec:G38-probe
-        if(gcode->subcode != 2 && gcode->subcode != 3) {
-            gcode->stream->printf("error:Only G38.2 and G38.3 are supported\n");
+        if(gcode->subcode < 2 || gcode->subcode > 5) {
+            gcode->stream->printf("error:Only G38.2, G38.3, G38.4, and G38.5 are supported\n");
             return;
         }
 
@@ -336,9 +336,16 @@ void ZProbe::on_gcode_received(void *argument)
             return;
         }
 
-        if(this->pin.get()) {
-            gcode->stream->printf("error:ZProbe triggered before move, aborting command.\n");
-            return;
+        if(gcode->subcode == 4 || gcode->subcode == 5) {
+            if(!this->pin.get()) {
+                gcode->stream->printf("error:ZProbe triggered before move, aborting command.\n");
+                return;
+            }
+        } else {
+            if(this->pin.get()) {
+                gcode->stream->printf("error:ZProbe triggered before move, aborting command.\n");
+                return;
+            }
         }
 
         // first wait for all moves to finish
@@ -362,7 +369,17 @@ void ZProbe::on_gcode_received(void *argument)
             return;
         }
 
+        if(gcode->subcode == 4 || gcode->subcode == 5) {
+            gcode->stream->printf("Inverting pin\n");
+            pin.set_inverting(pin.is_inverting() != 1);
+        }
+
         probe_XYZ(gcode, x, y, z);
+
+        if(gcode->subcode == 4 || gcode->subcode == 5) {
+            gcode->stream->printf("Inverting pin back\n");
+            pin.set_inverting(pin.is_inverting() != 1);
+        }
 
         return;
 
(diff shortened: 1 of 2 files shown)


src/modules/tools/zprobe/ZProbe.cpp
```diff
@@ -362,8 +369,16 @@ void ZProbe::on_gcode_received(void *argument)
             return;
         }
 
+        if(gcode->subcode == 4 || gcode->subcode == 5) {
+            pin.set_inverting(pin.is_inverting() != 1);
```
wolfmanjm: This is an ugly hack, sorry but not acceptable

wolfmanjm (changes requested):

src/modules/tools/zprobe/ZProbe.cpp
```diff
@@ -336,9 +336,16 @@ void ZProbe::on_gcode_received(void *argument)
             return;
         }
 
-        if(this->pin.get()) {
-            gcode->stream->printf("error:ZProbe triggered before move, aborting command.\n");
-            return;
+        if(gcode->subcode == 4 || gcode->subcode == 5) {
```
wolfmanjm: This code is way too verbose and can be significantly simplified.

src/modules/tools/zprobe/ZProbe.cpp
wolfmanjm: We just need a bit to say invert the test and xor that in the read_probe function.

src/modules/tools/zprobe/ZProbe.cpp
wolfmanjm: eg 
```
if(((gcode->subcode == 4 || gcode->subcode == 5) && !this->pin.get()) || ((gcode->subcode == 2 || gcode->subcode == 3) && this->pin.get())) {
  gcode->stream->printf("error:ZProbe triggered before move, aborting command.\n");
}

or even better....

if(this->pin.get() ^ (gcode->subcode >= 4)) {
￼             gcode->stream->printf("error:ZProbe triggered before move, aborting command.\n");
￼             return;
}
```

Zaaphod: I like the 'even better' code.  very smart and takes advantage of gcode.subcode being already constrained to be 2-5.

Zaaphod (commented):
Great!

src/modules/tools/zprobe/ZProbe.cpp
```diff
@@ -362,8 +363,17 @@ void ZProbe::on_gcode_received(void *argument)
             return;
         }
 
+        if(gcode->subcode == 4 || gcode->subcode == 5) {
+            invert_probe = true;
+        } else {
+            invert_probe = false;   // not sure this is needed since it is always reset to 0 below
```
wolfmanjm: not needed

src/modules/tools/zprobe/ZProbe.cpp
```diff
@@ -362,8 +363,17 @@ void ZProbe::on_gcode_received(void *argument)
...
+        }
         probe_XYZ(gcode, x, y, z);
 
+        if(gcode->subcode == 4 || gcode->subcode == 5) {
```
wolfmanjm: this should ALWAYS be set to false so remove the if clause.

wolfmanjm:
invert probe needs to be initialized to false, at the moment a regular G30 or other probe type will probably fail due to the fact invert_probe is not set one way or the other. It needs to be ALWAYS false unless it is a g38.4/5

wolfmanjm:
It is extremely bad practice to not initialize a variable like this. It should be initialized to false in the constructor. It is not sufficient to just do it where you have. It is prone to errors in case someone were to add another case statement or zprobe were to be called from a different module. This is simply not good coding practice.

Zaaphod:
Understood, I will add proper initialization in the constructor.

I have invert_probe= false; on line 259 for all Gcodes from 29 - 32 and again on line 329 for all G38.x Gcodes... I didn't want to put it before
`if( gcode->has_g && gcode->g >= 29 && gcode->g <= 32)`
because then I think it would run for every single Gcode.

my thinking is that this would guarantee invert_probe was false before any command that would use it if there was any condition that could possibly cause the invert_probe = false after probe_xyz to not be run?   
If so then it would be good to also re-set it before every G29-32.  G38.x could either always set it to false or set it conditionally.   If the invert_probe = false after probe_xyz is guaranteed to always run, then we don't need to reset it for every G29-G32 or set it to anything but true when needed for G38.4 and G38.5.

wolfmanjm:
Yes it should also be set to false for safety before any command that requires it to be false, but it also needs to be initialized in the ctor otherwise its state is indeterminate. when the zprobe class is constructed.

Zaaphod:
I hope this is the right place to initialize it,  it looks like the only place the "probing" variable is initialized.

wolfmanjm:
technically it is not. the ctor in the header is. but this will do

wolfmanjm (approved):

Zaaphod:
Do I have it initialized correctly now?  It compiles and I've been having great results in my testing, but I would like to make sure I have it technically correct.

PR #1320, opened 2018-05-01
Improve axis definitions to allow dual-motor axes

MZachmann:
Add an optional axis parameter to each motor to define which axis the motor is on

Also add a MappableCartesianSolution that supports dual-motor axes

Over time this should allow creating an axis<->motor map for supporting extruders and non-cartesian axes and dual motor axes in a natural fashion. This also enables a cleaner axis <-> actuator translation than using cartesian something for X,Y,Z and special casing A,B,C.

### DIFF
--- a/src/libs/StepperMotor.h
+++ b/src/libs/StepperMotor.h
@@ -10,6 +10,8 @@
 #include "Module.h"
 #include "Pin.h"
 
+#define AXIS_EXTRUDER 255
+
 class StepperMotor  : public Module {
     public:
         StepperMotor(Pin& step, Pin& dir, Pin& en);
@@ -53,6 +55,8 @@ class StepperMotor  : public Module {
         void set_selected(bool b) { selected= b; }
         bool is_extruder() const { return extruder; }
         void set_extruder(bool b) { extruder= b; }
+        uint8_t get_axis(void) { return axis; }
+        void set_axis(uint8_t a) { axis = a; }
 
         int32_t steps_to_target(float);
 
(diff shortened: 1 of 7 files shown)


MZachmann:
Hmmm, compiles fine for me. I guess your compiler doesn't have c++11. I'll put older code in.

wolfmanjm:
This was not compiled with the recommended tool chain, and fails to build.

Also This is really not required and IMO offers no significant advantages.

However if you wish please add a link to the third part branches in the wiki in case anyone could use this.

Thank you.

wolfmanjm:
Additionally this adds a level of complexity which is very nasty without really adding any extra functionality that cannot already be achieved.

wolfmanjm:
You did not follow the guidelines. Any code MUST be compiled with the provided toolchain. However as this will not be accepted into mainline code feel free to use whatever tool chain you prefer.

We do use C++11.

MZachmann:
Umm, ok but the error indicates missing std::to_string and it absolutely compiles for me with the old toolchain you use in Ubuntu.

Also, there is literally no way to support dual-motor axes today and this change not only adds that support but it cleans up the axis mapping stuff. w/e

wolfmanjm:
Dual motor axis are indeed supported in a much simpler way. Please check the wiki. Also This looks like it breaks the WCS mapping and various other things we use in CNC.

wolfmanjm:
You currently can also map any driver to any motor by simply changing the pin definitions in the config, no need for this complex, convoluted mapping scheme.

MZachmann:
I understand. I really just wanted a dual-motor axis but comparisons assumed that every motor was an axis or an extruder and that changing XYZ had no impact on other actuators so... since it ended up not adding code really I added the primary axis remapping mainly so I could test that dual was working. w/e I've got it working here. Thanks.

PR #1327, opened 2018-05-13
Fix for switch PWM value of 0

cdedwards:
Fix so that a 0 value turns off the PWM pin instead of having it pulse.
Currently having a S value of 0 will have the pin pulsing at a low rate. 0 should turn off the pin.

M255 S0   ; turn off case light.
M255 S40 ; turn on case light at a 40 rate.


wolfmanjm:
This is superfluous as this is already what happens...

https://github.com/Smoothieware/Smoothieware/blob/edge/src/libs/Pwm.cpp#L45

PR #1329, opened 2018-05-17
Add TMC2208 stepper motor driver support

tiagojbalmeida:
This pull request features an implementation of the TMC2208 stepper motor drivers. It is based on the TMC26X module and TMC21X module (from PR #1315). Different from previous modules, TMC2208 communicates with microcontroller via UART. 

Libraries containing software uart were moved from Spindle module to libs folder and used for TMC2208 driver. As soft UART stalls the whole firmware and cannot be used on a regular basis, it is just used during driver SETUP.

SpreadCycle and StealthChop modes are introduced as well as the velocity threshold to combine between these two modes.

Located on Snippets folder, a new tmc2208_motor driver.config reveals an example for this driver. All these driver parameters are set to default:

- Current: 1000 mA
- Sense Resistor: 50 mOhm
- Microsteps: 16
- UART TX pin: 0.27   (I used this pin for TX but you can use any other pin as it's used soft UART)
- UART RX pin: 0.28  (I used this pin for RX but you can use any other pin as it's used soft UART)
- UART Baudrate: 9600
- Mode: 0     (SpreadCycle mode)
- Threshold: 100   (Combines SpreadCycle and StealthChop mode. Use this only in StealthChop mode because it does NOT work in SpreadCycle)


arthurwolf:
Looks very cool, thanks for the contribution, we'll see if Jim has comments on code quality/structure.

src/modules/utils/motordrivercontrol/MotorDriverControl.h
```diff
@@ -38,20 +40,26 @@ class MotorDriverControl : public Module {
 
         void enable(bool on);
         int sendSPI(uint8_t *b, int cnt, uint8_t *r);
-
+        int sendUART(uint8_t *b, int cnt, uint8_t *r);
+        
         Pin spi_cs_pin;
         mbed::SPI *spi;
+        Pin sw_uart_tx_pin;
```
wolfmanjm: this is a waste of valuable RAM. Why do we need to have BOTH SPI and UART RAM used when it can only use one or the other? We really do not have the luxury of wasting RAM like this.

wolfmanjm (changes requested):

wolfmanjm:
Why was modbus moved to libs? when it is not used by this module. We are rapidly running out of FLASH space which is why softserial was originally only compiled when spindle was needed (and spindle is not compiled except for CNC), now we have to always compile softserial we are getting closer to running out of FLASH, so moving modbus back to where it was may be a good idea so we save some flash.
We could also make these recent changes optional too as they are rarely used at the moment. In fact making the entire advanced driver stuff optional to compile and not included by default may be worth considering if we add any more stuff as flash is very tight. TBH I don't think ANYONE is currently using the advanced driver stuff except a few Azteeg GT users using the bigfoot drivers. (probably no more than 1 or 2 users).

src/modules/utils/motordrivercontrol/MotorDriverControl.h
tiagojbalmeida: We can actually have both if we use a UART driver and also a SPI driver in the same printer for any reason. It would be better if we have defined any UART pins, so I could assign uart channels to UART pins as it's done for SPI drivers, however we are using soft uart and any board pins can be configured. Perhaps this should be configured inside the specific driver module instead of main MotorDriver module?

tiagojbalmeida:
Alright, i understand now why you moved the libs to spindle module. I will move both back to Spindle, and maybe just copy SoftSerial folder to motordriver folder instead of libs folder as it is only used for spindle and motor driver and this way would save flash too?

tiagojbalmeida:
I agree that this module should be included optionally, but i don't know yet how i would do that. I thought disabling driver in the config file would do the work

wolfmanjm:
No please leave softserial where it is in libs, as it is used by two modules now.
just move the modbus back.

> We can actually have both if we use a UART driver and also a SPI driver in the same printer for any reason.

No, each instance of a motor module has both serial and spi but can only use one or the other, that is the point I was trying to make. A single instance cannot use both.

I'll make the motor module optional, I just wanted feedback as if we need to do that yet.

wolfmanjm:
Disabling the driver saves RAM but not FLASH. Duplicating the code in motordriver and spindle would double the flash used for it if both are enabled.

wolfmanjm:
To exclude the motor driver you would do `EXCLUDE_MODULES = utils/motordrivercontrol`
As `tools/spindle` is in that by default for a 3d printer build.

Look in the src/makefile

wolfmanjm:
To save RAM you would put the soft uart pins in a union (and make them pointers, also make the cs_pin a pointer). Like I do with the driver pointer. That way they use the same memory but can only use one or the other.

tiagojbalmeida:
Ok, i will try to fix that. 

After doing this, just to make it clear, if we want to include motordriver module, the point is to manually edit the makefile to include that?

wolfmanjm:
No.
The makefile has the defaults, there is a user specific file that specifies what modules to build or not build. `default_excludes.mk`
Or it can be specified on the make command line, but for now it will be built by default, we may exclude it by default if we start to run out of FLASH.

You will never need to touch the makefile.

tiagojbalmeida:
Alright, thank you.

I will then proceed with the changes

tiagojbalmeida:
Sorry Jim, i got confused with your messages and i thought i might exclude motordriver module now and added this last commit. I can remove this commit and later you deal with the exclusion if needed, is that right?

wolfmanjm:
Right. please not not change the makefile.

src/modules/utils/motordrivercontrol/MotorDriverControl.cpp
```diff
@@ -164,20 +204,29 @@ bool MotorDriverControl::config_module(uint16_t cs)
...
                 case TMC2660: tmc26x->setRawRegister(&StreamOutput::NullStream, 255, 0); break;
+                case TMC2208: tmc22x->setRawRegister(&StreamOutput::NullStream, 255, 0); break;
             }
         }
 
     }else{
         rawreg= false;
     }
 
+    if(chip==TMC2208) {
```
wolfmanjm: no this is incorrect

tiagojbalmeida:
Exclude motordriver commit removed

src/modules/utils/motordrivercontrol/MotorDriverControl.cpp
tiagojbalmeida: yes, i need a better solution to make sure UART is not used after configuration. I was thinking about deleting the instance after configuration, but then i found this approach..

tiagojbalmeida: Last commit fixes this and previous one fixes SPI/UART pins issue

tiagojbalmeida: EDIT: I inserted a new commit moving also SPI and UART objects inside union, can you validate that please

src/modules/utils/motordrivercontrol/MotorDriverControl.h
```diff
@@ -42,16 +42,17 @@ class MotorDriverControl : public Module {
...
         union {
-            Pin *spi_cs_pin;
+            struct {
+                Pin *spi_cs_pin;
+                mbed::SPI *spi;
+            };
             struct {
                 Pin *sw_uart_tx_pin;
                 Pin *sw_uart_rx_pin;
                 bool setup;
```
wolfmanjm: This is incorrect, the unions need to be the same size to save any memory, so uart pointer and the bool should NOT be in the union.

tiagojbalmeida: Fine, but UART currently uses two pins and SPI just one, that means the size will be always equivalent as the size of two pins, is that it?

src/modules/utils/motordrivercontrol/MotorDriverControl.cpp
```diff
@@ -189,7 +235,15 @@ bool MotorDriverControl::config_module(uint16_t cs)
         this->register_for_event(ON_SECOND_TICK);
     }
 
-    THEKERNEL->streams->printf("MotorDriverControl INFO: configured motor %c (%d): as %s, cs: %04X\n", axis, id, chip==TMC2660?"TMC2660":chip==DRV8711?"DRV8711":"UNKNOWN", (spi_cs_pin.port_number<<8)|spi_cs_pin.pin);
+    //finish driver setup
+    if(chip == TMC2208) {
+        THEKERNEL->streams->printf("MotorDriverControl INFO: configured motor %c (%d): as %s, tx: %04X, rx: %04X\n", axis, id, "TMC2208", (sw_uart_tx_pin->port_number<<8)|sw_uart_tx_pin->pin, (sw_uart_rx_pin->port_number<<8)|sw_uart_rx_pin->pin);
+        //we don't want to use soft UART on a regular basis, so we delete unused UART object
+        setup = false;
+        delete this->serial;
```
wolfmanjm: You cannot do this as the serial will be needed for the periodic alarm checks etc. so it cannot be deleted.

tiagojbalmeida: ok, however the bool is maintained to not send/receive any byte via UART, which is our point since the beginning

src/modules/utils/motordrivercontrol/MotorDriverControl.h
```diff
@@ -38,20 +40,34 @@ class MotorDriverControl : public Module {
...
+        // use SPI or UART
+        union {
+            struct {
+                Pin *spi_cs_pin;
+                mbed::SPI *spi;
+            };
+            struct {
+                Pin *sw_uart_tx_pin;
+                Pin *sw_uart_rx_pin;
+                bool setup;
```
wolfmanjm: remove the setup and serial from this union and the sizes will be the same and use the same amount of memory as before.

tiagojbalmeida: ok, so i will keep uart pins inside a struct and also spi pin

src/modules/utils/motordrivercontrol/MotorDriverControl.cpp
wolfmanjm: No You cannot stop serial being used during operation, if you do then this whole thing is useless and the pins should be used to configure this chip instead of the serial, and this PR is not needed.

tiagojbalmeida: so i will keep uart being fully operating the whole time like spi or with a few restrictions? In fact, I just added full support to this driver as it's in other spi drivers but if serial is not used after setup, most of the support would be useless

src/modules/utils/motordrivercontrol/MotorDriverControl.h
tiagojbalmeida: i think union structure is correct now

src/modules/utils/motordrivercontrol/MotorDriverControl.h
```diff
@@ -60,6 +76,7 @@ class MotorDriverControl : public Module {
         uint32_t microsteps;
 
         char axis;
+        bool setup;
```
wolfmanjm: remove this it is not needed and a waste of 4 bytes per instance

tiagojbalmeida: ok, at the moment uart does not have restrictions using this driver

GriffinPaquette:
Is this code all good to go? I want to get testing with it. Awesome work Tiago and Jim on getting this sorted it seems!

tiagojbalmeida:
Yes @griffin-117, you may test it and provide some feedback. Thank you very much ;)

GriffinPaquette:
Has this been fully merged with the official Smoothie commit? As in, if I were to get the newest firmware.bin file from the smoothie main branch, would this code work in it currently?? Thanks!

wolfmanjm:
no as can be seen it is still open.

GriffinPaquette:
Yep I see now. Any reason it hasn't been merged aside from the lack of testing from other users? I am in the process of setting a board up to test this...

-Griff

sl1288:
The TMC2208 uses an one pin half-duplex UART Interface. This implementation uses two pins on the uC, therefor it is necessary to use an Y-cable to connect the uC to the driver. 
Would it not be better to use just one pin alternating for send and receive ?

Shaische:
@tiagojbalmeida can you please provide a link to a firmware.bin file so we can test this? So far have been unsuccessful at testing this, even by compiling it ourselves. Hoping we can try your firmware.bin file. Thanks!

Shaische:
@tiagojbalmeida Unless I'm doing something wrong, it seems this is not working. Griffin and I tested this. Also noticed that these settings don't have an axis designation in config. You specify the axis with "motor_driver_control.driver.axis" but then how is the TX/RX pins assigned to the X axis in your example if the config setting is simply named "motor_driver_control.driver.sw_uart_tx_pin" ? What if we have multiple motors?

To test, we changed the current between 200 and 1,000 mA to see if there would be any skipping with torque after jogging.

Shaische:
Figured out that you need to change 'driver' to 'alpha' or 'motor1'. (tried both). However, I get the following error when sending M911, but it only happens the first time I send it. Second and consecutive times I send M911, the errors are gone: https://pastebin.com/3mzHTiv7

Current seems to update and reflect what is written in config when sending M911, but LCD keeps resetting (can't fully navigate without going back automatically) and the board keeps getting disconnected from pronterface via USB.

LastDragon-ru:
Hello. I want to test this PR on my little monster with sbase. So, I connected [2208 to sbase like described here](https://www.instructables.com/id/UART-This-Serial-Control-of-Stepper-Motors-With-th/) then merged this PR into `edge` branch, built it and flashed the board with the following config ([full config](https://pastebin.com/gW4TJBxQ))

```
motor_driver_control.alpha.enable               true                # Enable motor driver module
motor_driver_control.alpha.axis                 X                   # Axis designator on which the motor driver is set
motor_driver_control.alpha.chip                 TMC2208             # Motor driver chip type
motor_driver_control.alpha.sw_uart_tx_pin       P1.30               # Transmitting line from the board to the driver
motor_driver_control.alpha.sw_uart_rx_pin       P1.31               # Receiving line to the board
motor_driver_control.alpha.sw_uart_baudrate     9600                # Software UART baudrate. Value higher than 50k is not recommended
motor_driver_control.alpha.current              1000                # Motor currents in mA
motor_driver_control.alpha.sense_resistor       50                  # Sensing resistor value as a reference in milliohms
motor_driver_control.alpha.microsteps           256                 # Set microstepping value per step pulse. It can be 1, 2, 4, 8, 16, 32, 64, 128 or 256 microsteps
motor_driver_control.alpha.chopper_mode         1                   # 0 = spreadCycle ; 1 = stealthChop

motor_driver_control.beta.enable                true                # Enable motor driver module
motor_driver_control.beta.axis                  Y                   # Axis designator on which the motor driver is set
motor_driver_control.beta.chip                  TMC2208             # Motor driver chip type
motor_driver_control.beta.sw_uart_tx_pin        P0.16               # Transmitting line from the board to the driver
motor_driver_control.beta.sw_uart_rx_pin        P0.18               # Receiving line to the board
motor_driver_control.beta.sw_uart_baudrate      9600                # Software UART baudrate. Value higher than 50k is not recommended
motor_driver_control.beta.current               1000                # Motor currents in mA
motor_driver_control.beta.sense_resistor        50                  # Sensing resistor value as a reference in milliohms
motor_driver_control.beta.microsteps            256                 # Set microstepping value per step pulse. It can be 1, 2, 4, 8, 16, 32, 64, 128 or 256 microsteps
motor_driver_control.beta.chopper_mode          1                   # 0 = spreadCycle ; 1 = stealthChop

motor_driver_control.gamma.enable               true                # Enable motor driver module
motor_driver_control.gamma.axis                 Z                   # Axis designator on which the motor driver is set
motor_driver_control.gamma.chip                 TMC2208             # Motor driver chip type
motor_driver_control.gamma.sw_uart_tx_pin       P0.7                # Transmitting line from the board to the driver
motor_driver_control.gamma.sw_uart_rx_pin       P0.8                # Receiving line to the board
motor_driver_control.gamma.sw_uart_baudrate     9600                # Software UART baudrate. Value higher than 50k is not recommended
motor_driver_control.gamma.current              1000                # Motor currents in mA
motor_driver_control.gamma.sense_resistor       50                  # Sensing resistor value as a reference in milliohms
motor_driver_control.gamma.microsteps           256                 # Set microstepping value per step pulse. It can be 1, 2, 4, 8, 16, 32, 64, 128 or 256 microsteps
motor_driver_control.gamma.chopper_mode         1                   # 0 = spreadCycle ; 1 = stealthChop

motor_driver_control.delta.enable               true                # Enable motor driver module
motor_driver_control.delta.axis                 A                   # Axis designator on which the motor driver is set
motor_driver_control.delta.chip                 TMC2208             # Motor driver chip type
motor_driver_control.delta.sw_uart_tx_pin       P0.28               # Transmitting line from the board to the driver
motor_driver_control.delta.sw_uart_rx_pin       P3.25               # Receiving line to the board
motor_driver_control.delta.sw_uart_baudrate     9600                # Software UART baudrate. Value higher than 50k is not recommended
motor_driver_control.delta.current              1000                # Motor currents in mA
motor_driver_control.delta.sense_resistor       50                  # Sensing resistor value as a reference in milliohms
motor_driver_control.delta.microsteps           256                 # Set microstepping value per step pulse. It can be 1, 2, 4, 8, 16, 32, 64, 128 or 256 microsteps
motor_driver_control.delta.chopper_mode         1                   # 0 = spreadCycle ; 1 = stealthChop

motor_driver_control.epsilon.enable             false               # Enable motor driver module
motor_driver_control.epsilon.axis               B                   # Axis designator on which the motor driver is set
motor_driver_control.epsilon.chip               TMC2208             # Motor driver chip type
motor_driver_control.epsilon.sw_uart_tx_pin     P0.9                # Transmitting line from the board to the driver
motor_driver_control.epsilon.sw_uart_rx_pin     P3.26               # Receiving line to the board
motor_driver_control.epsilon.sw_uart_baudrate   9600                # Software UART baudrate. Value higher than 50k is not recommended
motor_driver_control.epsilon.current            1000                # Motor currents in mA
motor_driver_control.epsilon.sense_resistor     50                  # Sensing resistor value as a reference in milliohms
motor_driver_control.epsilon.microsteps         256                 # Set microstepping value per step pulse. It can be 1, 2, 4, 8, 16, 32, 64, 128 or 256 microsteps
motor_driver_control.epsilon.chopper_mode       1                   # 0 = spreadCycle ; 1 = stealthChop
```

But when I send `M911` I receive "ok" instead of drivers status

```
SENT: M911
READ: ok
```

Looks like `motor_driver_control` doesn't work  :( Maybe I missed something?

wolfmanjm:
@LastDragon-ru Hook up the debug uart and see what error messages you get. That is if the sbase has one.
As we do not support MKS you are pretty much on your own to debug this yourself. But the debug uart port should tell you if there are any obvious config errors.

wolfmanjm (approved):

LastDragon-ru:
> That is if the sbase has one.

Sbase has it, but I haven't :(

LastDragon-ru:
Ok, that was pins conflict. Seems work now. But the first `M911` call returns errors:

```
Motor 1 (X)...
designator X, Chip type TMC22X
X - WARNING: Overtemperature Prewarning!
X - ERROR: SHORT to ground on channel A!
X - ERROR: SHORT to ground on channel B!
X - ERROR: Channel B seems to be unconnected!
INFO: Motor is standing still.
Microstep position phase A: -554713427
Current setting: 80mA
Microsteps: 1/256
Register dump:
 gconf register: 00000180(384)
 slaveconf register: 00000000(0)
 ihold_irun register: 00000000(0)
 tpowerdown register: 00000014(20)
 tpwmthrs register: 00000000(0)
 chopconf register: 10020050(268566608)
 pwmconf register: C10D0024(-1056112604)
 motor_driver_control.xxx.reg 00180,00000,00000,00014,00000,10020050,C10D0024
```

wolfmanjm:
@LastDragon-ru that is because you don't have the chip powered up. this is noted in the wiki.

LastDragon-ru:
> Alternatively the Smoothieboard must be hard reset after power is on.

Oh, I see... Is there any other way how I can fix this error without hard reset?

wolfmanjm:
make sure vbb is on before you power on board

LastDragon-ru:
Unfortunately, I'm not sure that this is possible :(

wolfmanjm:
that is how that chip works, it is not a fault of smoothie.

LastDragon-ru:
> it is not a fault of smoothie

Why smoothie cannot wait/sleep until drivers boot?

LastDragon-ru:
Good (undocumented) news: if `sw_uart_tx_pin` = `sw_uart_rx_pin` it works 😄 

Bad news: boot failed if I enable more than 3 drivers 😭

wolfmanjm:
You are probably running out of memory then.
You can disable other modules to make more memory.

wolfmanjm:
You are mistaken you CANNOT use the same pin for TX and RX, as one is input and the other is output and they will fight each other. Which one will win is indeterminate.

src/modules/utils/motordrivercontrol/MotorDriverControl.cpp
```diff
@@ -112,34 +110,79 @@ bool MotorDriverControl::config_module(uint16_t cs)
...
-    // select SPI channel to use
-    PinName mosi, miso, sclk;
-    if(spi_channel == 0) {
-        mosi = P0_18; miso = P0_17; sclk = P0_15;
-    } else if(spi_channel == 1) {
-        mosi = P0_9; miso = P0_8; sclk = P0_7;
+        PinName txd = port_pin((PortName)(sw_uart_tx_pin->port_number<<8), sw_uart_tx_pin->pin);
+        PinName rxd = port_pin((PortName)(sw_uart_rx_pin->port_number<<8), sw_uart_rx_pin->pin);
+
+        this->serial = new BufferedSoftSerial(txd, rxd);
```
wolfmanjm, line 147: As the Pin is not used after this and the BufferedSerial uses MBED pinnames and not the Pin class, why not delete the Pins here as they are never used again. Then move the BufferedSerial into the union instead of the Pin* to save memory?

wolfmanjm (changes requested):

LastDragon-ru:
> You are mistaken you CANNOT use the same pin for TX and RX

image

My first attempt was like the left image, but I haven't 8 free pins on board, so I tried only one wire and 😃 

```
Motor 1 (X)...
designator X, Chip type TMC22X
X - WARNING: Overtemperature Prewarning!
X - ERROR: SHORT to ground on channel A!
X - ERROR: SHORT to ground on channel B!
X - ERROR: Channel B seems to be unconnected!
INFO: Motor is standing still.
Microstep position phase A: -554713427
Current setting: 964mA
Microsteps: 1/256
Register dump:
 gconf register: 00000180(384)
 slaveconf register: 00000000(0)
 ihold_irun register: 00000B00(2816)
 tpowerdown register: 00000014(20)
 tpwmthrs register: 00000000(0)
 chopconf register: 10020050(268566608)
 pwmconf register: C10D0024(-1056112604)
 motor_driver_control.xxx.reg 00180,00000,00B00,00014,00000,10020050,C10D0024
Motor 2 (Y)...
designator Y, Chip type TMC22X
Y - WARNING: Overtemperature Prewarning!
Y - ERROR: SHORT to ground on channel A!
Y - ERROR: SHORT to ground on channel B!
Y - ERROR: Channel B seems to be unconnected!
SENT: M105
INFO: Motor is standing still.
Microstep position phase A: -554713427
Current setting: 964mA
Microsteps: 1/256
Register dump:
 gconf register: 00000180(384)
 slaveconf register: 00000000(0)
 ihold_irun register: 00000B00(2816)
 tpowerdown register: 00000014(20)
 tpwmthrs register: 00000000(0)
 chopconf register: 10020050(268566608)
 pwmconf register: C10D0024(-1056112604)
 motor_driver_control.xxx.reg 00180,00000,00B00,00014,00000,10020050,C10D0024
Motor 3 (B)...
designator B, Chip type TMC22X
B - WARNING: Overtemperature Prewarning!
B - ERROR: SHORT to ground on channel A!
B - ERROR: SHORT to ground on channel B!
B - ERROR: Channel B seems to be unconnected!
INFO: Motor is standing still.
Microstep position phase A: -554713427
Current setting: 964mA
Microsteps: 1/256
Register dump:
 gconf register: 00000180(384)
 slaveconf register: 00000000(0)
 ihold_irun register: 00000B00(2816)
 tpowerdown register: 00000014(20)
 tpwmthrs register: 00000000(0)
 chopconf register: 10020050(268566608)
 pwmconf register: C10D0024(-1056112604)
 motor_driver_control.xxx.reg 00180,00000,00B00,00014,00000,10020050,C10D0024
```

LastDragon-ru:
> You can disable other modules to make more memory.

Thank you, I will try :) (Am I understand correctly that you mean modules in config, not in makefile?)

wolfmanjm:
Using the same pin for RX and TX will not work. If you look at the code you can see a pin can ONLY be either an output OR an input not both.

src/modules/utils/motordrivercontrol/MotorDriverControl.cpp
wolfmanjm: Also maybe have an option to only use the TX pin and then disable all read functions, as that appears to be a recommended use case for this chip (write only). Most people will not have enough memory available anyway to have more than 3 drivers when each has a buffered serial for each pin.

LastDragon-ru:
> Most people will not have enough memory available anyway to have more than 3 drivers when each has a buffered serial for each pin.

I removed unused modules and it boots fine, thanks :) But M911 always says "INFO: Motor is standing still" - is this ok?

vbolshakov:
Not work. Just tried on SKR 1.3 board with 0.0 and 0.1 as TX and RX for X axis.
M911 show status even for not existing TMC2208.
Oscilloscope show serial communication, but no params of TCM2208 changed.
Will try on original Smoothieboard, just need to remove it from CNC.

wolfmanjm:
@vbolshakov P0.0 and P0.1 are not available pins (at least on a genuine smoothieboard)
http://smoothieware.org/lpc1769-pin-usage

vbolshakov:
> @vbolshakov P0.0 and P0.1 are not available pins (at least on a genuine smoothieboard)
> http://smoothieware.org/lpc1769-pin-usage

Yes, i know - i just need more time to extract original board from my CNC and try.

LastDragon-ru:
> Your version request interrupt capable pin, "original" not.

@vbolshakov, AFAIK `RX` pin must be interrupt capable in any case (the #1377 has only one pins change https://github.com/Smoothieware/Smoothieware/pull/1377#pullrequestreview-281002030). And this is one of the reason why I added "write only" mode (just do not set `RX` or set it to `NC`).

vbolshakov:
Not work on original smoothieboard (v1.0b)
Smoothieboard with TMC2208

wolfmanjm:
ok so here is what I am thinking.... As this chip has OTP which is programmed by a pseudo serial port, it is really designed to be setup off board then plugged into a board. IMO this PR is not needed on smoothie at all, as you should really simply program the drivers OTP from a FTDI using the very nice setup S/W provided then plug it into your controller board. Due to this I am leaning toward closing this PR. (The SPI drivers do not have OTP and need to be configured each time they are powered up).

vbolshakov:
I think about dynamic current control. And maybe microsteps.

wolfmanjm:
IMO neither of those should be dynamic anyway. Besides the software serial must not be used other than during setup, it is not designed to be used while running.

LastDragon-ru:
> IMO this PR is not needed on smoothie at all, as you should really simply program the drivers OTP from a FTDI using the very nice setup S/W provided then plug it into your controller board

2208 cannot save settings (OTP is the bad solution imho) so you should write setting every time after power on. Moreover, if you want fast speed/acceleration you probably must enable hybrid mode and set `TPWMTHRS` or it can randomly stop while printing (but this PR doesn't support `TPWMTHRS` at all). This is the main reason why UART is important. The second - UART may be useful to set microstep resolution and current.

> IMO neither of those should be dynamic anyway.

2208 have over-heat and over-current ("random stop") protections when this happened it require a power reboot to resume normal operation. So you should monitor the printing process and reboot the board if something bad happened... OR board can read the status of the drivers by UART and stop printing by himself (third reason) :)

wolfmanjm:
closed in favor of PR #1447

PR #1331, opened 2018-06-10
Replace spaces in names with underscores by default

lordofhyphens:
Smoothieware's console doesn't play nice with spaces (at least as I've seen) in filenames. This patch replaces them with underscores by default in smoothie-upload.py and introduces the _-s_ flag to allow for the old behavior.


arthurwolf:
Thanks !

PR #1349, opened 2018-10-04
Z-Homing of rack and pinion head configuration "Peter's Head"

sl1288:
This pull request features an option "home_z_rack_and_pinion" to be able to home the Z-axis of an rack and pinion head also known as "Peter's Head" (http://www.betztechnik.ca/store/p34/Pick_and_Place_head_-_dual_nozzle_-_OpenPnP_compatible.html). To home this head it is necessary to dynamically set the homing direction and endstop configuration depending on the current status of the sensor.


wolfmanjm:
This seems of limited usefulness, so will not merge into code base. It is suggested you keep your own fork for your machine.

PR #1351, opened 2018-11-19
MISTAKENLY committed - Read pcb revision from gpio state

andySigler:
## overview

Adds functionality to read the PCB revision from gpio state.

Also adds new homing/probing command `G28.7`, which allows an axis to probe the distance between it's current position, and when either the `B` or `C` axis is triggered.


wolfmanjm:
too many different things submitted please read the contribution guidelines.

Thank you anyway.

PR #1354, opened 2018-12-07
Added support for real scara math

alexkv:
- Adding calculation for real SCARA support in MorganSCARASolution.
- Adding switcher to the config file that enables this calculation. By default this param is false.


wolfmanjm:
I suspect @qharley will object to the naming of the option being 'real_scara' :)

Thank you for the submission.

PR #1371, opened 2019-03-02
Adds program memory protection and optional manual probe attachment

qharley:
This pull request adds a transparent method to prevent unnecessary program memory writes.
Allocation of the buffer always occurs at the same physical program memory position between resets.  The initialisation will then clear and rewrite the same data to the position.
This update will have the system check if the data is the same, and then skip the write if so.

There is also the addition of an optional (default disabled) manual probe attachment section when the grid is run.  It will allow you to set an attachment point and will continue the action when the probe button is pressed.

### DIFF
--- a/src/modules/tools/zprobe/CartGridStrategy.cpp
+++ b/src/modules/tools/zprobe/CartGridStrategy.cpp
@@ -106,6 +106,8 @@
 #define x_size_checksum              CHECKSUM("x_size")
 #define y_size_checksum              CHECKSUM("y_size")
 #define do_home_checksum             CHECKSUM("do_home")
+#define m_attach_checksum            CHECKSUM("m_attach")
+#define mount_position_checksum      CHECKSUM("mount_position")
 #define only_by_two_corners_checksum CHECKSUM("only_by_two_corners")
 #define human_readable_checksum      CHECKSUM("human_readable")
 #define height_limit_checksum      CHECKSUM("height_limit") 
@@ -135,7 +137,8 @@ bool CartGridStrategy::handleConfig()
     do_home = THEKERNEL->config->value(leveling_strategy_checksum, cart_grid_leveling_strategy_checksum, do_home_checksum)->by_default(true)->as_bool();
     only_by_two_corners = THEKERNEL->config->value(leveling_strategy_checksum, cart_grid_leveling_strategy_checksum, only_by_two_corners_checksum)->by_default(false)->as_bool();
     human_readable = THEKERNEL->config->value(leveling_strategy_checksum, cart_grid_leveling_strategy_checksum, human_readable_checksum)->by_default(false)->as_bool();
- 
+    do_manual_attach = THEKERNEL->config->value(leveling_strategy_checksum, cart_grid_leveling_strategy_checksum, m_attach_checksum)->by_default(false)->as_bool();
+
     this->height_limit = THEKERNEL->config->value(leveling_strategy_checksum, cart_grid_leveling_strategy_checksum, height_limit_checksum)->by_default(NAN)->as_number();
     this->dampening_start = THEKERNEL->config->value(leveling_strategy_checksum, cart_grid_leveling_strategy_checksum, dampening_start_checksum)->by_default(NAN)->as_number();
 
@@ -167,6 +170,16 @@ bool CartGridStrategy::handleConfig()
         }
     }
 
+    //  manual attachment point xxx,yyy,zzz
+    if (do_manual_attach)
+    {
+        std::string ap = THEKERNEL->config->value(leveling_strategy_checksum, cart_grid_leveling_strategy_checksum, mount_position_checksum)->by_default("0,0,50")->as_string();
+        std::vector<float> w = parse_number_list(ap.c_str());
+        if(w.size() >= 3) {
+            this->m_attach = std::make_tuple(w[0], w[1], w[2]);
+        }
+    }
+
     // allocate in AHB0
     grid = (float *)AHB0.alloc(configured_grid_x_size * configured_grid_y_size * sizeof(float));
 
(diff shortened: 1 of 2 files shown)


src/modules/tools/zprobe/CartGridStrategy.cpp
```diff
@@ -175,7 +188,8 @@ bool CartGridStrategy::handleConfig()
         return false;
     }
 
-    reset_bed_level();
```
wolfmanjm, line 178: This makes absolutely no sense at all. It is RAM, it is totally irrelevant whether you read or write to RAM. Why do you believe this is necessary? What is the advantage?

I recommend you submit these as two separate requests as the manual probe stuff I may consider but this reducing memory writes is nonsense IMO.

wolfmanjm (changes requested):

src/modules/tools/zprobe/CartGridStrategy.cpp
```diff
@@ -307,13 +321,21 @@ bool CartGridStrategy::load_grid(StreamOutput *stream)
...
         for (int x = 0; x < configured_grid_x_size; x++) {
-            if(fread(&grid[x + (configured_grid_x_size * y)], sizeof(float), 1, fp) != 1) {
+            float read_value;
+            if(fread(&read_value , sizeof(float), 1, fp) != 1) {
                 stream->printf("error:Failed to read grid\n");
                 fclose(fp);
                 return false;
+            } else {
+                if (read_value != grid[x + (configured_grid_x_size * y)]) {
+                    grid[x + (configured_grid_x_size * y)] = read_value;
```
wolfmanjm, line 331: This optimizes nothing it is RAM it makes no difference if it is written to or not. This is not an optimization IMO, it just complicates the code and adds an extra variable.

src/modules/tools/zprobe/CartGridStrategy.cpp
qharley: It is not RAM.  We allocate memory in the program memory block.  It is Flash if I am not mistaken?
If you do not clear it, the values remain even after resetting.

src/modules/tools/zprobe/CartGridStrategy.cpp
```diff
@@ -506,6 +528,23 @@ bool CartGridStrategy::doProbe(Gcode *gc)
...
+        float x, y, z;
+        std::tie(x, y, z) = m_attach;
+        zprobe->coordinated_move( x, y, z , zprobe->getFastFeedrate());
+
+        gc->stream->printf(" ************************************************************\n");
+        gc->stream->printf("     Ensure probe is attached and trigger probe when done\n");
+        gc->stream->printf(" ************************************************************\n");
+
+        while( !zprobe->getProbeStatus()) {
+            THEKERNEL->call_event(ON_IDLE);
```
wolfmanjm, line 545: This needs to check for halt condition so it does not loop indefinitely. Adding this will fix it...
`if(THEKERNEL->is_halted()) break;`

wolfmanjm (changes requested):

src/modules/tools/zprobe/CartGridStrategy.cpp
qharley: Again... not RAM.  There was not enough RAM to implement the function so we allocated the block in program memory.  It has limited write cycles, and some of my older systems are starting to fail randomly especially in the z leveling

src/modules/tools/zprobe/CartGridStrategy.cpp
wolfmanjm: It is RAM, AHB0 is RAM! Exactly the Same as main RAM, it is just one of the three non contiguous blocks of RAM in the LPC1769.

src/modules/tools/zprobe/CartGridStrategy.cpp
qharley: Before the while?

src/modules/tools/zprobe/CartGridStrategy.cpp
wolfmanjm: May I suggest you dig out the LPC1768 manual and read it ;)

src/modules/tools/zprobe/CartGridStrategy.cpp
qharley: Interesting.
Please ignore this pull request then.  I will do more faultfinding to find the cause of my problems

qharley: I will separate the two features

src/modules/tools/zprobe/CartGridStrategy.cpp
```diff
@@ -506,6 +528,24 @@ bool CartGridStrategy::doProbe(Gcode *gc)
...
+        float x, y, z;
+        std::tie(x, y, z) = m_attach;
+        zprobe->coordinated_move( x, y, z , zprobe->getFastFeedrate());
+
+        gc->stream->printf(" ************************************************************\n");
+        gc->stream->printf("     Ensure probe is attached and trigger probe when done\n");
+        gc->stream->printf(" ************************************************************\n");
+
+        while( !zprobe->getProbeStatus()) {
+            if(THEKERNEL->is_halted()) break;
```
wolfmanjm, line 544: Actually `return false;` would be more appropriate than `break` if you are not going to test when it exists the loop. Basically this would abort the probe if control X is hit (or the kill button).

src/modules/tools/zprobe/CartGridStrategy.cpp
wolfmanjm: FWIW the only time we ever write to flash is when you flash a new firmware binary which I believe you could do 100,000 times before wearing out the memory. (it may be 10,000 but still more than anyone is likely to do). Also you cannot randomly write to flash, you must first erase a block then write a block (I think a block is 4k bytes).

PR #1372, opened 2019-03-02
Adds an optional manual probe attachment position.

qharley:
Adds a manual probe attachment option

### DIFF
--- a/src/modules/tools/zprobe/CartGridStrategy.cpp
+++ b/src/modules/tools/zprobe/CartGridStrategy.cpp
@@ -110,7 +110,7 @@
 #define mount_position_checksum      CHECKSUM("mount_position")
 #define only_by_two_corners_checksum CHECKSUM("only_by_two_corners")
 #define human_readable_checksum      CHECKSUM("human_readable")
-#define height_limit_checksum      CHECKSUM("height_limit") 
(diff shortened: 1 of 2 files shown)


src/modules/tools/zprobe/CartGridStrategy.cpp
```diff
@@ -506,6 +519,23 @@ bool CartGridStrategy::doProbe(Gcode *gc)
...
+        float x, y, z;
+        std::tie(x, y, z) = m_attach;
+        zprobe->coordinated_move( x, y, z , zprobe->getFastFeedrate());
+
+        gc->stream->printf(" ************************************************************\n");
+        gc->stream->printf("     Ensure probe is attached and trigger probe when done\n");
+        gc->stream->printf(" ************************************************************\n");
+
+        while( !zprobe->getProbeStatus()) {
+            THEKERNEL->call_event(ON_IDLE);
```
wolfmanjm, line 536: need to add if(THEKERNEL->is_halted()) return(false);
To avoid hanging the smoothie in the case of no zprobe.

wolfmanjm (changes requested):

wolfmanjm (approved):

Issue #1379, opened 2019-03-11
CartGridStrategy::doCompensation() does not work for right and top edges of probed area


mdelanno:
The [documentation ](http://smoothieware.org/zprobe#rectangular-grid-compensation)states that:

> the very latest firmware will not jump when outside the probed area, it will just use the nearest edge heights instead.

But when I'm looking to the code, I see that in this case, the computed indices to access the grid values are wrong. If we have a 7x7 grid, this code:

`int floor_y = (int) Floor(grid_y);`

will put 6 (the max) in `floor_y`.

After that, the following code is used to read the point on the right:

`float z4 = grid[floor_x + 1 + (floor_y + 1) * current_grid_x_size];`

The problem is that one unit is added to a coordinate which is already to the max : `floor_y + 1`

The computed index will be above the bounds of the array, making the code fails (I don't know how exactly but when I'm testing this, my printer is going crazy).

The firmware should halt the printer in this case because there is no way it can compute the compensation. Even extrapolation can produce results which are not good because the bed is not supposed to be flat and the user will experiment a strange behavior.

wolfmanjm:
I don't think you are taking into consideration the previous lines of code, which clamps the area correctly.

```
    // find min/maxes, and handle the case where size is negative (assuming this is possible? Legacy code supported this)
    float min_x = std::min(this->x_start, this->x_start + this->x_size);
    float max_x = std::max(this->x_start, this->x_start + this->x_size);
    float min_y = std::min(this->y_start, this->y_start + this->y_size);
    float max_y = std::max(this->y_start, this->y_start + this->y_size);

    // clamp the input to the bounds of the compensation grid
    // if a point is beyond the bounds of the grid, it will get the offset of the closest grid point
    float x_target = std::min(std::max(target[X_AXIS], min_x), max_x);
    float y_target = std::min(std::max(target[Y_AXIS], min_y), max_y);

    float grid_x = std::max(0.001F, (x_target - this->x_start) / (this->x_size / (this->current_grid_x_size - 1)));
    float grid_y = std::max(0.001F, (y_target - this->y_start) / (this->y_size / (this->current_grid_y_size - 1)));

```

A quick scan of the code looks correct to me, as far as limiting the scope to the probed area (NOTE further comments which show where the actual bug is).

wolfmanjm:
Also your grid is 7x7, and the floor_y is 6, 6+1 == 7, thus it is NOT out of bounds.
[...]

PR #1380, opened 2019-03-11
Fix cartesian grid probe for right and top edges which currently go ou…

wolfmanjm:
…t of bounds

This should fix issue #1379


mdelanno:
Works ok on my printer. I can go outside of the probed area.

PR #1391, opened 2019-04-14
 add before probe and after probe optional gcodes to cart grid

wolfmanjm:
1. Allow cart grid to send optional gcodes before and after probing, basically to automatically deploy a bltouch (or any deployable probe).

2. Allow park after homing, this allows going to a predefined position (the park position) after homing instead of going to 0,0. This is useful for when the probe is offset from the head so much that probing cannot go to 0,0 due to the probe offsets.

3. Fix $J to not turn on laser.

4. Fix crashes when saving grids to sdcard.

### DIFF
--- a/src/modules/robot/Robot.cpp
+++ b/src/modules/robot/Robot.cpp
@@ -1394,6 +1394,7 @@ bool Robot::delta_move(const float *delta, float rate_mm_s, uint8_t naxis)
         target[i] += delta[i];
     }
 
+    is_g123= false; // we don't want the laser to fire
     // submit for planning and if moved update machine_position
     if(append_milestone(target, rate_mm_s)) {
          memcpy(machine_position, target, n_motors*sizeof(float));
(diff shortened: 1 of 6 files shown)


qharley:
Thanks!

gena-bayanov:
Could you please implement the same for ZProbe module? Or it would be even better to move this functionality to the zprobe.

wolfmanjm:
no, I specifically didn't do that for various reasons. One being is you do not want it to happen every time the probe dips. you need to know when the operation begins and ends

gena-bayanov:
I think if you do not want it to happen every time you do not specify before/after gcodes for zprobe.
For BLTouch it is better to stow the pin after each probe. Marlin does it if you define BLTOUCH macro, see video: https://youtu.be/sUlqrSq6LeY?t=785 as you can see the pin stows after each probe test.
Also, you can add an additional parameter to cart grid like "override_zprobe_gcodes" and if it is set to true use gcodes from cart grid and ignore from zprobe.
It is not convenient to manually send codes to deploy/stow probe, it should be done automatically.
Thanks

PR #1368, opened 2019-02-26
 Added RRDGLCD support for the new Watch Screen.

garikdan:
- Added graphics functions.
pixel
drawLine
drawHLine
drawVLine
drawBox
- Added function to set the cursor position.
setCursorPX and setCursor.
- Improved functions.
displayChar
displayString
renderChar
- Changed deg symbol.

PR #1393, opened 2019-04-15
Graphical menu screens

unlimitedbacon:
## Changes
* Added titles to the menu screens
* Added a scroll bar
* Inverted the color of the selected item.

## Todo
Right now this breaks the Probe > Status display, because it writes directly to the screen instead of using `display_menu_line()`. I intend to fix this by making a new screen to show the endstop status. This will come as a separate PR.

### DIFF
--- a/src/modules/utils/panel/Panel.cpp
+++ b/src/modules/utils/panel/Panel.cpp
@@ -511,16 +511,18 @@ void Panel::enter_menu_mode(bool force)
     encoder_cb_fnc= nullptr;
 }
 
-void Panel::setup_menu(uint16_t rows)
+void Panel::setup_menu(uint16_t rows, bool reset_pos)
 {
-    this->setup_menu(rows, min(rows, this->max_screen_lines()));
+    this->setup_menu(rows, min(rows, this->max_screen_lines()), reset_pos);
 }
 
-void Panel::setup_menu(uint16_t rows, uint16_t lines)
+void Panel::setup_menu(uint16_t rows, uint16_t lines, bool reset_pos)
 {
-    this->menu_selected_line = 0;
-    this->menu_current_line = 0;
-    this->menu_start_line = 0;
+    if (reset_pos) {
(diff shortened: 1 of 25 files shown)


wolfmanjm:
I'm sorry but I cannot accept something that breaks existing functionality, and I cannot accept that being fixed in a separate PR. There is no reason it cannot be fixed in this PR, and it must be.
Also as this changes many files but there is still no rrd glcd implementation for now I cannot accept this.

Once that is all working it can be merged. For now it is closed, however you can still commit changes to your branch which will show up here.

wolfmanjm:
Actually I'll leave it open.

I do not think adding a title but losing two lines is a good trade off IMO.

I'll just mark this as a work in progress.

arthurwolf:
Thanks ! Any way you can free up more space ? I don't think we need the box all around the screen, maybe try just removing that and keeping the rest ? I wonder how it'd look.

unlimitedbacon:
Sorry, I was just following the rules.
> Only one feature per pull request, if you change multiple files with unrelated changes it will be rejected.

Reformatting the probe status display sounds like a separate discrete change.

It looks like someone else was working on the RRDGLCD support (https://github.com/Smoothieware/Smoothieware/pull/1368). Whatever came of this?

wolfmanjm:
It is very much related, this PR breaks existing functionality, and a fix would be dependent on this PR, and would not even pass build unless this was merged first, so it is one atomic piece of code.

The guy deleted his branch before I could do anything with it.

wolfmanjm:
Have you looked at how any of this affects the CNC screens? do they even build with these changes?

unlimitedbacon:
Ok, the probe screen is fixed.

The CNC version builds and works fine, as far as I can tell.

The guy deleted his RRD branch, but it looks like you can still find the code he had if you go to the PR and look at the files changed. If anything, we could just copy and paste that.

PR #1398, opened 2019-05-13
refactor max actuator speeds and acceleration using major axis motion

wolfmanjm:
A correct fix to the rotary actuator acceleration issue, for instance ```G1 X0.001 A360```.

It turned out this was actually a planner issue, where step generation uses a major axis (the one with the most steps) for all its calculations, but it expects the distance to be related to that major axis.
Now we calculate and pass the distance for the major axis, keeping that contract in place. This means that we decide if this is a primary axis move or auxiliary moved based on the major axis.

### DIFF
--- a/src/modules/robot/Robot.cpp
+++ b/src/modules/robot/Robot.cpp
@@ -1323,7 +1323,7 @@ bool Robot::append_milestone(const float target[], float rate_mm_s)
     }
     if(auxilliary_move) {
         distance= sqrtf(sos); // distance in mm of the e move
-        if(distance < 0.00001F) return false; // avoid divide by zero furter on
+        if(distance < 0.00001F) return false; // avoid divide by zero further on
     }
 #endif
 
@@ -1340,7 +1340,7 @@ bool Robot::append_milestone(const float target[], float rate_mm_s)
         // actual distance moved for this actuator
         // NOTE for a rotary axis this will be degrees turned not distance
         float d = fabsf(actuator_pos[actuator] - actuators[actuator]->get_last_milestone());
-        if(d == 0 || !actuators[actuator]->is_selected()) continue; // no movement for this actuator
+        if(d < 0.00001F || !actuators[actuator]->is_selected()) continue; // no perceptible movement for this actuator
 
         // find approximate min time this axis needs to move its distance
         float actuator_min_time= d / actuators[actuator]->get_max_rate();
--- a/src/modules/robot/Robot.cpp
+++ b/src/modules/robot/Robot.cpp
@@ -1261,13 +1261,26 @@ bool Robot::append_milestone(const float target[], float rate_mm_s)
     if(!move) return false;
 
     // see if this is a primary axis move or not
-    bool auxilliary_move= true;
-    for (int i = 0; i < N_PRIMARY_AXIS; ++i) {
-        if(deltas[i] != 0) {
-            auxilliary_move= false;
-            break;
+    // do this by finding the major axis based on steps moved not distance
+    // if we find a non primary axis (other than E) is moving more than the primary axis
+    // we treat it as an auxilliary move
+    bool auxilliary_move= false;
+    int major_axis= 0;
+    uint32_t max_steps= 0;
+    for (int i = 0; i < n_motors; ++i) {
+        // collect the number of steps for each actuator and remember the major axis
+        uint32_t s= labs(actuators[i]->steps_to_target(deltas[i]));
+        if(s > max_steps) {
+            max_steps= s;
+            major_axis= i;
         }
     }
+
+    DEBUG_PRINTF("major axis is : %d\n", major_axis);
+
+    // find the major axis which is the one with the most steps, if
+    // this is a non extruder but non primary axis then treat this as an auxilliary move
+    auxilliary_move = (major_axis >= N_PRIMARY_AXIS);
 
     // total movement, use XYZ if a primary axis otherwise we calculate distance for E after scaling to mm
     float distance= auxilliary_move ? 0 : sqrtf(sos);
@@ -1309,7 +1322,7 @@ bool Robot::append_milestone(const float target[], float rate_mm_s)
 
 #if MAX_ROBOT_ACTUATORS > 3
     sos= 0;
-    // for the extruders just copy the position, and possibly scale it from mm³ to mm
+    // for the extruders and rotary axis just copy the position, and possibly scale it from mm³ to mm
     for (size_t i = E_AXIS; i < n_motors; i++) {
         actuator_pos[i]= transformed_target[i];
         if(actuators[i]->is_extruder() && get_e_scale_fnc) {
@@ -1320,7 +1333,7 @@ bool Robot::append_milestone(const float target[], float rate_mm_s)
             actuator_pos[i] *= get_e_scale_fnc();
         }
         if(auxilliary_move) {
-            // for E only moves we need to use the scaled E to calculate the distance
+            // for E and rotary axis moves we need to use the scaled E or ABC to calculate the distance
             sos += powf(actuator_pos[i] - actuators[i]->get_last_milestone(), 2);
         }
     }
@@ -1350,25 +1363,10 @@ bool Robot::append_milestone(const float target[], float rate_mm_s)
         if (secs < actuator_min_time ) {
             // this move at the default rate will move too quickly for this actuator
             // so we decrease the overall feed rate so it can complete within the min time for this actuator
-            if(actuator >= N_PRIMARY_AXIS && !actuators[actuator]->is_extruder()) {
-                // if this is a rotary axis we check what the time would be if the given feedrate was
-                // for a solo move in degrees/sec so we do not go faster than that
-                // this is for rotary engraving where the feedrate is calculated to maintain
-                // a certain speed over the surface.
-                actuator_min_time= std::max(actuator_min_time, d/rate_mm_s);
-            }
-            if(!auxilliary_move && actuator >= N_PRIMARY_AXIS) {
-                // if this is a rotary axis we need to increase the distance
-                distance *= actuator_min_time/secs;
-                secs= actuator_min_time;
-                DEBUG_PRINTF("new distance: %f - %d\n", distance, actuator);
-            }else{
-                // primary axis we just change the feed rate
-                rate_mm_s= distance / actuator_min_time;
-                DEBUG_PRINTF("new rate: %f - %d\n", rate_mm_s, actuator);
-                // recalculate time from new rate
-                secs = distance / rate_mm_s;
-            }
+            rate_mm_s= distance / actuator_min_time;
+            DEBUG_PRINTF("new rate: %f - %d\n", rate_mm_s, actuator);
+            // recalculate time from new rate
+            secs = distance / rate_mm_s;
         }
 
         DEBUG_PRINTF("act: %d, d: %f, distance: %f, min: %f, rate: %f, secs: %f, acc: %f\n", actuator, d, distance, actuator_min_time, rate_mm_s, secs, acceleration);
--- a/src/modules/robot/Robot.cpp
+++ b/src/modules/robot/Robot.cpp
@@ -1308,7 +1308,9 @@ bool Robot::append_milestone(const float target[], float rate_mm_s)
 
     // total movement, use XYZ if a primary axis otherwise we calculate distance for E after scaling to mm
     // or ABC
-    float distance= 0.0F;
+    float p_distance= sqrtf(xyz_sos);
+    float a_distance= sqrtf(abc_sos);
+    float distance;
 
     // find the major axis which is the one with the most steps, if
     // this is an extruder solo move or an ABC move then treat this as an auxilliary move
@@ -1317,23 +1319,23 @@ bool Robot::append_milestone(const float target[], float rate_mm_s)
     bool auxilliary_move= false;
     if(major_axis < N_PRIMARY_AXIS) {
         auxilliary_move= false;
-        distance= sqrtf(xyz_sos);
+        distance= p_distance;
 
     }else if(actuators[major_axis]->is_extruder() && xyz_sos > 0.00001F) {
         // if it is an extruder but there is any XYZ move then not an auxilliary move
         auxilliary_move= false;
-        distance= sqrtf(xyz_sos);
+        distance= p_distance;
 
-    }else if(xyz_sos > 1.0F) {
+    }else if(xyz_sos >= 0.25F) { // this is about 0.5mm
         // it is a ABC axis move but there is a significant XYZ move so not an auxilliary move
         // but we use the ABC movement as distance
         auxilliary_move= false;
-        distance= sqrtf(abc_sos);
+        distance= a_distance;
 
     }else{
         // if ABC axis is a major axis or a solo move then it is an auxilliary move
         auxilliary_move= true;
-        distance= sqrtf(abc_sos);
+        distance= a_distance;
     }
 
     DEBUG_PRINTF("major axis is : %d, distance: %f, aux_move: %d\n", major_axis, distance, auxilliary_move);
@@ -1347,7 +1349,7 @@ bool Robot::append_milestone(const float target[], float rate_mm_s)
         // calculate unit vector for junction deviation if a primary axis move
         for (size_t i = X_AXIS; i < N_PRIMARY_AXIS; i++) {
             // find distance unit vector for primary axis only
-            unit_vec[i] = deltas[i] / distance;
+            unit_vec[i] = deltas[i] / p_distance;
 
             // Do not move faster than the configured cartesian limits for XYZ
             if ( i <= Z_AXIS && max_speeds[i] > 0 ) {
--- a/src/modules/robot/Robot.cpp
+++ b/src/modules/robot/Robot.cpp
@@ -1349,21 +1349,26 @@ bool Robot::append_milestone(const float target[], float rate_mm_s)
             // so we decrease the overall feed rate so it can complete within the min time for this actuator
             rate_mm_s= distance / actuator_min_time;
             // recalculate time from new rate
-            secs = distance / rate_mm_s;
+            float new_secs = distance / rate_mm_s;
+            // we need to reduce the acceleration by the same ratio
+            acceleration *= secs/new_secs;
+            secs= new_secs;
         }
+
+        THEKERNEL->streams->printf("act: %d, d: %f, min: %f, rate: %f, secs: %f, acc: %f\n", actuator, d, actuator_min_time, rate_mm_s, secs, acceleration);
 
         // we do not bother with extruder acceleration, becuase if we did in a 3d printer we would always
         // be using the extruders acceleration and not the XY acceleration which is usually a lot higher
         // and extruder moves are usually tiny
-        if(!auxilliary_move && actuators[actuator]->is_extruder()) continue;
+        // if(!auxilliary_move && actuators[actuator]->is_extruder()) continue;
 
-        // select the lowest acceleration of the axis in motion
-        float ma = actuators[actuator]->get_acceleration(); // in mm/sec²
-        if(!isnan(ma)) {  // if axis does not have acceleration set then it uses the default_acceleration
-            if (acceleration > ma) {
-                acceleration= ma;
-            }
-        }
+        // // select the lowest acceleration of the axis in motion
+        // float ma = actuators[actuator]->get_acceleration(); // in mm/sec²
+        // if(!isnan(ma)) {  // if axis does not have acceleration set then it uses the default_acceleration
+        //     if (acceleration > ma) {
+        //         acceleration= ma;
+        //     }
+        // }
     }
 
     // if we are in feed hold wait here until it is released, this means that even segmented lines will pause
--- a/src/modules/robot/Robot.cpp
+++ b/src/modules/robot/Robot.cpp
@@ -1278,10 +1278,24 @@ bool Robot::append_milestone(const float target[], float rate_mm_s)
     DEBUG_PRINTF("major axis is : %d\n", major_axis);
 
     // find the major axis which is the one with the most steps, if
-    // this is a non extruder but non primary axis then treat this as an auxilliary move
-    bool auxilliary_move = (major_axis >= N_PRIMARY_AXIS);
+    // this is an extruder solo move or an ABC move then treat this as an auxilliary move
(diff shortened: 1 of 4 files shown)


src/modules/robot/Robot.cpp
```diff
@@ -1352,19 +1352,16 @@ bool Robot::append_milestone(const float target[], float rate_mm_s)
...
+        if(!auxilliary_move && actuators[actuator]->is_extruder()) continue;
 
-        // adjust acceleration to not exceed the actuators acceleration, based on the ratio it
-        // presents for the overall move (NOTE this may be incorrect)
-        float ma =  actuators[actuator]->get_acceleration(); // in mm/sec²
+        // select the lowest acceleration of the axis in motion
+        float ma = actuators[actuator]->get_acceleration(); // in mm/sec²
         if(!isnan(ma)) {  // if axis does not have acceleration set then it uses the default_acceleration
-            // this is the proportion of the total move scaling the acceleration
-            float ca = fabsf((d/distance) * acceleration);
```
markmaker: Hi @wolfmanjm 

EDIT: like mentioned in the comment, this is [SOLVED].

This scaling `d/distance` is needed, because the whole move is **coordinated** and each axis motion is proportional to the `d/distance` including speed and acceleration.   

`distance` is the Euclidean XYZ distance, it does not include the ABC components. Which is correct according to the NIST standard.

`d` is the distance of the axis at hand. If it is an ABC axis, it can be much larger than `distance`, because it is not in the Euclidean sum. Consequently the scaling `d/distance` can be much larger that 1.0, Consequently the scaled acceleration `ca` can easily be larger than the actuator maximum acceleration `ma` and therefore the overall acceleration must be reduced by `( ma / ca )`.

The old code [#L1363-L1367](#L1363-L1367) is correct.

If you watch my video, that is exactly the case. X, Y is very small (few tenths of millimeters) compensating the nozzle runout, so we have a very small Euclidean `distance` (again, few tenths), while the nozzle rotation is reset by 180°, i.e. a very large `d` in comparison. The `acceleration` in X, Y must be therefore be reduced dramatically for the C stepper not to stall.

EDIT (added): the opposite is also true: Let's assume you have a very "heavy" Y axis that has a low maximum accelerated (typical for a heavy portal CNC mill or PNP), but a light X axis that can speed up quickly. Then assume you have a relative `G0 X100 Y1`. Now it would be a waste to limit the acceleration of X by the unscaled max acceleration of Y. Because Y ist only moved ~1% of the Euclidean distance, the acceleration of Y will be only 1% of the overall acceleration, so it will never be limited by the Y actuator maximum. It would be wasteful not to scale. Again the old code is correct.  

_Mark

wolfmanjm:
Unfortunately this still does not fix the G1 X0.001 A360 issue with instant acceleration. turns out that is a planner issue which can't handle the very small distance but long execution time. I'll continue to see how to fix this.

markmaker:
Just keep the old lines 1363-1367
https://github.com/Smoothieware/Smoothieware/blob/e61107b5f0a56dd10232404f872cb3b6d61d17cf/src/modules/robot/Robot.cpp#L1363-L1367
and it will work.

_Mark

wolfmanjm:
Actually I found the bug in the planner, it was causing instant acceleration, and is quite easily fixed by reducing the acceleration if I have to reduce the rate. Now I just need to figure out the acceleration limiting.
However G1 X0.01 A360 now works on my rotary axis., There is still a problem that this now breaks the rotary engraving as it no longer honors the calculated feed rate to keep the movement over the surface of a cylinder constant

src/modules/robot/Robot.cpp
```diff
@@ -1340,31 +1340,45 @@ bool Robot::append_milestone(const float target[], float rate_mm_s)
...
             // so we decrease the overall feed rate so it can complete within the min time for this actuator
-            rate_mm_s= distance / actuator_min_time;
-            // recalculate time from new rate
-            secs = distance / rate_mm_s;
+            if(actuator >= N_PRIMARY_AXIS && !actuators[actuator]->is_extruder()) {
+                // if this is a rotary axis we check what the time would be if the given feedrate was
+                // for a solo move in degrees/sec so we do not go faster than that
+                // this is for rotary engraving where the feedrate is calculated to maintain
+                // a certain speed over the surface.
+                actuator_min_time= std::max(actuator_min_time, d/rate_mm_s);
```
markmaker: Hi @wolfmanjm 

am I interpreting this correctly? you want to drive the actuator outside of its specs i.e. above `actuators[actuator]->get_max_rate()` to keep up the feedrate that the G code requested?

I'm not sure that is a good idea. The requested G code feedrate can always just be a goal. It is always subject to machine constraints. If the actuator has to accelerate first, the feedrate will equaly not be met at first. 

However all moves are coordinated, so the axes will always be in sync to each other. If the rotation axis limits the overall toolpath feedrate, then the other axes are slowed to accomodate it. 

If a minmum feedrate is critical for the engraving (material melts otherwise?) the user must upgrade the rotation motor or change the engraving pattern. Stalling the motor won't help the engraving result either :-)

_Mark

markmaker (commented):
Hi @wolfmanjm 

I'm glad you changed the acceleration scaling back. This makes my first review comment there obsolete. Just wanted to submit it anyway, to document the rationale behind it all for posterity. 

The second comment is about the maximum rate for engraving. 

_Mark

wolfmanjm:
Thank you for the comments, however I think you are underestimating the complexity. The bug is actually in the planner where the speeds get so low the calculations overflow and this is what causes the instant acceleration as the ticks to accelerate are 0 which is always incorrect. Your change to acceleration by dividing two totally unrelated values (distance in mm vs d in degrees) is also incorrect. The only reason reducing the acceleration works is that it gets the planner calculations back into range, but a better fix is to fix the cases in the planner where the acceleration ticks become 0. (This may be to force a lower limit in feedrate which is what GRBL does, at the expense of loss of sync movement).
My fixes do fix the G1 X0.01 A360 issue, but introduce other unacceptable side effects so I need to find a different solution to this problem, and it does not include hacking the acceleration to make planner happy. The acceleration should always be the given acceleration, and making sure an axis does not exceed its acceleration override if given.

As for speed over the surface for engraving this is what Lightburn calculates for instance and it is necessary to honor that, and no it does not drive a rotation axis faster as it takes the minimum of the given feedrate vs the maximum feedrate for that axis. It is essential to drive the laser over the surface at a fixed rate.

markmaker:
> however I think you are underestimating the complexity. The bug is actually in the planner where the speeds get so low the calculations overflow and this is what causes the instant acceleration as the ticks to accelerate are 0 which is always incorrect. Your change to acceleration by dividing two totally unrelated values (distance in mm vs d in degrees) is also incorrect

I humbly disagree with all of that. 

I'm not nearly as fast and frequent as you are with throwing out merges, but I put a lot of thought in mine. My firmware works in reality with extreme test cases. 

You are missing one critical change in my PR:
https://github.com/Smoothieware/Smoothieware/pull/1323/files#diff-4ff74fa824a22e7611630e9b99054e82L1264

I'll let you ponder what it means in relation to when the planner ticks become 0. 

I'll gladly run a G Code test case against my firmware, if you think you can demonstrate what you claim. 

Oh, and BTW. You are now in the same position that I was. If everything is thorrowly thought through, you end up fixing two issues in one PR. Against the rules apparently :-) 
https://github.com/Smoothieware/Smoothieware/pull/1323#issuecomment-398657341

Yep.

_Mark

wolfmanjm:
@markmaker I think this version will fix your issues without having to change the acceleration which was just treating the symptom not the cause. Now everything works as I expect, acceleration is not reduced to some ridiculous small number. max actuator speeds are honored and acceleration too although with this that rarely gets activated. Your fix had a side effect of hiding the math issues in the planner by reducing acceleration to a tiny number. That is no longer needed unless the axis actually has a very small acceleration. I do need to test this for extruders with very high steps/mm

src/modules/robot/Robot.cpp
wolfmanjm: no misinterpreted, but irrelevant now anyway

wolfmanjm:
both @arthurwolf and I agree this is the correct way to fix these issues.

wolfmanjm:
Note this is not currently calculating the major axis correctly. Still working on it.

markmaker:
> @markmaker I think this version will fix your issues without having to change the acceleration ...

This is mathematically impossible, as long you have coordinated moves. 

In coordinated moves, all axes move with a speed trapezoid that has the corners at the same point in time. That means that the acceleration/cruising/deceleration phases happen at the same time and have the same duration for all axes.  

In coordinated moves the toolpath must also be a straight line in multidmensional space.  The **relative position** along the toolpath as well as the derivates **speed** and **acceleration** all must have the same **ratio** between the axes (a vector). Some math - most of all [the Power Rule](https://en.wikipedia.org/wiki/Power_rule) saying that derivatives are linearly coupled - proves that to you. And this ratio between axes also works across units, so it doesn't matter if you mix axes with millimeters, degrees or cubic millimeters of filament. 

If you move X by 100mm and Y by 200mm, then the ratio is simply 2. If the X axis accelerates with 1000mm/s^2 then the Y axis must accelerate with 2000mm/s^2. If the X axis reaches a speed of 10m/min then the Y axis must reach a speed of 20m/min. Otherwise: not a straight line. It's that simple! 

In the example above I chose the X axis to be the norm (ratio=1.0) for simplicity. It is numerically more robust to take the Euclidean distance over the X, Y, Z deltas as the norm (if tose are zero, take the A, B, C deltas). This makes even more sense, as it is also the reference for the given G code feedrate according to NIST.  

So all you need to do is figure out the **ratio** of each **axis delta _versus_ Euclidean distance**. Then you can **divide** each axis' actuator speed and accleration limit by that ratio. Take the spped and acceleration minimum over all of these divided limits respectively. That's how you get the your overall limits. Take it, if it is smaller than the limit given in the G code.

If the X axis has a tiny delta, but say C has a very large one, this must mean that X also has a tiny acceleration. Could it be that you simply have difficulty in accepting that? I'm afraid the math is not going to budge for you. 

> ... which was just treating the symptom not the cause ... acceleration is not reduced to some ridiculous small number

My code works as described above. While keeping it a coordinated move, it squeezes the maximum out of the most challenged axis both for speed and acceleration (need not be the same axis btw.). It does so in a numerically robust way (not so in the original code). Your repeated accusations are completely unfounded. 

> Your fix had a side effect of hiding the math issues in the planner by reducing acceleration to a tiny number. 

Not so. Dare you to prove it. My code and firmware are there.

_Mark

arthurwolf:
> If the X axis has a tiny delta, but say C has a very large one, this must mean that X also has a tiny acceleration

But in this case we'd be calculating acceleration based on C, not X, right ?

wolfmanjm:
well I believe you have not really looked at the step generation code. The fix we have come up with is correct. Your solution IMO is not correct as it provides the incorrect distance for the given major axis which the step generation code requires.
You can choose to help or not, that is your choice, I believe this fix will fix your issues and not break other use cases, which your solution would do even though it may narrowly fix your one particular use case.

I would like to know if this fixes your issue or not.

markmaker:
> But in this case we'd be calculating acceleration based on C, not X, right ?

Well the X was still moved a tiny bit, therefore the X is still giving you the Euclidean distance. It is this Euclidean distance that governs the whole move including the **correct** interpretation of the given (G code `F` value) feedrate according to NIST standard. So the overall governing `acceleration` passed to the block will in-deed be tiny. 

But the absolute scale of the `acceleration` does not matter as this is float math. Scale in float is of no concern as long as you only multiply and divide. Remember a [floating point number uses a **significant** and an **exponent**](https://en.wikipedia.org/wiki/Floating-point_arithmetic). Multiplication and division become addition and subtraction in the exponent. The significant is multiplied more or less as if it were an int, completely ignorant of the exponent/scale. The numeric precision in multiplying or dividing two floats is compeletely scale independent. Trust it. 

>well I believe you have not really looked at the step generation code.

I have. 

The only integral that matters is the `steps_event_count` and that is the maximum over **all** axes, so in our example the C axis becomes governing in terms of step generation here: 
https://github.com/Smoothieware/Smoothieware/blob/5246c85123ccb29e86dbfd2b72f54a3bf4e9c883/src/modules/robot/Planner.cpp#L112-L113

The centerpiece is then `Block::calculate_trapezoid()`. There the tiny `acceleration` is multiplied with the abovementioned `steps_event_count` and divided by the `millimeters` coming from `distance` which is the Euclidean distance, in this example only given by X. 
https://github.com/Smoothieware/Smoothieware/blob/5246c85123ccb29e86dbfd2b72f54a3bf4e9c883/src/modules/robot/Block.cpp#L143-L145

So because X was tiny, it created a tiny `acceleration` which is again divided by tiny X. Consequently "tiny X" is cancelled out. Effectively the whole trapezoid calculation and the stepping is now governed by C. Even if X was so small that it translates into much less than a microstep, nothing bad happens. Because of float math the whole calculation chain is as precise as if X was large. 

I believe your whole crusade here was started [when you removed the acceleration scaling](https://github.com/Smoothieware/Smoothieware/pull/1398#issuecomment-492414186). That breaks everything of course. 

I will look at the new code and still try to understand it. Will also try the firmware. But now I need to go to bed. ;-) 

_Mark

src/modules/robot/Robot.cpp
```diff
@@ -1289,73 +1346,58 @@ bool Robot::append_milestone(const float target[], float rate_mm_s)
...
-                float ca = fabsf((d/distance) * acceleration);
-                if (ca > ma) {
-                    acceleration *= ( ma / ca );
-                }
+        DEBUG_PRINTF("act: %d, d: %f, distance: %f, min: %f, rate: %f, secs: %f, acc: %f\n", actuator, d, distance, actuator_min_time, rate_mm_s, secs, acceleration);
+
+        // adjust acceleration to not exceed the actuators acceleration
+        float ma =  actuators[actuator]->get_acceleration(); // in mm/sec²
+        if(!isnan(ma)) {  // if axis does not have acceleration set then it uses the default_acceleration
+            float ca = fabsf((d/distance) * acceleration);
```
markmaker: `fabs()` unnecessary as both `d` and `sos` are always positive.

src/modules/robot/Robot.cpp
```diff
@@ -1289,73 +1346,58 @@ bool Robot::append_milestone(const float target[], float rate_mm_s)
...
+            // if it exceeds the axis acceleration then we reduce the acceleration by the ratio it is over
+            if (ca > ma) {
+                acceleration *= ( ma / ca );
```
markmaker, line 1403: While mathematically correct, this is numerically not very robust, as you are daisy-chaining multiplications.  You could just take 
`acceleration = ma*distance/d;`
In my solution I use a factor to save repeated `distance/d` calculations.

src/modules/robot/Robot.cpp
```diff
@@ -1289,73 +1346,58 @@ bool Robot::append_milestone(const float target[], float rate_mm_s)
...
-
-        float actuator_rate= d * isecs;
-        if (actuator_rate > actuators[actuator]->get_max_rate()) {
-            rate_mm_s *= (actuators[actuator]->get_max_rate() / actuator_rate);
-            isecs = rate_mm_s / distance;
+        if(d < 0.00001F || !actuators[actuator]->is_selected()) continue; // no perceptible movement for this actuator
+
+        // find approximate min time this axis needs to move its distance
+        float actuator_min_time= d / actuators[actuator]->get_max_rate();
+        if (secs < actuator_min_time ) {
```
markmaker, line 1386: This whole `secs` (or the original `isecs`) calculation is not used anywhere and burns unnecessary fpu cycles. Like I described yesterday, it's all about a simple **ratio** between the Euclidean toolpath `distance` and the axis speed and acceleration. Because the **ratio** is mostly used in division, I implemented it using a `limit_factor` which is simply the inverse of the ratio. It is still 14 times faster to [mulitply than to divide on the fpu](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0439b/BEHJADED.html).

wolfmanjm: what FPU :)

wolfmanjm: of course it is used, it is what handles different actuator speeds that are lower.

src/modules/robot/Robot.cpp
wolfmanjm: thanks missed that

markmaker:
Hi @wolfmanjm 

Analyzed the code an made some demos with your firmware on my machine to show the bugs found.

The determination of a "major axis" is unnecessary due to float math, as discussed above and as demonstrated at the end of this post. The switch to `auxiliary_move` when a rotary axis is major is downright wrong and creates several problems.

1. The feedrate is no longer interpreted as the [NIST RS274NGC Interpreter - Version 3, Section 2.1.2.5 rule A standard demands](https://github.com/Smoothieware/Smoothieware/pull/1323). Suddenly the feedrate given by the G code `F` value is interpreted as degrees/min over the Euclidean space distance of the rotary axes. That will typically dramatically slow down the machine. Steps to reproduce:
```
G91 ; relative mode
G0 F600 Y10 A0     ; both moves should move equally fast in Y as per NIST standard
G0 F600 Y10 A180 ; both should take about 1 second
G90 ; absolute mode

```
Made a video. Do you need to see it?

For applications such as this...
https://www.youtube.com/watch?v=9fvpQUQBAz8
... it may burn up the tool and the work piece. 

2. By switching to `auxiliary_move` you no longer check the X, Y, Z max speeds here:
    https://github.com/Smoothieware/Smoothieware/blob/4b7d94dbef0855c26cec5cf255dcdd96c499abf0/src/modules/robot/Robot.cpp#L1341-L1342
    This may create problems with arm solutions that have very large leverage ratios at times. So the low step count might be misinterpreted as "slow" when in fact ist may be quite "fast" due to the leverage. 

3. Because junction deviation is buggy with mixed moves anyways, the following is a behaviour change rather than a real new bug: You are no longer supporting junction deviation in those moves with dominant rotary axis. All move segments with dominant step-count rotations will halt to `minimum_planner_speed` at the junction, even if segments form a straight line. Steps to reproduce:
```
; NOTE the large F value is due to bug number 1.   
G91 ; relative mode
G0 F1000000 Y10 A180 ; two equal relative moves should be done in one 
G0 F1000000 Y10 A180 ; smooth stride. Instead it now stops in the middle. 
G90 ; absolute mode 
```
I don't know if this behaviour change might even impact 3D printing when the extruder is running thin filament, thick nozzle, high extrusion rate, has a stepper with high microstep count but X, Y steppers have not and/or the X, Y arm solution has large leverage. 

The right thing to do here would be to try and expand the junction deviation calculation to all axes. I'll do that in my PR,

I recreated all the tests with my firmware and they are OK there.

I also tested 
```
G91 ; relative mode
G0 F600 Y0.000005 A180 ; this one is even suppressed in Y
    ; Note that smaller values than approx. this are parsed to 0.0 in float decimal
M400
G0 F600 Y0.00001 A180
M400
G0 F600 Y0.0001 A180
M400
G0 F600 Y0.001 A180
M400
G0 F600 Y0.01 A180
M400
G0 F600 Y0.1 A180
M400
G0 F600 Y1 A180
M400
G0 F600 Y10 A180
G90 ; absolute mode
```
To prove that with my firmware even the smallest moves in Y (even those suppressed [here](https://github.com/markmaker/Smoothieware/blob/0c2d9cd35bae8e78304fdd14a366ab10f19b12e2/src/modules/robot/Robot.cpp#L1260)) are 
a) numerically stable and generare steps correctly 
b) respect the given G code `F` value feedrate according to NIST, meaning that
c) the rotary axis moves as fast as its actuator limit allows, as long as the move in Y is very small (see bug 1)
d) when the primary Y move gets larger in proportion, it is properly limited by the given feedrate, slowing the rotary axis.  

Try it.

_Mark

wolfmanjm:
I think I have fixed your points 2 and 3. 1 is definitely an issue and I see it in my laser engraving tests.
I don't think I can fix it for all cases but I think I can minimize it. I just need to figure out exactly when we get our fixed math underflows in the step generation.

markmaker:
Hi @wolfmanjm 

> I just need to figure out exactly when we get our fixed math underflows in the step generation.

Will you let me help you? 

You first mentioned the problem here:
> Unfortunately this still does not fix the G1 X0.001 A360 issue with instant acceleration. **turns out that is a planner issue** which can't handle the very small distance but long execution time. I'll continue to see how to fix this.
https://github.com/Smoothieware/Smoothieware/pull/1398#issuecomment-492397126

Unfortunately this is a misdiagnosis ... No blame, I know how quickly it can happen. It is not a planner issue but a problem of the change here:
https://github.com/Smoothieware/Smoothieware/pull/1398/commits/114e996a01fce7f2e1bfa38e6dc476ea2dfcdeae#diff-4ff74fa824a22e7611630e9b99054e82L1363

The problem would have disppeared if you'd just reverted these few lines. You did later:
https://github.com/Smoothieware/Smoothieware/pull/1398/commits/578f96c85b7c0e93200fbec38cd59c07c254150e#diff-4ff74fa824a22e7611630e9b99054e82R1377

But it seems it was too late then, to convince you that the planner was OK before. The planner has no problem with tiny distances/feedrates/accelerations in float. The float data type can reach down to  1.1754943508 × 10−38 without losing any precision, that is 0.000000000000000000000000000000000000011754943508 so I think you can safely relax with 0.00001.  
https://en.wikipedia.org/wiki/Single-precision_floating-point_format#Single-precision_examples

Before it ever goes into integral math, the planner scales the whole calculation up to the "event count" which is determined by the axis with the largest step count. Scaling uses multiplication and division in float so it incurs no more loss of precision with tiny numbers than with normal or huge numbers Therefore it simply doesn't matter to the planner, how tiny the distance/feedrate/acceleration are.

I'm really honestly interested in getting this right for Smoothieware and OpenPNP, no fork. Ideally we figure out how to discuss this on a purely technical basis, with the egos safely confined. But you would need to be able to take a good critical look at your changes and be prepared to take some steps back with no hard feelings. 

It gave me perspective to try your firmware and make sure I did not give bogus feedback. I think it wouldn't hurt if you did the same and tried my firmware. If you can demonstrate how it does not work, we would both be much wiser. If it works it should be no problem for you to accept, that **your previous work** - i.e. the original planner - was just fine, after all. Whatever will turn out is a better situation than now.

> definitely an issue and I see it in my laser engraving tests.

Could be a good test case. Please feel free to test:
https://makr.zone/firmware_edge_2019-05-11.bin (original PR #1323 + edge merge)
https://makr.zone/firmware_PR1323_2019-05-16.bin (including mixed axes junction deviation bugfix and support)

_Mark

PR #1401, opened 2019-05-16
Add config and command to use raw hexdata to initialize/write/read stepper driver SPI

ean365:
- Added feature to MotorDriverControl.
- Added config and command to allow raw hexadecimal data words to initialize/write/read any motor drivers on the SPI.
- See comments in MotorDriverControl.cpp for function WriteReadSPIstr().
- This will work for any drivers on the SPI, even if they are not preprogrammed as options in Smoothie.

- I added it in order to be able to work with the new TMC5160 stepper driver.  In my case, I have 5 of them daisy-chained off of SPI-0, and also using the STEP, DIR, & EN pins from Smoothieboard.
(Note, however, that none of the code changes are specific in any way to the TMC5160, and is made to be used for anything on the SPI.)

- Added additional config:  motor_driver_control...reg0 through ...reg99      (see below for further comment on this)
- Example config section (in this case for initializing my 5x TMC5160 chips):

```
## Motor Driver Control SPI
motor_driver_control.motor1.enable		true		  # Whether to activate the SPI motor driver module.
motor_driver_control.motor1.axis		X			  # Motor driver axis designation [XYZABC]
motor_driver_control.motor1.chip		SPIDRVR		  # Motor driver chip name.
motor_driver_control.motor1.spi_channel		0			  # SPI bus 0 or 1.
motor_driver_control.motor1.spi_cs_pin		0.16		  # SPI SSEL pin.
motor_driver_control.motor1.spi_frequency	1000000		  # SPI SCLK frequency.

# Write raw TMC5160 driver config register data.
motor_driver_control.motor1.reg0		900005140A_9000090804_900005140Ax3	  # XYZZ:IRUN=2A,IHOLD=1A,IHOLDDELAY=1.1s; E0:IRUN=0.86A,IHOLD=0.48A,IHOLDDELAY=0.79s
motor_driver_control.motor1.reg1		91000000FFx5  # TPOWERDOWN: 5.6s
motor_driver_control.motor1.reg2		EC13008043x5  # CHOPCONF: TOFF=3,HSTRT=4,HEND=0,TBL=1,MRES=3,INTPOL=1
motor_driver_control.motor1.reg3		F0000D0000x5  # PWMCONF: ~35KHz, pwm_autoscale, pwm_autograd
motor_driver_control.motor1.reg4		800000000Cx5  # GCONF: en_pwm_mode, multistep_filt
motor_driver_control.motor1.reg5		9300000000x5  # TPWMTHRS: 
```

- Also added `sendSPIdrv` command that takes the exact same string as the .reg line.
- The instructions on how the hexdata string is formatted are in the comments before the WriteReadSPIstr() function in MotorDriverControl.cpp.  It is very versatile and useful.

```
// Parse a raw hexadecimal data string and write to SPI, then optionally print received data.
// String format:  [p]hexdata[xN][_hexdata[xN]][,hexdata[xN][_hexdata[xN]]],...]
// Writes arbitrary hexdata bytes/words.  Assumes hexadecimal -- do not prefix with '0x' or postfix with 'h'.
// Leading zeros in hexdata are significant since they are used to determine word length; i.e., number of bits to shift onto the SPI.
// If first char is 'p', then prints received data after write. If first char is 'P', then prints received data after writing hexdata twice.
// 		('p' is typically for use with a no-op/status command to view the result of a previously written command.)
//		('P' is typically for use with sending a status command and immediately sending it again to shift out, and print, its response.)
// If 'xN' at end of hexdata, then repeats hexdata N times in buffer before writing -- typically used to write the same data to multiple similar devices that are daisy-chained on the SPI bus and similarly initialized/commanded.
// Underscores '_' (and most other extraneous characters) between hexdata words denote the start of data for 'xN', otherwise ignored.
// Comma separators or end of line will write previous data as an individual SPI write (SSEL) cycle.
// Powerful but potentially hazardous, this writes free-form raw data and has no knowledge of what drivers are on the SPI bus, their word lengths, or how they are arranged on the bus.
```

- Note that in order to make this work, I had to make one small other change, which does not change how Smoothieware works or compiles, but necessary to allow the above features to work....

    - Changed util.cpp and util.h to allow pre-seeding the get_checksum() functions.  Added the 'basesum' parameter.  The function prototypes have a default parameter seed of zero, so no other code that uses it has to change.  I did this as a convenient and code-space-saving method of allowing iteration on a config item, such as above. This allows code to use the pre-calculated checksum, in this case, for "reg" and use a for-loop to add, in this case,  0, 1, 2, 3, ... onto it in order to read consecutive config lines, as above: ...reg0, ...reg1, ...reg2, and so on.   (I hope this makes sense.  See how I used this in the config section of MotorDriverControl.cpp.)

I believe that all of the above are very useful and versatile additions to Smoothieware that will allow easy experimentation and implementation of any motor driver (or anything else) on the SPI bus.  Also, the changes to util.cpp&h will probably be useful for enabling other future features.

Looking through the rest of the MotorDriverControl code implementation, I think it would benefit from being refactored, but that is for another day.  For now, I just added this feature to the existing code.


wolfmanjm:
My first comment is that M911.7 will not be allowed as it is a non standard gcode and violates NIST rules for gcodes, and so the changes to gcode also will not be allowed.
We have switched to using commands for this kind of thing ( commands being lowercase). If you could switch to that technique it would be appreciated. 
For instance a ```loadreg <hexdata_str>``` could be used instead.

I'll take a look at the rest when I have some time. Thank you.

src/modules/communication/utils/Gcode.h
```diff
@@ -48,9 +48,10 @@ class Gcode {
 
         StreamOutput* stream;
         string txt_after_ok;
+		
+        char *command;
```
wolfmanjm: I am not comfortable with this. it is private for a reason.

src/modules/utils/motordrivercontrol/MotorDriverControl.cpp
```diff
@@ -304,25 +328,27 @@ void MotorDriverControl::on_gcode_received(void *argument)
             // M911.3 S3 Zn setDoubleEdge Z=on|off Z1 is on Z0 is off
             // M911.3 S4 Zn setStepInterpolation Z=on|off Z1 is on Z0 is off
             // M911.3 S5 Zn setCoolStepEnabled Z=on|off Z1 is on Z0 is off
+			// M911.7 [p]hexdata[xN]  (Raw hexdata to/from SPI. See comments for WriteReadSPIstr function.)
 
-            if(gcode->subcode == 0 && gcode->get_num_args() == 0) {
+            if (gcode->subcode==7) { // Raw spi xmit and recv
```
wolfmanjm: Not allowed, use a command.

wolfmanjm (changes requested):

src/modules/utils/motordrivercontrol/MotorDriverControl.cpp
ean365: Jim:
I appreciate your comments and will be happy to do what makes the most sense for the project, but before that, I would like to poke at them a bit, constructively...

1. I thought that the "strict NIST RS274 compliance" train left the station a long time ago.  Isn't that getting a little orthodox at this point?

    - RepRap has several instances of strings in g-code line.  (I know Smoothieware g-code and RepRap g-code aren't meant to be totally compatible, but still it's a point.)  Perhaps I should add a letter ('S' or 'P') before the string and then put double quotes around the string, to make it a little more distinct and somewhat more compliant, as in the RepRap `M587`, etc.  I don't know if it really makes much more of a difference, though.

    - Smoothie's own (and RepRap's) `M28 <filename>` already "broke" NIST rules, and no quotes around this string either.

    - Also, gcode comments are now denoted by a semi-colon instead of inside parenthesis, as NIST rules require.

    - Smoothie's `M117 <panel_str>` is another case in point that has to be handled in such a special way that it had to be put into GcodeDispatch.cpp instead of where it truly belongs with the utils/panel code.  (I also note that this is another clear reason for providing access to `*command` in gcode.h, with respect to my comments to your other change request.  Providing `*command` access, however it is implemented [direct const char pointer access or accessor/iterator], actually allows Smoothieware to be more modular and self-consistent.)

    - I do not think it is in the Smoothieware contribution rules (i.e., strict NIST compliance), which is fine, but when did it become a thing?

    - What do you feel is really being broken, or what might this break, otherwise?

1. At first, I would have agreed that a simple_shell command would make sense here, but after thinking about it, it does not really work well for all cases, as it is intended.

    - On the one hand: a simple_shell command might make sense, if all of the drivers are daisy-chained together -- SDO of one tied to SDI of next, with a single common SSEL line to all of them (which is how I am currently configured).  This is one way to do it, such that one long write or read of the SPI bus writes/reads all chips at once (25 bytes every time in the case of 5x TMC5160 chips).  However, this is the only case that really makes sense for a simple_shell command.

    - On the other hand: the drivers could be configured such that all share a common SDI and a common SDO, with separate individual SSEL per chip.  (This is probably even the best way to configure them for future designs -- shorter write/read times, not forced to write/read all of them, which is actually more complex when all you wanted was to access just one.)  This configuration would (should) then require a simple_shell command to know which "axis" you wish to send the data to, which means it probably needs to have access to the MotorDriverControl module where it was assigned.  In which case, I assert that it is better off residing in the MotorDriverControl module where the axes and their individual SSEL lines are configured together -- again for proper modularity and consistency of Smoothieware.

    - In fact, the code that I added to the MotorDriverControl module is currently able to account for both SPI bus configurations -- daisy-chained, individual, or even a combination.  At least this is true for the configuration lines, as each `MotorDriverControl.{motor#}.reg#` line is written to the corresponding "motor#" chip-select.  _Having said that, I now realize that I should probably add an axis designator to the M911.7 gcode spec (if we keep it), so that it also properly handles both bus configurations._

1. The sendSPI function (short piece of code) that currently resides in MotorDriverConrol should really be moved out to lib, in any case.  (...and we would certainly want to if we moved my PR to simple_shell, otherwise we would need it in two places.)

Anyway, lots here, hopefully some good thoughtful philosophical arguments for you...

Best Regards,
Eric

src/modules/communication/utils/Gcode.h
ean365: Jim:

There are several (many?) cases where access to the gcode line as a string makes total sense, and is being handled in a non-standard, inconsistent, non-modular way right in GcodeDispatch (see my further comments to this later in the second change request below).  If you are concerned about some other module **_changing_** the string, or being able to, I can understand that.  However, it could be done with an object access and iterator function.....but do you relaly think that is truly necessary?  That would mean that Smoothieware would have to incorporate unchecked, uncooperative, or malicious code to make hiding `*command` a necessity.  Right?  Interested in your further thoughts on this...

Best,
Eric

wolfmanjm:
I have phased out the illegal gcodes as far as I could and I will not add new ones at this point. V2 has no exceptions either.
A command can do everything that is needed so there really is no excuse to not use a command.

Also there are historical exceptions to the data hiding principle, but again in this case I will not accept making that variable public and as the gcode will not be allowed anyway it should not be an issue.
This is the same as the two gcodes that violate the NIST standard, they were historical and would be hard to remove as we like to try to keep backward compatibility.

Moving the SPI to a library is ok.

ean365:
The NIST RS274 standard was written almost 20 years ago, has not kept up with the state of technology, and is quickly becoming outdated -- with many CNC manufacturers adding their special flavor in order to accommodate this -- it is non-extensible.  

I understand that, in a way, creating the lower-case simple-shell commands, is Smoothieware's own attempt at adding our special flavor to accommodate "non-standard" (21st century) features/capabilites.  Even if it is a "neat" place to stick all this non-standard stuff, inserting odd lower_case simple_shell commands in the middle of a gcode stream, itself is probably even more non-compliant, because it is lower_case, and RS274 doesn't allow that, and obviously because it does not even start with 'G', 'M', or 'N', etc.  For instance, NIST did not foresee how software like Smoothieware might be generic to any number of machine configurations with any number of completely different motor drivers.

I honestly don't believe that you are making the right decisions for the future, on this particular subject.  That said, I happily concede that it is not so wrong for the moment, and will "work".  I will modify this PR to work through simple_shell as I believe it is a powerful tool for the benefit of the community.  So....

With reference to my "on the other hand" bullet in my previous comment back to you:
How can a simple_shell command best access the list of objects that it works on/with?
What is the best way for a simple_command to send a request that iterates through all of the registered modules that it is relevant to (similar to ON_GCODE_RECEIVED), such that the correct module can respond to the request?  In my case, if I want my new simple_shell command to have an axis designator, such that the data is written/read to SPI using the correct spi_chip_sel pin, that axis to pin mapping is held in multiple MotorDriverControl objects.  What is the best way to do this?  I would think that we would want to keep as much of the MotorDriverControl specific code within that module, and just use the simple_shell to (minimally) parse it and send it off to the appropriate object for processing. ????

Best,
Eric

wolfmanjm:
I think there are plenty of examples of iterating over modules look at the switch command. This command would not be in the SimpleShell file it would be handled in the advanced driver module, look at the laser module for an example for the fire command.

FWIW NIST works very well. It was reprap that corrupted it, and we are not about to continue that corruption mainly due to the fact that a great deal of users use it for CNC not for 3d printing.

BTW you are more than welcome to do it however you think is correct, and then just publish your work under the third party forks section of the wiki for anyone who may agree with your decisions.

ean365:
Thanks.  I will review your references; I will put `*command` back behind the curtain; and recommit with a new simple_shell command.

wolfmanjm:
NOTE it is a command not a simpleshell command. Commands are now handled in the module they relate to.
When you revert changes to files make sure they revert cleanly and no longer show up in the changed files section, that just makes reviewing PRs harder for me.

ean365:
Ah!  I see.  Thanks again.  I never had a reason to look into those modules and did not know the distinction between "commands" and simple_shell.
I just looked over Laser.cpp and the fire command -- on the ON_CONSOLE_LINE_RECEIVED event.
When a lower-case comes in to the line parser, why do you not checksum it and dispatch like config items?  Wouldn't that be a better way and save some more flash space (multiple strings and string compares, and processing time), especially if the number of commands starts to grow?  Potentially create an ON_COMMAND_RECEIVED event?

src/modules/utils/simpleshell/SimpleShell.cpp
```diff
@@ -283,7 +283,10 @@ void SimpleShell::on_console_line_received( void *argument )
 
         } else if (cmd == "fire") {
             // these are handled by Laser module
-
+			
+		} else if (cmd == "sendSPIdrv") {
+			// these are handled by MotorDriverControl module
+			// This isn't the best way to do this. Event handler could pass a flag to see if anyone handled the event, then, if not, call a default event handler; or register the cmd_str or hash with the cmdline event dispatcher, etc...Other ways to work this out.
```
wolfmanjm: Your comment is not required thank you.

src/modules/utils/motordrivercontrol/MotorDriverControl.h
```diff
@@ -35,14 +36,15 @@ class MotorDriverControl : public Module {
         void dump_status(StreamOutput*, bool);
         void set_raw_register(StreamOutput *stream, uint32_t reg, uint32_t val);
         void set_options(Gcode *gcode);
-
         void enable(bool on);
-        int sendSPI(uint8_t *b, int cnt, uint8_t *r);
+		void WriteReadSPIstr(const char *p, StreamOutput *stream);
```
wolfmanjm: Tabs are not what we use, please make sure your whitespace conforms to our coding standards, 4 spaces per indent. Thank you.

src/modules/utils/motordrivercontrol/MotorDriverControl.cpp
```diff
@@ -153,32 +157,52 @@ bool MotorDriverControl::config_module(uint16_t cs)
...
-            switch(chip) {
-                case DRV8711: drv8711->set_raw_register(&StreamOutput::NullStream, 255, 0); break;
-                case TMC2660: tmc26x->setRawRegister(&StreamOutput::NullStream, 255, 0); break;
-            }
-        }
-
-    }else{
-        rawreg= false;
-    }
+	rawreg=false;
```
wolfmanjm: Please avoid whitespace changes, it is annoying and makes it hard to review.

src/modules/utils/motordrivercontrol/drivers/DRV8711/drv8711.cpp
```diff
@@ -15,7 +15,7 @@
 #define REGWRITE    0x00
 #define REGREAD     0x80
 
-DRV8711DRV::DRV8711DRV(std::function<int(uint8_t *b, int cnt, uint8_t *r)> spi, char d) : spi(spi), designator(d)
+DRV8711DRV::DRV8711DRV(std::function<void(uint8_t *b, int cnt, uint8_t *r)> spi, char d) : spi(spi), designator(d)
```
wolfmanjm: Why was the signature of this changed? it seems unnecessary and forces changes to many files, including some outstanding PRs that probably rely on existing signatures.

wolfmanjm:
Please fix the whitespace changes to conform to the coding standards. Then I will take another look. Thank you.

src/modules/utils/motordrivercontrol/MotorDriverControl.h
ean365: Replaced tabs with spaces.

src/modules/utils/simpleshell/SimpleShell.cpp
wolfmanjm: I am well aware that there are better ways, however the historical structure of the code does not allow events to return values, so this was the easiest way to accomplish the task. It would be too disruptive to change the code at this time. Therefore you may remove this comment.

src/modules/utils/motordrivercontrol/MotorDriverControl.cpp
```diff
@@ -51,6 +52,7 @@ MotorDriverControl::~MotorDriverControl()
 {
 }
 
+
```
wolfmanjm: more whitespace changes, please avoid these.

src/modules/utils/motordrivercontrol/MotorDriverControl.cpp
```diff
@@ -113,8 +116,9 @@ bool MotorDriverControl::config_module(uint16_t cs)
         tmc26x= new TMC26X(std::bind( &MotorDriverControl::sendSPI, this, _1, _2, _3), axis);
 
     }else{
-        THEKERNEL->streams->printf("MotorDriverControl %c ERROR: Unknown chip type: %s\n", axis, str.c_str());
-        return false;
+		chip=SPIDRVR;
```
wolfmanjm: tab

src/modules/utils/motordrivercontrol/MotorDriverControl.cpp
```diff
@@ -461,14 +516,82 @@ void MotorDriverControl::set_options(Gcode *gcode)
...
+					stream->printf("\n");
+				}
+			}
+		}
+	} while (*p++==',');
 }
 
+
+// Called by the drivers codes to send and receive SPI data to/from the chip
+void MotorDriverControl::sendSPI(uint8_t *b, int cnt, uint8_t *r /* =0 */)
```
wolfmanjm: There was no reason to change this to void.

wolfmanjm (changes requested):

src/modules/utils/motordrivercontrol/MotorDriverControl.cpp
ean365: This was not a whitespace change.

I moved `rawreg=false;` up to the top of the "rawreg" area in the config module, since I added further "rawreg" code right after it.  It made sense, at the time, moving it up there as I was adding my additional rawreg config code, in which I also set `rawreg=true;`.  Looking at how the code turned out, it would have worked the same where it was, but all things considered, it is better where it is now, yes?  (Not even sure what purpose rawreg flag serves, but included it in my code for consistency since it is also writing raw register data to the drivers.)

src/modules/utils/simpleshell/SimpleShell.cpp
ean365: Removed the gratuitous comment.  (Sorry, I know you are well aware.)

src/modules/utils/motordrivercontrol/MotorDriverControl.cpp
ean365: All the tabs are fixed.....just haven't re-committed yet.  When I do, at some point, would it be best if I rebased and collapsed down to one commit?

src/modules/utils/motordrivercontrol/MotorDriverControl.cpp
ean365: Yep, I got your first comment on this.  I'll change these back.  I was going to move it out to libs, but hadn't gotten around to it, yet.  It returns cnt, which is useless.  I was trying to clean some (simple) things up, but I get what you said about other PRs and potential conflicts.

src/modules/utils/motordrivercontrol/MotorDriverControl.cpp
ean365: Doing my best.

ean365:
Wolfman: I'm thinking it's best if I rebase back down to a single commit.  Let me know what's best for you.

wolfmanjm:
yes you can rebase that is upto you.

ean365:
All ready, I believe.

I collapsed it all down to a single commit since things changed and changed back a bit from the original idea of using a gcode to that of using a command instead.

I also updated the original pull-request comment accordingly.  (I hope that is OK, since it does make some of the later comments in this thread moot....almost like I "got it right the first time" ;-)

Also, I considered moving the sendSPI function out to libs.  But after looking around some, I thought that it would be better done as a separate pull, and in order to be done "right", should require (at least) some changes to both the MotorDriverControl and the max31855 modules, and any others that configure, reserve, use, and potentially share the SPI.

src/modules/utils/motordrivercontrol/MotorDriverControl.cpp
```diff
@@ -113,8 +114,9 @@ bool MotorDriverControl::config_module(uint16_t cs)
         tmc26x= new TMC26X(std::bind( &MotorDriverControl::sendSPI, this, _1, _2, _3), axis);
 
     }else{
-        THEKERNEL->streams->printf("MotorDriverControl %c ERROR: Unknown chip type: %s\n", axis, str.c_str());
-        return false;
+        chip=SPIDRVR;
```
wolfmanjm: Sorry I missed this, the default cannot be this SPIDRVR, it MUST be explicitly set. This saves surprises for typos in config etc.

wolfmanjm (changes requested):

src/modules/utils/motordrivercontrol/MotorDriverControl.cpp
ean365: No problem...

The raw SPI features that I added are intended to work, if invoked, for any "chip", since the feature is generic and allows for some expert debugging and tuning of any chip(s) on the SPI.  (Note that I do not actually check `chip==SPIDRVR` anywhere else.)  So, I intended for SPIDRVR to be the default.   Some of philosophical questions/options....

1. Since the raw SPI features do not depend on `chip=="SPIDRVR"`, I could get rid of the SPIDRVR enum altogether, but on second thought this is not really workable, because I want to be able to use the SPIDRVR features when I am using an otherwise undefined chip.  So, scratch this option.

1. If you prefer an error message:   If chip is not set to something recognized, then I could still have SPIDRVR be the default, but if not specifically set to "SPIDRVR", then print an error message that it is "Unknown chip type.  Defaulting to generic SPIDRVR."

1. Otherwise, if you prefer to just print an error and fail out of the module, I can just add the `switch "SPIDRVR"`, and put back the old error and fail-out.

I am partial to #2.  What do you think?

wolfmanjm: I'd prefer the original behavior, so 3

ean365: Done.

alfie0:
ean365:
I use tmc2130, how can I use M code to chang spi current, M906 should not be available. 
And enable coolStep or other projects requires real-time adjustment of many parameters, such as: monitor SG_RESULT value, increase SG value....Thank you.

ean365:
alfie0:  
If you are using the TMC2130, though my PR ([#1401](https://github.com/Smoothieware/Smoothieware/pull/1401)) works well with any driver on the SPI, you would probably be better off working with PR [#1315](https://github.com/Smoothieware/Smoothieware/pull/1315), since it is implemented specifically for configuring the TMC2130.  My PR ([#1401](https://github.com/Smoothieware/Smoothieware/pull/1401)) is meant for being able to configure other unimplemented drivers and/or unimplemented configuration options and other arbitrary data IO on the SPI bus.

alfie0:
Your method is very good. I can set up every function more reasonably.  I have made "stallGuard" and "coolStep" work very well. It would be perfect if I could send M code to read some status information.

ean365:
alfie0:

Thank you.  Have you compiled and tested this PR and found it useful, bug-free?  If so, please post a little more on your experience with it, since it is currently in the "needs testing" state, and could benefit from some good votes of confidence.  (I am certainly using it to configure, fine-tune, and run my TMC5160s with no issues.)

In any case, I tend to agree with you.  I did originally have it attached to an (otherwise unused) M-code.  However, if you read the second entry above, Wolfman had a hard philosophical reason against it.  I argued as rationally as I could -- all due RESPECT, truly, to Wolfman, I still believe his thoughtful reasons are unreasonably purist -- but I ultimately acquiesced in order to potentially offer this feature to the general Smoothie community rather than fork it in obscurity.  So, I changed it to a "command", as he suggested -- the `sendSPIdrv` command.

Nonetheless, you should be able to do everything with the `sendSPIdrv` "command" that you want to do with the M-code to write configuration data and read status, including, I think, inserting "commands" within a stream of G-code (and M-code).  Let us know if this is not the case.

You might also consider joining into the forum topic I created for this PR:  [Use any currently unsupported stepper drivers on SPI such as TMC5160](https://forum.makerforums.info/t/use-any-currently-unsupported-stepper-drivers-on-spi-such-as-tmc5160/77764).

alfie0:
Sorry, I've been a little busy lately.           

 There are some bugs in the compilation. I just recompiled it and extracted the result in the attachment, but there seems to be no problem with the firmware in use. I'll try Sendspidrv "command" later. Thank you for your prompt.
[SPI_compiled_bug.txt](https://github.com/Smoothieware/Smoothieware/files/3512676/SPI_compiled_bug.txt)

ean365:
Smoothie compiles with a lot of "warnings", which can usually be ignored, but there should be no "errors" (or "bugs").  Let us know how it goes!  Thanks.

AaronHerrick:
Question on the hardware to make this work with smoothieboard v1, did you throw them on a breadboard?  build a daughterboard, or find one?

I'm interested in trying this with Trinamic drivers, but I'm not sure which way to go with the hardware.

ean365:
Note, this PR to Smoothieware is only needed for programming Trinamic drivers wired to SPI.

Azteeg X5 and other LPC1769 boards with plug-in drivers makes it all a little easier/neater, but, assuming you want to stick with your Smoothie V1 and you are somewhat competent with electronics...

Mount the step sticks on a RAMPS board, using it as a convenient breadboard.  Then, jumper each of the ENABLE, DIR, STEP, and GND pins for each driver from the Smoothie to the appropriate pins on the RAMPS, and apply motor power to the RAMPS.  You also have to do some wiring to the top of each TMC driver (solder some of the header pins to top instead of bottom).  I used 30AWG wire-wrap to make all signal connections conveniently between header pins.

AaronHerrick:
after giving it some thought,  I've decided to get a bigtreetech board.   thanks for your advice

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

mvadu:
> alfie0:
> If you are using the TMC2130, though my PR ([#1401](https://github.com/Smoothieware/Smoothieware/pull/1401)) works well with any driver on the SPI, you would probably be better off working with PR [#1315](https://github.com/Smoothieware/Smoothieware/pull/1315), since it is implemented specifically for configuring the TMC2130. My PR ([#1401](https://github.com/Smoothieware/Smoothieware/pull/1401)) is meant for being able to configure other unimplemented drivers and/or unimplemented configuration options and other arbitrary data IO on the SPI bus.

Given the age of #1315 i think we better make this more robust and make as one SPI motor driver. I am running a Smoothie clone (tinyfab) with TMC2130 sticks. I have them configured using jumpers (so they are configured for STEP/DIR mode now), i will pull this PR and give it a spin. Only problem initially would be to read the datasheet to figure out proper values, but soon we should be able to get a good baseline of configs.

dzhang97:
Can also confirm this pr works so far with my tmc2130 setup. Will be doing more tests this week, but love the flexibility this provides, even if it has some initial overhead to understand the registers. Maybe a config snippet could be provided.

wolfmanjm:
closed in favor of PR #1447 maybe this could be added to that if people still want raw registers.

dzhang97:
@wolfmanjm I might be wrong, but #1447 doesn't have support for 2130 or 5160 drivers. This pr is a great way of having a catch all motor control type that can be used before specific driver support is added.

wolfmanjm:
It will need to be added to #1447 as #1447 would make this code not even compile.

ean365:
I'll add raw register support back in to #1447, as soon as I get a chance, assuming this refactor of MotorDriverControl is here to stay.  From what I can see with a brief look at #1447, it will be fairly straightforward.  I'm actually surprised it doesn't practically just merge right in.  #1447 doesn't break it, it just didn't include it.  I'll also take a look at making it work with the SPI -and- UART.  Yes, we absolutely do need the flexibility and universality of raw register support (in the base class).

alfie0:
Yes, we need raw register  support, tmc2130 has worked very well so far.

ean365:
In the meantime, all who need raw register SPI support, or tmc2130 support, which were closed...
Until such time as #1447 is merged into edge -- probably not for a while since it requires a good deal of testing -- my PR for raw SPI register support (this one, #1401) still merges and compiles perfectly with edge.  You can GIT it here -- [feature/SPIdriver](https://github.com/ean365/Smoothieware/tree/feature/SPIdriver) -- I will try to periodically keep it up to date with Smoothie/edge until I work it logically into #1447, maybe the new MotorDriverControl.

alfie0:
The following usage doesn't seem to work.

```
# Switch module for sendSPIdrv1 control
switch.sendSPIdrv1.enable                        true               # Enable this module
switch.sendSPIdrv1.input_on_command         M40        # Command that will turn this switch on
switch.sendSPIdrv1.output_on_command   sendSPIdrv_X_ED00018000_ED00018000_ED00028000_ED00008000          #
```

wolfmanjm:
switch module does not work that way. it is not a macro definition. Although it may be useful in some cases however someone would need to rewrite the switch module to act as a general macro facility.

ean365:
Interesting usage...  Clever!

Actually, alfie0, I don't see why it shouldn't work.  I think you just need to fix your syntax.  I haven't tested it, but I looked through the relevant code and it should support your usage.

The problem is that you used underscores within the register hexdata argument. Underscores usually work to string together the hexdata in the normal command, but when used in `switch` the underscores are all being replaced by spaces.  So, only the first "word" of hexdata is parsed.

Read `MotorDriverControl::WriteReadSPIstr` and `MotorDriverControl::on_console_line_received` (sendSPIdrv) comments -- it ignores most extraneous characters in the hexdata string.

**Try this instead:**


`switch.sendSPIdrv1.output_on_command   sendSPIdrv_X_ED00018000-ED00018000-ED00028000-ED00008000`

-or-

`switch.sendSPIdrv1.output_on_command   sendSPIdrv_X_ED00018000+ED00018000+ED00028000+ED00008000`

-or-

`switch.sendSPIdrv1.output_on_command   sendSPIdrv_X_ED00018000^ED00018000^ED00028000^ED00008000`


-or, less readable, but more succinct -- you could even do this:

`switch.sendSPIdrv1.output_on_command   sendSPIdrv_X_ED00018000ED00018000ED00028000ED00008000`


You get the idea.

Again, the code looks like this should work, though I haven't actually tested it.
Please let me/us know if it actually works!

alfie0:
I tried this, but it still didn't seem to respond. The correct result is that the x-axis should stop running.

`switch.sendSPIdrv1.output_on_command sendSPIdrv_X_ED00018000-ED00018000-ED00028000-ED00008000`

wolfmanjm:
As I said switch does not work this way. It is not setup to do macros like that, the output command only triggers on a pin level change

ean365:
Oh. OK, thanks Wolfman.

From the `switch` documentation on smoothieware.org:
> 
> This allows you to do one of the following :
> 
>     Make a G-code control a GPIO pin
>     Make a GPIO pin cause a G-code to be executed
>     Make a GPIO pin control another GPIO pin
>     Make a G-code cause another G-code to be executed
> 

I took that last bullet to mean that an M-code could trigger the command to be executed.
If it is not implemented that way, I don't think it should be too difficult to make it work that way.
I guess I need to look deeper into the `switch` code to see what I am (it is) missing in this case.

ean365:
alfie0:
Just to make sure....  When you send the
      `sendSPIdrv X ED00018000-ED00018000-ED00028000-ED00008000`
command (note no underscores) directly (not through switch), does it work the way you expect?

alfie0:
Sorry, there is no computer beside the machine now. I can't try this command

alfie0:
ean365:
Just tried, sending command through USB is OK.
`sendSPIdrv X ED00018000-ED00018000-ED00028000-ED00008000`

ean365:
alfie0:
I created a new Smoothieware pull request (#1454) that modifies the `switch` module to implement very basic "macro" functionality the way you tried above.  Give it a spin, and reply on (#1454), with comments.

PR #1403, opened 2019-05-17
even Simpler fix for A instant acceleration bug.

wolfmanjm:
Abandoning PR #1398 and partially based on PR #1323 

Makes minimal changes, but does partially fix the instant acceleration bug caused by ```G1 X0.01 A360```

However in this case it ends up using the maximum rate for the A axis rather than the requested feedrate. If this is right or wrong is up for debate. As this does not break any existing functionality or behavior, and does at least partially fix a known behavioral anomaly this will be merged and become the basis of any further work on mixing rotary axis with linear axis.

### DIFF
--- a/src/modules/robot/Robot.cpp
+++ b/src/modules/robot/Robot.cpp
@@ -100,6 +100,9 @@
 
 #define PI 3.14159265358979323846F // force to be float, do not use M_PI
 
+//#define DEBUG_PRINTF THEKERNEL->streams->printf
+#define DEBUG_PRINTF(...)
+
 // The Robot converts GCodes into actual movements, and then adds them to the Planner, which passes them to the Conveyor so they can be added to the queue
 // It takes care of cutting arcs into segments, same thing for line that are too long
 
@@ -1245,7 +1248,7 @@ bool Robot::append_milestone(const float target[], float rate_mm_s)
     // find distance moved by each axis, use transformed target from the current compensated machine position
     for (size_t i = 0; i < n_motors; i++) {
         deltas[i] = transformed_target[i] - compensated_machine_position[i];
-        if(deltas[i] == 0) continue;
+        if(fabsf(deltas[i]) < 0.00001F) continue;
         // at least one non zero delta
         move = true;
         if(i < N_PRIMARY_AXIS) {
@@ -1259,7 +1262,7 @@ bool Robot::append_milestone(const float target[], float rate_mm_s)
     // see if this is a primary axis move or not
     bool auxilliary_move= true;
     for (int i = 0; i < N_PRIMARY_AXIS; ++i) {
-        if(deltas[i] != 0) {
+        if(fabsf(deltas[i]) >= 0.00001F) {
             auxilliary_move= false;
             break;
         }
@@ -1326,6 +1329,8 @@ bool Robot::append_milestone(const float target[], float rate_mm_s)
     }
 #endif
 
+    DEBUG_PRINTF("distance: %f, aux_move: %d\n", distance, auxilliary_move);
+
     // use default acceleration to start with
     float acceleration = default_acceleration;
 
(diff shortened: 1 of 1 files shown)


markmaker:
Hi @wolfmanjm 

Thank you. I just compiled and tested the new edge and I am happy to **confirm that it resolves the issue for OpenPNP + runout calibration**. 

Like you said, this is a good new basis to iron out the remaining issues. As far as I know, **none of these remaining issues are relevant for OpenPNP** in the current version. People should not be running runout calibration on near perfect nozzles, so the extreme cases that my PR covers do not matter.

Feel free to have a look at this commit, for inspiration to fix the junction deviation for mixed axes scenarios and also make it symmetrical: 

https://github.com/Smoothieware/Smoothieware/commit/4e5c5477b1c2cea51e886068c8b4a4499cf7274a

The main trick was to include all axes in the unit_vec and then calculation the cosinus theta over all the axes. 

The issue can be demonstrated by the following G code. The move should slow down in the middle, allowing the rotation axis to reverse direction. Currently it breezes through, stalling the rotation stepper in the second leg.

```
G91 ; relative mode
G0 X100 A180 
G0 X100 A-180
G90 ; absolute mode
```

More elaborate tests demoing the other benefits are here:
https://github.com/Smoothieware/Smoothieware/pull/1323#issuecomment-493180836 

_Mark

markmaker:
Hi @wolfmanjm 

is the nightly build system still working?

It seems the firmware-latest is not being built with this PR #1403 merged:
https://github.com/Smoothieware/Smoothieware/tree/edge/FirmwareBin

Thanks for checking,
_Mark

wolfmanjm:
No the nightlies were never under our control that is @triffid servers. I have no idea if they are running or not.
The latest is built manually by me once I feel that edge is stable enough for a build like that.

Daedalus1400:
I'm still having the instant acceleration bug.  I compiled it with freshly cloned source code today with the commands "make clean" then "make AXIS=4 PAXIS=4 CNC=1", and no errors were generated.  Mixed linear/rotary moves ignore acceleration limits on the rotary axis.  This causes the rotary axis stepper to stall at speeds well below its practical maximum.

wolfmanjm:
@Daedalus1400 that is becuase a rotary axis is not a linear axis and uses totally different units to a linear axis (mm vs degrees), so PAXIS=4 will just make things worse.

Daedalus1400:
There is no observable difference between omitting PAXIS and PAXIS=4.  Both cases resulted in the _mechanical_ acceleration limits of the rotary axis being exceeded by over an order of magnitude. 

I was under the impression that this PR fixed that, so that the slowest accelerating axis sets the overall acceleration for the move.  Am I misunderstanding things here?

wolfmanjm:
PAXIS is not an appropriate setting as is discussed in the WIKI and should never be used (except under exceptional circumstances).
If this PR did not fix it I am afraid you are running into some edge condition we have not encountered. If you look at the code you will  see the acceleration is set to the lowest, however note that a rotary axis acceleration is in degrees/sec/sec and a linear axis is in mm/sec/sec so they are not directly comparable, and sometimes this makes the math impossible to reconcile.
I also suggest you look at whatever is generating the gcode for the rotary axis and make sure it is correctly setting the feedrate taking into account the diameter of the object in the rotary axis (I have seen CAM programs incorrectly genberating gcode in these circumstances).

Also it is quite possible this is not an acceleration issue but a feedrate issue. How are you determining this is an acceleration issue?

As this PR did in fact fix the issue for several people I will have to presume you have a different issue entirely.

Daedalus1400:
`G91`
`G1 A360 F21600` (one rotation in one second)
`G1 X1 A360 F60` (also one rotation in one second)

The first G1 follows the rotary axis's acceleration, which is very gentle.
The second G1 accelerates the A axis almost instantly.

There is no gcode generator or surface calculation involved, I'm manually sending gcodes as a test.

wolfmanjm:
This is a common misunderstanding I am afraid. 
In the above you are telling it to move 1mm (X1) at 60mm/min (F60) which it completes in 1 second (The F60 only applies to the X axis as per NIST standards) this means that your rotary axis has to turn 360° in 1s which is obviously way faster than is possible. What you need to do is calculate how long you want the move to take in this case I'd try 10 seconds (if you can rotate 360° in 10 seconds), and in order to move 1mm in 10 seconds means F6 not F60. When you specify an F with a primary axis (XYZ) that tells us how long the move will take for the primary axis, and the rotary axis simply has to move that fast, if you give a more realistic speed based on total time then that should give your rotary axis time to move. The NIST specification does not mention anything about acceleration for rotary axis leaving it up to the CAM to generate appropriate feeds taking into consideration the rotary axis. NIST DOES say that only the Primary (XYZ) axis are considered, and the rotary axis must simply follow that.
(In your first case there is no primary axis specified so it is considered an auxilliary move and the time is simply the speed in degrees/min you have specified, giving the rotary axis time to accelerate).

I hope that makes it clear (this is also the mistake many CAM programs make when trying to generate rotary axis speeds).

Daedalus1400:
Alas, it's not as simple as a misunderstanding.

Firstly, my rotary axis is a 2:1 belted NEMA 23 stepper driving a Sherline head stock, so it's more than capable of running up to about 5 rotations per second reliably, and accelerating up to that speed in less than 1 second.  I've tested these parameters using single axis moves.  It has no problem completing the mentioned G1 commands without stalling or missing steps.

Secondly, I am aware of how the controller calculates distances and times from the given coordinates.  The feed rates of both commands were intentionally set to produce about a 1 second move accounting for the difference in the way the microcontroller calculates feed rate.  This was done so that a difference in acceleration would be clearly visible, which it is.

I have the rotary acceleration set quite low, so the single axis move is accelerating/decelerating for almost the whole time.  The mixed axis move accelerates almost instantly and then sits at a constant speed.  To reiterate: for these example moves, the motors do not stall in either case.

wolfmanjm:
Well I just tested the same thing on my Rotary system, and the acceleration on the rotary axis did exectly what I expected, although mine cannot do 360° in 1 sec. But so long as you give the rotary axis enough time to accelerate and decelerate at its specifed rate within the time the X axis move takes it will work. Try specifying a much slower F and see what happens. But as far as this PR is concerned it works. Maybe you still have PAXIS=4 set? in which case try without specifying PAXIS and giving the A axis more time to complete with its acceleration.

Daedalus1400:
> Maybe you still have PAXIS=4 set?

I have tried recompiling without PAXIS set multiple times.  There was no change.

> Well I just tested the same thing on my Rotary system, and the acceleration on the rotary axis did exectly what I expected

Is it possible that your machine has comparable accelerations for all axes?  That would make it so that the acceleration times would all be similar, no matter which is being used to determine the final move speed, masking the problem.

> But so long as you give the rotary axis enough time to accelerate and decelerate at its specifed rate within the time the X axis move takes it will work.

The A axis is having no trouble keeping up with the X axis in _these specific_ example moves.  That's not the problem.

I have intentionally set the rotary axis's acceleration in the config file well below what it can mechanically achieve due to **other limitations** in the system, and that acceleration is not being respected during multi-axis moves.  That's the problem.

If the linear axis is moving slow enough, the acceleration time will be negligibly small even at low linear accelerations.  This means that the rotary axis simply has no time to accelerate.  If the rotary axis's motion is significantly larger, then it will have to achieve insane accelerations to match the linear axis's move.

> Try specifying a much slower F and see what happens.

There was no change in behavior.  Both the X axis and A axis accelerate in the same amount of time, governed by the acceleration of the linear axis despite the fact that the A axis is forced to exceed it's maximum acceleration.

> But as far as this PR is concerned it works.

I've looked at the code, and I agree that it **should** work, but I have a counterexample that says it doesn't.

wolfmanjm:
Well I don't know what you are doing wrong. For my test I set the A axis acceleration to a really slow one 10 times slower than the X axis so I could see if it was accelerating, and so long as it had enough time during the X move then it did accelerate and decelerate as expected. As I said your 1 sec is way too fast IMO to see the issue. try setting the speed 10 times slower and try again (without PAXIS set at all as that will undo any changes in the PR). I have done all I can to test this and AFAIK it works. So I suggest you play around some more and see if you can pinpoint the issue.

Daedalus1400:
I figured out what I was doing wrong.  It had been so long since I last compiled the firmware myself that I forgot that I had to rename "main.bin" to "firmware.bin" when moving it to the microSD card.  It now works flawlessly.

I am an idiot, and I would like to apologize for wasting your time.

Perhaps the compiler output file should be "firmware.bin" to prevent this kind of mistake in the future.

PR #1408, opened 2019-06-27
Add new panel

Serhiy-K:
Add new graphical panel for TFT LCD

TFTGLCDAdapter is external adapter based on microcontroller. It made as low cost replace for other panels with low resolution monochrome LCD.
TFTGLCDAdapter may use color TFT LCD with different chips and different resolutions.
TFTGLCDAdapter uses text screen buffer insted off graphical buffer for other panels.
TFTGLCDAdapter has own encoder and may have up to 8 buttons (include encoder button).

Courently it built on low cost STM32F103C8T6 "Blue Pill" board and color TFT LCD (ILI9325) with 320x240 resolution. Hardware project and firmware sources for it with real fotos: https://github.com/Serhiy-K/TFTGLCDAdapter.git. Also discussion about this project: https://www.radiokot.ru/forum/viewtopic.php?f=25&t=162580

### DIFF
--- a/src/modules/utils/panel/Panel.cpp
+++ b/src/modules/utils/panel/Panel.cpp
@@ -30,6 +30,7 @@
 #include "panels/ReprapDiscountGLCD.h"
 #include "panels/ST7565.h"
 #include "panels/UniversalAdapter.h"
+#include "panels/TFTGLCDAdapter.h"
 
 #include "version.h"
 #include "checksumm.h"
@@ -49,6 +50,7 @@
 #define viki2_checksum             CHECKSUM("viki2")
 #define mini_viki2_checksum        CHECKSUM("mini_viki2")
 #define universal_adapter_checksum CHECKSUM("universal_adapter")
+#define tft_glcd_adapter_checksum  CHECKSUM("tft_glcd_adapter")
 
 #define menu_offset_checksum        CHECKSUM("menu_offset")
 #define encoder_resolution_checksum CHECKSUM("encoder_resolution")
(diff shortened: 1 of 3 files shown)


wolfmanjm:
Fails travis checks.

My opinion is that as this is not a commercially available board I suspect no one will be using it. So really should not be included in the mainline build. As it can not be tested by core developers it would be hard to maintain this code base.

However it should probably be pointed to in the third party links in the wiki in case anyone does want to use it.

Serhiy-K:
> this is not a commercially available board

For privat users STM32F103C8T6 "Blue Pill" board is available, for instance, on Aliexpress from many sellers. And boards may be very different.
> I suspect no one will be using it 

I made two for my two 3D printers. After I published my article on https://radiokot.ru/artfiles/6197 two people from https://3dtoday.ru asked me about this panel and one person from https://radiokot.ru already received ordered pcb for this panel from China. This only from two Russian forums. But I don't have feedbaсk from all people, who made this panel.
> As it can not be tested by core developers it would be hard to maintain this code base.

Realy hard? Courrent panel drivers didn't change by years, only whitespaces and documentation. My driver don't affects courrent panel protocol and it is more simple. I think people should have more choice.
How many people must make this panel to include it in main code? How I may inform you about this?
> However it should probably be pointed to in the third party links in the wiki in case anyone does want to use it.

As I mentioned above, some people will be use it soon. For start it would be nice.

wolfmanjm:
we have limited flash available and simply cannot add every module.
If this project gets more popular we will consider it.
For now you need to just include a link to your fork of smoothie for any of your users, and you need to maintain it.

PR #1420, opened 2019-08-18
Feature: Add option to skip suspend gcode when head is above specific height

Igorshp:
Hello,

This is a reimplementation of a feature present in bundled firmware of 3dQuality Delta printer  (that firmware is based on Smoothieware, but I do not have source code for).

**Description:**
It is common for suspend gcode to include head lifting while moving away from the model.
On a Delta printer, when the print head is homed (top most position) and the suspend code is run, it can force the carriages out of bounds.

This problem is specific to Detla printers.

This PR adds a config option `block_susp_gcode_height`. 
If set, the suspend gcode will not run, if the printhead is above `block_susp_gcode_height` at the time.

By default, the height is set sufficiently high (65meters) as to disable this feature and make run gcode as usual.

Tested on Re-Arm board Delta printer.

### DIFF
--- a/src/modules/utils/player/Player.cpp
+++ b/src/modules/utils/player/Player.cpp
@@ -41,6 +41,7 @@
 #define after_suspend_gcode_checksum      CHECKSUM("after_suspend_gcode")
 #define before_resume_gcode_checksum      CHECKSUM("before_resume_gcode")
 #define leave_heaters_on_suspend_checksum CHECKSUM("leave_heaters_on_suspend")
+#define block_susp_gcode_height_checksum  CHECKSUM("block_susp_gcode_height")
 
 extern SDFAT mounter;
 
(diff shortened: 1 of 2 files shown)


wolfmanjm:
I see absolutely no reason for this, just do not issue suspend if you do not need it to suspend. IMO suspend should always be executed if it is issued.

Igorshp:
@wolfmanjm In this case the suspend command is sent automatically by filament switch. If a mechanical filament switch configured in switches module, when filament runs out the condition is executed. 

Since the switches are always active, changing the filament when the head is homed will also run the suspend gcode and attempt to move the head, causing the carriages to go out of bounds.

Do you have any advice how to prevent this issue in other ways?

wolfmanjm:
Yes there is a gcode that disables the filament detector, just use that before homing.

wolfmanjm:
that is presuming you are using the filament detector module. if you are using switch then don't use that. the filament detector module is specifically designed to NOT issue suspend unless it is printing.

Igorshp:
I assume are referring to the codes M405 and M406 to enable/disable filament detector.

These work for` filament detector module` only and unfortunately not applicable in this case.

If a mechanical switch is utilised and hooked up to the `switches module` I do not think it's possible to control it with gcode.

Igorshp:
Sorry, saw your second message quite a bit after my reply.

I guess the main problems I'm having is configuring the `filament detector module` for dual extruders and also getting it to work with a simple mechanical switch instead of rotary encoder.

Do you know of any projects that have done something like this? Could you point me in the right direction?

wolfmanjm:
if you do not have an encoder, then you can use just the bulge pin instead. Then you would need to use M405/406 to enable and disable it as the checks for printing won't work.
There was a reason we use an encoder for filament detection and not a simple switch.

PR #1425, opened 2019-09-04
take care of kill button pressed when setting halted to false

fishpepper:
Proposed patch for the issue in Bugreport  #1424

Issue: 
Press the killswitch and hold it down.
Now homing via $H clears the halted flag even if the killswitch is still pressed.

From my point of view clearing the halted flag while the source that caused the flag to be set is still active is not correct. 
Before clearing it should be checked if the cause is still active.

Another possible fix would be that KillSwitch subscribes to ON_HALT and sets the flag again if the button is still pressed. But thats ugly, too ;)

### DIFF
--- a/src/libs/Kernel.cpp
+++ b/src/libs/Kernel.cpp
@@ -26,6 +26,7 @@
 #include "StepperMotor.h"
 #include "BaseSolution.h"
 #include "EndstopsPublicAccess.h"
+#include "KillButtonPublicAccess.h"
 #include "Configurator.h"
 #include "SimpleShell.h"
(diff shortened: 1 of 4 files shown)


wolfmanjm:
I am sorry but this is not a good solution. It is very ugly, and I do not approve doing a public access call during this event.

wolfmanjm:
This was a very good attempt however and I commend the attempt to cover all ways of clearing HALT in a way that does not violate the module system.

However given this is an extreme edge case, where using a latching estop as the kill button, and condemning every user to having to issue a public request because of that edge case I think is simply a bad approach.

PR #1426, opened 2019-09-05
respect pushed kill button when trying to leave halted state

fishpepper:
Proposed patch for the issue in Bugreport #1424

Issue:
Press the killswitch and hold it down.
Now homing via $H clears the halted flag even if the killswitch is still pressed.

From my point of view clearing the halted flag while the source that caused the flag to be set is still active is not correct.
Before clearing it should be checked if the cause is still active.

This patch fixes that, when homing while the kill button is pressed and the button is configured with toggle the system will enter the halted state again.

### DIFF
--- a/src/modules/utils/killbutton/KillButton.cpp
+++ b/src/modules/utils/killbutton/KillButton.cpp
@@ -89,7 +89,13 @@ uint32_t KillButton::button_tick(uint32_t dummy)
                 if(killed) state= KILLED_BUTTON_DOWN;
                 break;
             case KILLED_BUTTON_DOWN:
-                if(this->kill_button.get()) state= KILLED_BUTTON_UP;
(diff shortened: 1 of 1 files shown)


wolfmanjm:
Thank you, this is so much cleaner and simpler :) I'll try to test it ASAP.

mvadu:
There are no reported issues due to this proposed change, and idea seems to be encouraged and @wolfmanjm indicated he liked the proposed change. Why keep this PR open for more than 5 months?

wolfmanjm:
Because no one has tested it, and that indicates there is no interest in this change or it has extremely limited usefulness.

wolfmanjm:
Sorry this took so long. I was finally able to test it with a latching ESTOP button. It worked fine but I added a warning message to unlatch the estop if you try to clear it while still latched.

PR #1430, opened 2019-09-23
initial attempt to handle the step rounding error accumulation.

wolfmanjm:
This takes into account the actual actuator position based on the integral number of steps, and uses that to figure out the delta move for the target position, and converts that delta to steps.

This should mitigate the rare issue where rounding up (or down) a move into steps will accumulate an error over time if the same move is done hundreds of times.


wolfmanjm:
Turns out this is a non issue. There is no accumulation of step rounding.

PR #1439, opened 2020-01-13
make zprobe compatible with rotatable delta

tordex:
To make zprobe compatible with rotatable delta arm solution we have to translate actuator position into cartesian with ``THEROBOT->get_current_machine_position`` to know how far we moved. On rotatable delta actuators don't move the same amount in Z. Also the amount of "Z" actuator move is not same as Z move performed by probe.

Also the function ``ZProbe::run_probe`` should return value in cartesian coordinates in the parameter ``mm``. So I guess this will fix potential problems with some other arm solutions.

Tested with custom Delta Robot .

### DIFF
--- a/src/modules/tools/zprobe/ZProbe.cpp
+++ b/src/modules/tools/zprobe/ZProbe.cpp
@@ -192,8 +192,9 @@ bool ZProbe::run_probe(float& mm, float feedrate, float max_dist, bool reverse)
     probe_detected= false;
     debounce= 0;
 
-    // save current actuator position so we can report how far we moved
-    float z_start_pos= THEROBOT->actuators[Z_AXIS]->get_current_position();
+    // save current position so we can report how far we moved
+    float start_pos[3];
+    THEROBOT->get_current_machine_position(start_pos);
 
     // move Z down
     bool dir= (!reverse_z != reverse); // xor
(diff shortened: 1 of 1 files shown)


wolfmanjm:
I think this is unnecessary, I wrote the rotatable arm solution and I have used zprobe with no issues. As shown in the original comments it works because when moving in Z all actuators move the same amount.

wolfmanjm:
you need to look at src/modules/tools/rotarydeltacalibration/RotaryDeltaCalibration.cpp where it relies on zprobe returning angle and not mm. It is not clear to me that zprobe is required to return the amount moved in mm. Also your change affects all arm solutions regardless, and is inaccurate and incorrect for many. Hence it is rejected. If you really need zprobe to  return mm then it needs to be dependent on the arm solution selected.

tordex:
OK, I see. Thank you. I'll investigate this more.

wolfmanjm:
Things can get quite complicated as machine position is not the actuator position if compensation is activated. As G30 Z probe is rarely used to get anything other than a relative measurement used for various things it is quite ok to do what is currently done. If you want actual distances then I suggest you use G38.2 instead which I think should do what you want, although I have never tested G38 with a rotary delta. FWIW rotary delta is rarely used and is not fully tested, and most people abandon them for linear delta due to limited resolution. They certainly have never been used successfully for 3d printing AFAIK. Usually just for pick and place.

PR #1461, opened 2020-05-01
Remove arc_milestone

wolfmanjm:
It was not being initialized after homing or after delta move, so was leaving arcs exposed to bogus moves. (For instance try to do a G2 immediately after a home or reset).
It is also no longer needed as tiny substep moves are still accumulated to machine_position, so machine position is now correct.


wolfmanjm:
Reverts changes in #1296 as no longer required.

Issue #1463, opened 2020-05-03
Hard Limit issue potentially continues through a triggered endstop and/or cause stack overflow.


ean365:
Stock edge Smoothieware appears to have broken endstop operation.
I can recreate two different issues that may be related (see bold italics below for quick summary)...

### Issue 1:
(This works with any axis;  X-axis used here as example.)

Send `G0X500` and wait for it to complete.
Send `G0X100` and then quickly hold the -X endstop in and keep holding it in for the rest of this test.
Smoothieware displays:  `Limit switch -X was hit - reset or M999 required`
Send `M999`
Smoothieware displays:  `WARNING: After HALT you should HOME as position is currently unknown`
Send `G0X100`  <--Make sure you are still holding in the -X endstop
**_Smoothieware ignores the endstop and moves the axis until it mechanically crashes into the hard stop._**

```
ok
G0X500
ok
G0X100
Limit switch -X was hit - reset or M999 required
ok
G0X100
!!
M999
WARNING: After HALT you should HOME as position is currently unknown
ok
G0X100
ok
```

This is an easy mistake to make, and Smoothieware should instead refuse to move in the direction of the already depressed endstop, and should again display the "Limit switch -X was hit - reset or M999 required" message.

### Issue 2:
(This works with any axis;  Y-axis used here as example.)

Send `G0Y100` and wait for it to complete.
Send `G0Y500` and then quickly hold the +Y endstop in and keep holding it in for the rest of this test.
Smoothieware displays:  `Limit switch +Y was hit - reset or M999 required`
Send `M999`
Smoothieware displays:  `WARNING: After HALT you should HOME as position is currently unknown`
Send `G0Y500`  <--Make sure you are still holding in the +Y endstop
Smoothieware displays many lines of:  `Limit switch +Y was hit - reset or M999 required`
**_Then, Smoothieware firmware crashes with the following:_**
```
**Hard Fault**
  Status Register: 0x40000000
    Forced
**MPU Fault**
  Status Register: 0x92
    Fault Address: 0x100073f4
    Stacking Error w/ SP = 0x100073f8
    Data Access
$T0b0c:addeadde;0d:2c120600;0e:addeadde;0f:addeadde;#1e
```

```
ok
G0Y500
Limit switch +Y was hit - reset or M999 required
ok
G0Y500
!!
M999
WARNING: After HALT you should HOME as position is currently unknown
ok
G0Y500
Limit switch +Y was hit - reset or M999 required
[...]

Issue #1264, opened 2018-02-01
Max Limit switches causing crash?


FabCreator:
#### What were you doing?

During a job, smoothie will disconnect from the host, and move into a hold state.
This occurs randomly and getting it to trigger reliably during a job has not possible.
However with some experimenting I believe there may be an issue with the Max Limit switches. 
I can replicate the fault by pressing a max end stop during the job. 

#### What did you expect to happen?
Machine to halt

#### What happened instead?
First time it may act as it should machine halts. second or third time a mass of disconnect errors followed by Com3 corrupting the USB driver and smoothie locking up with all leds solid green.

As triggering a Max end stop will cause the smoothie to freeze up.
More information already posted on the laser web github, I thought perhaps it was a coms issue to being with but same error can be reproduced with pronterface, on both of our laser cutters.
https://github.com/LaserWeb/LaserWeb4/issues/455 

#### Branch & Commit or Version of Smoothieware

 Firmware is smoothie edge-4f622ed

#### Hardware model you are using (Make, and version)

    Smoothieboard 5XC 1.1

#### Host and Operating System running the Host

   Pipo 9x Windows 10 x86
#### Link to and console logs

    [On gist.github.com or pastebin.com. Always include and never truncate.]
    
#### Link to running config and M503 results

    [On gist.github.com or pastebin.com. Always include and never truncate.]
[config.txt](https://github.com/Smoothieware/Smoothieware/files/1685987/config.txt)

Currently I have disabled the abort when limit switch is pressed and so far i have not had a crash.

M503 after disabling the abort when limit pressed.
 M503
 ; No config override
 ;Steps per unit:
 M92 X266.66599 Y266.66599 Z1280.00000
 ;Acceleration mm/sec^2:
 M204 S1500.00000 X1500.00000 Y250.00000 Z50.00000
 ;X- Junction Deviation, Z- Z junction deviation, S - Minimum Planner speed mm/sec:
 M205 X0.10000 Z-1.00000 S0.00000
 ;Max cartesian feedrates in mm/sec:
 M203 X300.00000 Y50.00000 Z10.00000
 ;Max actuator feedrates in mm/sec:
 M203.1 X300.00000 Y50.00000 Z10.00000
 ;WCS settings
 G54
 G10 L2 P1 X20.000000 Y160.000000 Z0.000000 ; G54
 ;Digipot Motor currents:
 M907 X1.80000 Y1.80000 Z1.50000 A0.80000
[...]

PR #1464, opened 2020-05-07
Fix hard limits (issue #1463)

wolfmanjm:
issue #1463 identified several bugs in both the way hard limits were detected and in recursive calls to on_idle caused by on_halt calls from on_idle.

This PR attempts to fix these issues.
1. Hard limits are now checked in the same timer as endstops for homing.
2. on_halt is flagged in the timer ISR and executed in the on_idle loop.
3. limits are not released now until all limits show clear.
4. fixed Player module which was calling abort in on_idle, now defers to main_loop.
5. cleaned up the way flush is handled in Conveyor so it does not wait.

Flagged errors in Spindle module that will need to be fixed as it calls a delay() that is ultimately called from on_halt().

The first issue though is not an issue but is by design, that is if you release HALT state with $X (or M999) the hard limits are disabled until they are all released. The reason for this is to allow the operator to jog off the hard limit (with care). 
At the moment I see no easy way to only allow movement away from a hard stop (which would be the preferred solution). However I think this can be emulated by also enabling Soft endstops, and setting them to ignor emoves that would violate the softendstop limit.

### DIFF
--- a/src/libs/Kernel.cpp
+++ b/src/libs/Kernel.cpp
@@ -318,6 +318,7 @@ void Kernel::register_for_event(_EVENT_ENUM id_event, Module *mod)
 
 // This will stop the que and stop further commands, and stop motors
 // Optionally used before on_halt() is sent to do a quick stop
+// May be called from an ISR
 void Kernel::immediate_halt()
 {
     this->halted = true;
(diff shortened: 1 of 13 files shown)


ean365:
#1264 in 2018 may well have been experiencing this bug.

ean365:
I will test these out when I can and report back.

wolfmanjm:
I was able to get a crash on one of my machines, when running into the y endstop normally. This PR seems to have fixed that. So I'd say this works and fixes the crash.
The output  I find useful now explaining hard limits are disabled, and another when they are re enabled. I'll leave the discussion of whether adding only move away from limit for another day, just to point out that on a corexy with single endstops it is non trivial to determine which endstop was hit.

PR #1475, opened 2020-06-26
help to use the code in ramps 1.4

miguelbareno:
Hi Jeffrey, I am trying to replicate your code, to be used in a 1.4 ramps system, I have written the conversion formulas and the parameters you use, and even so the robot does not move in a Cartesian way as shown in the video, I would like to ask kindly your help before this inconvenience, thank you very much


arthurwolf:
Do you mean running on an Arduino board? Or on a Smoothieboard connected to a RAMPS shield? I'm a bit confused what's being asked here.

PR #1483, opened 2020-08-25
add continuous jog mode

wolfmanjm:
Add $J -c ... to V1. Did this differently, it basically just sticks a pile of small delta moves onto the planner queue until told to stop with a Control Y. It will decelerate so will not stop dead, and may continue on several mm.


arthurwolf:
This is awesome, thanks so much! Do you need to find testers?

PR #1487, opened 2020-08-30
Proper G92 offset for ABC axes. M114.x reports all axes. Allows setti…

markmaker:
# Description

The G92 offset is stored for all axes, i.e. also for ABC. 

Based on this, M114.x can now report all axes i.e. including ABC. 

Allows setting the G92 offset, while a move is still pending. 

# Justification

G92 offsets are currently only stored for the XYZ axes. G92 on ABC are implemented by resetting the `machine_position` along with the actuator position `current_position_steps`. This is not correct, if a move is still pending, as the actuator `current_position_steps` is still being updated. 

Some M114 commands only report XYZ axes. For true 6-axis operation, we need the ABC axes reported too. Trying to implement this, the shortcoming of missing G92 offsets for ABC was revealed. If G92 is issued during a pending move, M114.1 cannot report ABC coordinates correctly. 

# Use Case

In OpenPnP we have a rotating nozzle. For optimization purposes, the nozzle will always rotate the shorter way around to a target angle. Example: if it is currently at 350° and we  want it to go to 30°, it will _not_ rotate the long way around by -320° but the short way by +40° to 390° which is the same as 30°. 

In order to avoid winding up the angle to unreadable values, we reset the angle to a modulo 360° range using G92 (more specifically to the -180° ... 180° range).

The G92 command is issued immediately after the optimized move has been commanded, e.g. after G1 C390, there will be a G92 C30. For best performance, no M400 still-stand is acceptable to do this. 

At the same time, OpenPnP needs M114.1 real-time coordinate reporting. Currently, only XYZ are reported. In order to report all axes, i.e. including ABC, we need proper G92 offsets. The current implementation of just resetting  `current_position_steps` to the G92 axis reset value is incorrect, because the move may still be pending i.e. it has not reached the milestone coordinate yet, therefore `current_position_steps` will be changed after the G92 and incur an offset. Subsequent M114.1 reports will be wrong. 

# Testing

The changes have been tested on a controller both by hand coding G-code and by using OpenPnP with the described features.

### DIFF
--- a/src/modules/robot/Robot.cpp
+++ b/src/modules/robot/Robot.cpp
@@ -120,7 +120,7 @@ Robot::Robot()
     this->compensationTransform = nullptr;
     this->get_e_scale_fnc= nullptr;
     this->wcs_offsets.fill(wcs_t(0.0F, 0.0F, 0.0F));
-    this->g92_offset = wcs_t(0.0F, 0.0F, 0.0F);
+    memset(this->g92_offset, 0, sizeof g92_offset);
     this->next_command_is_MCS = false;
     this->disable_segmentation= false;
     this->disable_arm_solution= false;
@@ -201,7 +201,9 @@ void Robot::load_config()
         // optional setting for a fixed G92 offset
         std::vector<float> t= parse_number_list(g92.c_str());
         if(t.size() == 3) {
-            g92_offset = wcs_t(t[0], t[1], t[2]);
+            g92_offset[0] = t[0];
+            g92_offset[1] = t[1];
+            g92_offset[2] = t[2];
         }
     }
 
@@ -339,7 +341,7 @@ std::vector<Robot::wcs_t> Robot::get_wcs_state() const
     for(auto& i : wcs_offsets) {
         v.push_back(i);
     }
-    v.push_back(g92_offset);
+    v.push_back(wcs_t(g92_offset[X_AXIS], g92_offset[Y_AXIS], g92_offset[Z_AXIS]));
     v.push_back(tool_offset);
     return v;
 }
@@ -413,9 +415,13 @@ void Robot::print_position(uint8_t subcode, std::string& res, bool ignore_extrud
     for (int i = A_AXIS; i < n_motors; ++i) {
         n= 0;
         if(ignore_extruders && actuators[i]->is_extruder()) continue; // don't show an extruder as that will be E
-        if(subcode == 4) { // M114.4 print last milestone
+        if(subcode == 0) { // M114 print last milestone which is the machine position with g92 offset applied for ABC
+            n= snprintf(buf, sizeof(buf), " %c:%1.4f", 'A'+i-A_AXIS, machine_position[i] + g92_offset[i]);
+        }else if(subcode == 4) { // M114.4 print last milestone in machine coordinates
             n= snprintf(buf, sizeof(buf), " %c:%1.4f", 'A'+i-A_AXIS, machine_position[i]);
-
+        }else if(subcode == 1) { // M114.1 prints real time position which is the machine position with g92 offset applied for ABC
+            // current position 
+            n= snprintf(buf, sizeof(buf), " %c:%1.4f", 'A'+i-A_AXIS,  actuators[i]->get_current_position() + g92_offset[i]);
         }else if(subcode == 2 || subcode == 3) { // M114.2/M114.3 print actuator position which is the same as machine position for ABC
             // current actuator position
             n= snprintf(buf, sizeof(buf), " %c:%1.4f", 'A'+i-A_AXIS, actuators[i]->get_current_position());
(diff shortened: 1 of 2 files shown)


wolfmanjm:
I am sorry but the limitations to M114 were by design. M114 is not to be used anyore for hosts you should use ? (as grbl does) and you will get all axis that are enabled.
So I have to reject the changes to M114 as this would break 3d hosts.
I'll look at the other things you have done, but I suspect the outcome will be the same. I'll update this when I have studied the rest of the changes.

wolfmanjm:
Additionally you should not be using G92 as this is a gobal offset and discouraged in CNC operations. You should use the WCS settings like G10 P1 Lxx etc which does work for all enabled axis.

wolfmanjm:
The G92 settings are also the way they are for the ABC axis to accomodate 3d printers that use them for reseting the extruder positions, and again this would break backward compatibility for 3D printers.

If you cannot use ? and G10 as I have suggested then again it looks like openPNP usage of smoothieware will need to maintain their own fork.

We cannot break 3d printer usage (and backward compatibility) to accomodate openPNP.
I do believe the functionality you require is already there if you use the correct gcodes and query commands.

wolfmanjm:
Ok having looked at these changes I have to confirm these would break backward compatibilty for 3d printers hosts.

You need to use ? as I have said not M114 (which is a 3d printer host command and not to be used in CNC applications).
and as G92 was already being used by 3D printer hosts (incorrectly I may add, but we are stuck with it). You will need to use G10 L10 (or L20) to set WCS instead of G92. This should all work as expected and as 3d printer hosts do not use either they are setup to work work with CNC machines corrctly.
If you want code or examples of using both ? and G10 please look at the Smoopi code base.

markmaker:
Hi @wolfmanjm 

Closed after 4 minutes. Must be a new record. 

_> You should use the WCS settings like G10 P1 Lxx etc which does work for all enabled axis._ 

That would be nice, but unless I'm blind, that's not the case. Only implemented for XYZ. 

https://github.com/markmaker/Smoothieware/blob/25543ca9c44ba24d8cdb2e391f2e294b9a9a6b56/src/modules/robot/Robot.cpp#L513-L552

Same on the M114 reporting. No G10 offsets are applied.

I even tried it. Doesn't work. As expected. 

_> I have to confirm these would break backward compatibilty for 3d printers hosts._

Hardly. All other good firmwares I looked at, report the extruder axes and/or ABC on M114, if configured:
https://reprap.org/wiki/G-code/es#M114:_Get_Current_Position
https://duet3d.dozuki.com/Wiki/Gcode#Section_M114_Get_Current_Position

But hey, I just wanted to PR this, **so nobody can say I didn't try**. I'll just have to continue [my fork for OpenPnP](https://makr.zone/smoothieware-new-firmware-for-pnp/500/), not that I really want to. 

_Mark

arthurwolf:
Hey. 

Could I get a clarification why the need to add ABC to M114 ( the standard seems to support reporting E, but I don't see ABC reported ), and why using "?" or one of the M114.x isn't an acceptable solution?

Cheers.

markmaker:
Hello @arthurwolf,

> why the need to add ABC to M114 ( the standard seems to support reporting E, but I don't see ABC reported )

First, it should just support that for the sake of consistency as soon as these axes are not configured as extruders. It's not true 6-axis support if selective features are missing for _some_ axes. Other controllers support this too. 

OpenPnP uses M114 positional feedback for some tasks like after homing, after probing and as a means of unmistakable command confirmation after fast asynchronous operation (i.e. without the need to keep track of individual "ok"s. Note that on machines with more than two nozzles, the ABC axes are often used to drive linear axes e.g. a second shared Z for nozzles 3 and 4 so the tasks apply.

> why using "?" or one of the M114.x isn't an acceptable solution?

Because "?" and the other M114.x variants do not report back proper WCS coordinates, i.e. they do not take G92 into consideration. We use G92 for various tasks, such as visual homing or rotational wrap-around (modulo 360° coordinate usage). 

_Mark

wolfmanjm:
That is incorrect, ? does indeed report both WCS and MCS. as do the various M114.x.

Additionally ? is the only corerct way to get positional feedback as it is an instant command wheares M114 is queued and also only reports the last milestone and not the actual position.

markmaker:
@wolfmanjm, 

Before I wrote my response, I looked at the code, to make sure my answer would (still) be correct. For your convenience, here are the code sections dealing with ABC:

https://github.com/Smoothieware/Smoothieware/blob/58d8e3e5695fbd38d409478cf1f1a9bb39e27026/src/modules/robot/Robot.cpp#L411-L426

https://github.com/Smoothieware/Smoothieware/blob/58d8e3e5695fbd38d409478cf1f1a9bb39e27026/src/libs/Kernel.cpp#L213-L221

https://github.com/Smoothieware/Smoothieware/blob/58d8e3e5695fbd38d409478cf1f1a9bb39e27026/src/libs/Kernel.cpp#L265-L273

_Mark

wolfmanjm:
Oh I thought you meant XYZ. ABC only have machine position as there is no offset or WCS for ABC.

markmaker:
>  ABC only have machine position as there is no offset or WCS for ABC.

Which is what this PR is about :-)

arthurwolf:
@wolfmanjm what is preventing this from being merged? would cutting it up into smaller PRs/changes that are easier to review/disucss/make sure they don't interfere, help?

arthurwolf:
After talking with @wolfmanjm on this, here's the current status:
This is a planned feature, but it's planned to be implemented a different way than the one presented in this PR.
The current plan is to implement this in the v2 firmware (which Jim is actively working on right now more than v1), and once it works, see the RAM usage, and if the RAM usage is not too much (v1 is really really short on RAM), backport to v1.

arthurwolf:
Note from @wolfmanjm for future reference: this change (as far as we can see) only supports G92 offsets, not WCS in general. We plan to do this for WCS in general (which will also include G92 offsets), but that will require possibly more RAM than v1 has available currently.  Doing just G92 without doing WCS would cause issues for CNC users, so this needs a "full" implementation right away. As explained above, the current plan is to first implement this in v2, then see if the RAM usage is low enough that a v1 port makes sense.

Issue #1341, opened 2018-09-04
Delayed Telnet response


lalten:
I'd like the Smoothieboard to respond without much delay via Telnet so my host program can react to its responses directly. Any help is appreciated.

#### What were you doing?
I'm using the Smoothieboard in a cartesian portal as a manipulator for a robot. I control the endeffector with gcode, mostly `G1` and `G38.3`/`G38.5`. The code is streamed via Telnet. The communication is ping-pong because I need the output of e.g. the probe commands to know the endeffector position.

#### What did you expect to happen?
I expected the Smoothieboard to reply `ok` as soon as it received a command (`G1`) or a command was completed (`G38`). I expected very little delay, probably less than 10ms.

#### What happened instead?
Instead, all responses are delayed, by up to 500ms. If I strictly wait for every `ok` before sending the next command, I'll have a hard time getting over 2 gcodes per second.

#### Branch & Commit or Version of Smoothieware
Build version: edge-24da5ce, Build date: Jun 14 2018 12:13:18, MCU: LPC1769, System Clock: 120MHz
CNC Build 3 axis

#### Hardware model you are using (Make, and version)
Smoothieboard 0X v1.1

#### Host and Operating System running the Host
Ubuntu 18.04
The problem appears on all hosts, including Pronterface

#### Link to and console logs
Wireshark capture of some `M114.2` and responses:  
![image](https://user-images.githubusercontent.com/11611719/45046263-edacff00-b075-11e8-904f-29292c8851a7.png)
[telnet.pcapng.zip](https://github.com/Smoothieware/Smoothieware/files/2349523/telnet.pcapng.zip)

#### Link to running config and M503 results
config: https://gist.github.com/lalten/22a74f8c6838b6d512c87acd8bc81260

#### What did I try
Since this sounds like [TCP delayed acknowledgment](https://en.wikipedia.org/wiki/TCP_delayed_acknowledgment) I did try to set `TCP_QUICKACK`, but the Wireshark logs show that it doesn't seem to have an effect (altough I'm not sure I did this correctly, see below)
```python
import telnetlib
import socket

tn = telnetlib.Telnet('192.168.88.170')
tn.read_until('\n')

tn.write('M119\n'); tn.read_until('ok\r\n')
tn.write('M119\n'); tn.read_until('ok\r\n')
tn.write('M119\n'); tn.read_until('ok\r\n')

[...]

PR #1493, opened 2020-09-08
Trigger tcp data send after handling command from command queue

taubel:
Quick and (semi) dirty fix to avoid 500 ms delay when waiting for response from telnet server. In my case this greatly improves moving in small steps, when feedback (`M400`) is used. The 500 ms issue has been noted before #1341

Tested out on my setup. Also watched traffic on wireshark

### DIFF
--- a/src/libs/Network/uip/Network.cpp
+++ b/src/libs/Network/uip/Network.cpp
@@ -58,6 +58,7 @@ Network::Network()
     hostname = NULL;
     plan9_enabled= false;
     command_q= CommandQueue::getInstance();
+    flush_network_data = false;
 }
 
 Network::~Network()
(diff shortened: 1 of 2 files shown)


wolfmanjm:
This look sinteresting, but I need to study it a bit more to see why it works :) (Although the fact we run network in on_idle is the major cause of delays as on_idle is non deterministic and can be a long delay between calls).

taubel:
I don't know how often does `on_idle` run, but in the current situation the response speed is dictated by `periodic_timer`

wolfmanjm:
periodic_timer gets triggered by on_idle() here... 

https://github.com/Smoothieware/Smoothieware/blob/edge/src/libs/Network/uip/Network.cpp#L245

cdanneil:
I'm looking for a fast response to M114.3 and M119 commands. Making these changes did not reduce the response time for these commands. Any idea where should I start if I want to speed these up?

wolfmanjm:
M114 will never be fast. Use ? if you want fast response, it is similar to GRBLS query.

cdanneil:
I'm using a 5 axis CNC build. ? doesn't seem to create a response. 
I figured an alternate fix. I send the M114.3 and M119 commands, then have the control program do other things for half a second and then come back to read the response.

wolfmanjm:
? works in all builds. Make sure you have the latest builds. On network ? may need to have a \n after it. Runnign smoothie over a network is not recommended anyway USB serial is optimized for the system, network is definitely NOT optimized and will always be slow on this platform.

cdanneil:
? works on USB serial, but I'm having no luck with telnet. Also ? only gives XYZ and I need XYZAB. 
The fix I use above is doing 2hz updates without blocking my program, so I'm happy. Thanks for the quick responses.

wolfmanjm:
If ? doesn't work in the network then you are using a very old version of the firmware. Also ? DOES report XYZABC if ABC are correctly defined. I use it in my Host Smoopi so I know it works :)

wolfmanjm:
Oh if you are actually using the telnet host then ? is eaten by the client. You need to use a raw tcp socket not telnet so ? actually gets sent. telnet only sends line by line (unless you force it into character mode)

cdanneil:
Looks like I was using old firmware. I forgot the .bin extension when loading the new file. ? is working on telnet and I'm getting readbacks on all the axis.

PR #1495, opened 2020-11-17
Enhance ThreePointStrategy to correct skew from non-perpendicular axes

DavidJRobertson:
I ran into an issue with some 3d-printed gears being skewed because the print bed on my machine isn't quite level. Although I used the 3-point leveling already, this did not address the skew at all.

The existing implementation of 3-point leveling only shifts the Z coordinate to maintain correct spacing to the bed. If the bed is not perfectly horizontal, then a cuboid printed with 3-point levelling will become a rhomboid.

To address this, I have modified the 3 point strategy to effectively tilt the entire object such that the base is parallel with the bed plane. This is done by:
1. Taking the input point, saving the supplied Z coordinate (height) for later
2. Projecting this point onto the bed plane vertically (i.e. taking the x and y coordinates of the input point and finding the z value on the plane)
3. Using this projected point as the origin for casting a ray up out of the bed, in the direction of the bed plane's normal vector, and taking the point at a distance along this ray equal to the input height saved in step 1. This point is the corrected coordinate.

Here is a quick video clip showing the impact of this change on the gear I was printing: https://www.youtube.com/watch?v=f3tT3_7rRfM&feature=youtu.be

### DIFF
--- a/ConfigSamples/Smoothieboard/config
+++ b/ConfigSamples/Smoothieboard/config
@@ -280,6 +280,7 @@ zprobe.probe_height                          5               # How much above be
 #leveling-strategy.three-point-leveling.tolerance      0.03        # the probe tolerance in mm, anything less that this will be ignored, default is 0.03mm
 #leveling-strategy.three-point-leveling.probe_offsets  0,0,0       # the probe offsets from nozzle, must be x,y,z, default is no offset
 #leveling-strategy.three-point-leveling.save_plane     false       # set to true to allow the bed plane to be saved with M500 default is false
+#leveling-strategy.three-point-leveling.correct_skew   false       # set to true to correct skew caused by unlevel bed, rather than only correcting z offset
 
 ## Panel
 # See http://smoothieware.org/panel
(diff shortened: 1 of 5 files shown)


wolfmanjm:
Although this is technically the correct way to do it, the bed would have to be considerably out of plane to make any difference. So I'd prefer this to be an optional setting with a config entry to enable/disable it and that defaults to the old behavior. In your case I would highly recommend you attempt to mechanically level your bed as best you can to minimize the skew.

Daedalus1400:
It would be nice if there was a way to account for skew between the X and Y axes as well.  My CNC's X and Y axes are out of perpendicular with each other.  I've already tried to mechanically fix it, but there's not enough adjustment to do it.  Marlin already has the M852 command which does exactly what I need, maybe their code could be adapted?

https://marlinfw.org/docs/gcode/M852.html
https://marlinfw.org/docs/configuration/configuration.html#bed-skew-compensation

wolfmanjm:
I deleted that comment as it is not relevant to this PR. If you want to open a new issue then do so, however correcting for XY errors is not something I would personally think is appropriate.

PR #1497, opened 2020-11-29
Proof of Concept to fix feedrates and accelerations of A,B, and C axis

Zaaphod:
First of all, this is a proof of concept for evaluation.  It needs more extensive testing, and the method may need to be implemented in a different way.   It has only been tested on basic movement commands, and further development will probably be needed to make sure all Gcode functions are covered.  

The problem with rotary axis is that they often not geared anywhere near the same as the linear axis on CNC machines, therefore if you have a global feedrate and try to move the rotary axis with the same feed rate as a linear axis, it is not in proportion to the linear axis.  This is not a problem when moving one axis at a time, however it becomes a problem when you want to move the rotary axis at the same time as linear axis.. then you must use the feedrate and accelerations of the linear axis and the rotary axis must follow the linear moves proportionally.  Another issue with rotary axis is that the units are not the same as linear axis, they are typically in degrees, where 360 degrees is a full turn, this is nowhere near in scale with linear axis where a move of 360 would move it 360 inches.  

It is very difficult to figure out how to make the feedrates and accelerations work between rotary and linear axis, however that is not necessary,   All that is needed is to define the rotary axis to some arbitrary unit that is on the same scale as the linear axis, then all the linear feedrates and accelerations work well with the rotary axis.. but programming the machine in arbitrary units is not very useful,  so the purpose of this experiment is to see if we simple scale all the Gcodes by a 'unit scale' and also scale all the reported coordinates by the same scale very similar to the way inches and mm are handled, then internally Smoothie can use the arbitrary units that make the feedrates and accelerations work nicely and the user can use units that they understand and that CAM programs generate.

a new config variable has been added to each axis:
alpha_unit_scale    = 1
beta_unit_scale      = 1
gamma_unit_scale =1
delta_unit_scale     =1
epsilon_unit_scale =1
zeta_unit_scale       =1

These can be overridden with M250 (just a randomly chosen Mcode to test with)

the procedure is to pick out steps_per_mm that play nice with the linear axis, for example I have X and Y steps_per_mm of 41.584645 but my rotary axis is a 50:1 gearbox so I end up with 111.11111111 steps_per_mm so that if I give it a command of G0 A360, the output goes around once  (the motor has 800 steps per rotation * 50 = 40000 steps per rotation of the output / 360 degrees = 111.111111111)   but this is nearly 3 times the steps_per_mm of my X and Y axis  so I can take my steps_per_mm and divide it by 3, and I get 37.037037...  so I set that as my steps_per_mm.  Of course if I just did that, then I would have to do a G0 A1080 to get enough steps to go around once, but the important part here is that I am now in range of my X and Y axis, so it's not trying to accelerate my A axis too fast and it's not trying to spin it faster than it can go... so now I just set delta_unit_scale = 3  and now if I do a G0 A360 it will go around once, and it will report 360 when I query the position.. but it will still do it at the accelerations and feedrates that are in proportion to my linear axis

Sometimes this needs to be done the opposite way, for example if you wanted to just connect a tap to a stepper motor shaft, now your steps_per_mm would be 2.222222,  this is completely out of range the other way so I could multiply by 20 and set my steps_per_mm of 44.44444 and that would be in range of my linear axis, then set delta_unit_scale=0.05 and now my feedrates and accelerations will be in proportion to my linear axis and 360 degrees will make one full rotation

There are times when this could be used on a linear axis as well, for example, many 3D printers and CNC routers have a really high gear ratio on the Z axis that is completely out of range of the X and Y axis, this is because they are using screws on only the Z axis and it takes many many turns to make a full rotation.  for CNC routers in this configuration it is very difficult to do 3D Milling because the Z is so far out of range of X and Y, and a 3D milling command like G1X2Y4Z-1F120  can only have one feedrate for the entire move and one acceleration for the move.  This same exact method would work to correct for this as well... but it has one disadvantage,  and that is that the Feed and acceleration parameters would be applied only to the steps_per_mm value, so if you made that 1/2 of what it really is and then used gamma_unit_scale=2 to correct for it, then doing a G1Z-1 F100 would really only move it at F50... but the benefit is that when doing a 3D move the feedrate would be appropriate for Z as well.

The second part of this experiment is to re-scale the rotary axis by parameters that can be set by Gcodes.   Since this only applies to rotary axixs, A,B and C, it should probably only be included when compiling for more than 3 axis.

This is accomplished by setting a "Rotary Scale"  variables.  This is available for A,B, and C and and they all default to 1:1 I did not make a config variable for these.  They can be changed with M251 (again an arbitrary M code I just picked for testing) 

There are two uses for this capability:

Circumference programming.... An alternate way of programming rotary axis is to pretend it is a linear axis.. for example if you wanted to engrave around a cylinder that was 4" in diameter, instead of trying to figure out all the degrees of rotation necessary, you just re-scale the rotary axis by 360/(Pi*4) or M251 A28.6479 and then pretend it is your Y axis.. substituting A values for Y values in your Gcodes.. so a G1 A1.25 will move it 1.25 inches around the circumference of the work piece.  This scale number needs to change in accordance with the diameter.    This allows someone with a CAM program that can only generate linear moves to still engrave around a cylinder.

Another way to program rotary axis is to specify the number of full rotations, so G1 A1 would make one full turn.  This would be done with a command like M251 A360.   This is useful if cutting threads, you know you want to make so many full turns as you move the X axis along, so giving a Gcode of something like G1 X2 A10 would revolve the A axis 10 times as it moves the X axis 2 inches.  So if you had a tap connected to a huge stepper motor, or (or a heavily geared small stepper motor), you could tap 1/4-20 threads 1" deep with a command G1 Z-1 A20   This makes the programs easier to understand and more useful for this application.


wolfmanjm:
Interesting concept but highly irregular and violates the NIST specification for how motion control should work when a linear axis is mexed with a rotary axis. (The latter always follows the linear axis). as I try to explain here... https://github.com/Smoothieware/Smoothieware/pull/1403#issuecomment-735277465

PR #1502, opened 2021-02-25
PT1000  TEMPERATURE SENSOR

lukrow80:
Using PT1000 there is no need of using amplifier. It can be connected to temperature port.
PT1000 module was created using code from PT100_E3D module.
Temperature sensor tested at 0 degree (icy water) it shows 0.2 degre. In boilling water shows 97 degree.


arthurwolf:
If somebody else tests this code, would you mind posting a report here as a comment? Thanks!

wolfmanjm:
I am afraid the last commit really screwed this up, so I am closing it. Please re submit and squash the commits.
Thanks

wolfmanjm:
There was also some change to temperaturecontrol.cpp that made it look like the entire file changed this was probably a whitespace change or line ending change. You will need to fix that too before resubmitting.

Thanks

PR #1507, opened 2021-05-01
Added the Sako VDF 

Rickishier:


arthurwolf:
[For reference, this is mentionned in this support message on the Facebook group: [link](https://www.facebook.com/groups/2238743883046001/permalink/2860619447525105/?comment_id=2860627570857626&reply_comment_id=2860769230843460) ]

src/modules/tools/spindle/SoftSerial/SoftSerial.h
```diff
@@ -13,7 +13,7 @@
 /** A software serial implementation
  *
  */
-class SoftSerial: public Stream {
```
wolfmanjm, line 16: won't this break other things using SoftSerial?? I think you need to leave it in.

wolfmanjm: Never mind ;)

wolfmanjm, line 16: This shouldn't be necessary as it is not in the edge branch you simply needed to merge with the current edge branch. I think this may well conflict.

Rickishier: yeah sorry, my bad....
new to this github thing ;)

PR #1510, opened 2021-06-03
Add individual arm trimming to Delta arm solution. 

50an6xy06r6n:
Involved reimplementation of the forward kinematics code. Been running this for a couple months on my Mini Kossel and it helped me compensate for some badly assembled arms.

### DIFF
--- a/src/modules/robot/arm_solutions/LinearDeltaSolution.cpp
+++ b/src/modules/robot/arm_solutions/LinearDeltaSolution.cpp
@@ -95,45 +95,6 @@ void LinearDeltaSolution::cartesian_to_actuator(const float cartesian_mm[], Actu
     }
 }
 
-// void LinearDeltaSolution::actuator_to_cartesian(const ActuatorCoordinates &actuator_mm, float cartesian_mm[] ) const
-// {
-//     // from http://en.wikipedia.org/wiki/Circumscribed_circle#Barycentric_coordinates_from_cross-_and_dot-products
-//     // based on https://github.com/ambrop72/aprinter/blob/2de69a/aprinter/printer/DeltaTransform.h#L81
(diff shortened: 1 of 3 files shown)


wolfmanjm:
We have reviewed this PR, and it does have merit. However the current speed penalty to all users and the major changes in the FK make it problematic to include as is. (I am also concerned that the FK may not be correct and wonder how well tested the FK is? ie how well tested is it by comparing the IK and FK with a set of different values).

I suggest that it would be possible to rework this so that if there are no offsets required the original code is used, and only if there are offsets defined is the new code used.
I believe this can be done by having two function pointers setup in the init() or ctor. These function pointers point to the two original functions (actually with the tower offsets removed),  if there are no offsets, and point to the new functions if any offsets are defined. This would mean no performance penalty for most users who do not use the offsets, and it would be setup automatically by defining or not defining any offsets. We could also save valuable memory by dynamically allocating the offsets if they are required, rather than always having them there regardless.

A less palatable option is to use #defines to select which version we use, although we very rarely use them in the code this option would be so rarely used that I think in this case if someone needs the arm offsets they could compile their own version with the define set.

wolfmanjm:
PS if you are not comfortable making these changes I may consider doing it myself, I'd have to find some time and then figure out how to modify the PR accordingly.

wolfmanjm:
This is what I have in mind.. not tested but you get the idea, if you like this I can merge it instead of this PR.

https://github.com/Smoothieware/Smoothieware/blob/delta_arm_length_trim/src/modules/robot/arm_solutions/LinearDeltaSolution.cpp#L84

wolfmanjm:
PS it also saves a bunch of RAM, and offsets the extra RAM needed for the function pointers when not using the offset (99% of the users)

wolfmanjm:
closed in favor of PR #1511

PR #1517, opened 2022-03-28
Add A,B, and C to G92.4

Zaaphod:
My thinking behind this, is if I have a machine without limit switches that included rotary axis, and I wanted to manually home everything, it would be convenient to do G92.4 X5 Y6 Z2 A90 and do manually homing on all axis including the rotary in one command.

Since A,B, and C don't have WCS anyway, doing a G92 with one of them is really doing something more like a G92.4, but having them included in the G92.4 command would be helpful for those realizing that what they need to do with rotary axis is a manual home, and would think they would need G92.4 for this.

### DIFF
--- a/src/modules/robot/Robot.cpp
+++ b/src/modules/robot/Robot.cpp
@@ -632,7 +632,7 @@ void Robot::on_gcode_received(void *argument)
                 }
 
                 #if MAX_ROBOT_ACTUATORS > 3
-                if(gcode->subcode == 0  && (gcode->has_letter('E') || gcode->get_num_args() == 0)){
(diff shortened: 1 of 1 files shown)


wolfmanjm:
I am not sure I like this as the plan is/was to eventually add WCS for ABC as well, which would make this very confusing. i need to think about it. I certainly do not want to change the way G92 E works as that is a special case.

Zaaphod:
When WCS is added , this change to G92.4 will stay the same, it's the normal G92 that will have to change, as G92 now does what G92.4 should do for A,B, and C.   I removed if for E

PR #1518, opened 2022-04-04
Fix Probe when in G20 mode to use coordinates specified in inches.

Zaaphod:
G38.2 - G28.5 used to properly use G20 mode, but when I tried to use the current edge, I found that they now were not in inches anymore when in G20. 

G30 was half in inches, half in millimeters.  G30 F  was using the feedrate in mm, but the coordinates were in inches.  G30 Z would set the new value of Z in inches, however the return was trying to go to the mm return value... but in inch mode.  so what should have been a return to machine coordinate Z-1 would return to Z-25.4  (yea, it crashed my probe) 

This has all been fixed and tested in both G20 and G21

G30 with return is working properly, but for some odd reason, there is a double move on the return.  It returns part of the way, stops, then all the way.  It did this before, and both in G20 and G21.  I can't see where the first retract move is coming from, but it does get to the correct end position.  It's just in 2 moves.

### DIFF
--- a/src/modules/tools/zprobe/ZProbe.cpp
+++ b/src/modules/tools/zprobe/ZProbe.cpp
@@ -207,7 +207,7 @@ bool ZProbe::run_probe(float& mm, float feedrate, float max_dist, bool reverse)
 
     // now see how far we moved, get delta in z we moved
     // NOTE this works for deltas as well as all three actuators move the same amount in Z
-    mm= z_start_pos - THEROBOT->actuators[2]->get_current_position();
+    mm= z_start_pos - THEROBOT->actuators[Z_AXIS]->get_current_position();
 
     // set the last probe position to the actuator units moved during this home
     // TODO maybe we should store current actuator position rather than the delta?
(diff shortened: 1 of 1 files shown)


wolfmanjm:
I'll try to test this. It looks ok although somewhat confusing. I think maybe a better way is for the coordinated move to make sure it passes the coordinates in the correct units rather than in the G30. But it doesn't really matter. G38 uses delta move which is always in mm. I am a little unhappy with the saved z being in inches, as the rule is that all internal measurements are in mm and converted on entry. I think this may cause issues down the road. I'll try to make sure it is done correctly in V2. But as I never use inch mode, and I suspect you are the only person using inch mode who uses smoothie issues would rarely be noticed.

Zaaphod:
Here's an idea for the G30 return.. I save the position in mm, then force inch_mode= False in the coordinated_move.  The pop_state restores it if it was active.   I tested this and it's working on my machine.  

the G38's are already just converting the given values to mm.  

so now everything should be mm, internally.

PR #1519, opened 2022-04-09
Fix technicality on feedrate conversion from inches in G20 mode.

Zaaphod:
The end result would be the same, but this is more technically correct, because it's only the given value that should be converted to millimeters, and then the divide by 60 should happen after that.


Zaaphod:
I don't know.. this is splitting hairs..  
one is converting inches per second into millimeters per second because the divide by 60 is included
the other is converting inches per minute into millimeters per minute then converting to millimeters per second..  
I think this way is more technically correct, but it can go either way...

wolfmanjm:
splitting hairs

